PR #30, opened 2018-01-04
Add fuzzers "make run"

cliffordwolf:


fuzzers/071-ppips/Makefile
```diff
@@ -24,10 +24,13 @@ $(SPECIMENS_OK):
 	touch $@
 
 run:
```
mithro: Would it make sense to move run: and other common things into a common makefile?

cliffordwolf: As I mentioned yesterday during the video conference, `make run` is not doing the same for all fuzzers.

cliffordwolf:
Can I merge this or are there any objections? @mithro

PR #31, opened 2018-01-04
Add tileconnwire.py utils script

cliffordwolf:
Signed-off-by: Clifford Wolf <clifford@clifford.at>

### DIFF
--- a/fuzzers/070-tileconn/generate.py
+++ b/fuzzers/070-tileconn/generate.py
@@ -24,6 +24,22 @@ with open("nodewires.txt") as f:
                 tilenodes[wire_tile] = dict()
             tilenodes[wire_tile][node] = wire_name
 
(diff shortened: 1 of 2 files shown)


mithro:
I assume this script is suppose to find wires which have more than 2 connections?

mithro:
This does seem to find my issue;
```
$ python3 ../../utils/tileconnwire.py HCLK_R HCLK_SW6E3
 -1   0  HCLK_L  HCLK_SW6END3
  0   1  INT_R   SW6E3
  0  -1  INT_R   SW6D3
$
```

cliffordwolf:
There is nothing inherently wrong with a wire that has >2 connections. But what might be a bit problematic is routing nodes that are not trees, as I have explained already a few times. Those two cases obviously are not the same thing.

This script is just here to simply print all connections of a wire in all constellations, to make it easier to investigate the connections a wire has.

PR #41, opened 2018-01-09
Specify and enforce code style for C++ and Python

kc8apf:
C++ uses Chromium style but with tabs for indentation.  clang-format style file is provided and used by Travis CI to check for differences.

Python uses PEP8.  autopep8 is used by Travis CI to check for differences.

If you have clang-format and autopep8 installed, 'make format' will run them on all C++ and Python source files.

### DIFF
--- a/fuzzers/005-tilegrid/generate.py
+++ b/fuzzers/005-tilegrid/generate.py
@@ -1,6 +1,9 @@
 #!/usr/bin/env python3
 
-import os, sys, json, re
+import os
+import sys
+import json
+import re
 
 
 #######################################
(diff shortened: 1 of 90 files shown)


mcmasterg:
This looks okay to me, but we should make sure this is clearly broadcasted since it may cause misc issues (editing conflicts, etc)

fuzzers/005-tilegrid/generate.py
```diff
@@ -18,7 +21,7 @@
     with open(arg) as f:
         line = f.read().strip()
         site = arg[7:-6]
-        frame = int(line[5:5+8], 16)
+        frame = int(line[5:5 + 8], 16)
```
cliffordwolf: I don't think this is PEP 8. See examples at https://www.python.org/dev/peps/pep-0008/#pet-peeves.

There are a few other things related to indices and slices in that PR that is not conforming to PEP 8. (Just saying. I'm not a PEP 8 evangelists, so I'm not saying this should block the PR. All I'm saying is that it's strange to convert PEP 8 code to non-PEP 8  code in a commit that claims to perform the opposite. ;)

kc8apf: The changes were generated by the autopep8 tool.  I assume it implements PEP8 correctly but maybe not.

PR #42, opened 2018-01-09
Fix default build settings

kc8apf:
Running 'make go' with no options will generate a Release build (-O3, no debug info, etc).  Debug builds (-O0 -g) can be built by running 'cmake -DCMAKE_BUILD_TYPE=Debug ..' in the build directory.  Tests are not built by default.  To build them, add -DPRXRAY_BUILD_TESTING=ON to the cmake command line.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,7 +1,7 @@
 cmake_minimum_required(VERSION 3.5.0)
 
 project(prjxray)
-option(PRJXRAY_BUILD_TESTING "" ON)
+option(PRJXRAY_BUILD_TESTING "" OFF)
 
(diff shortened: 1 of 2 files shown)


mcmasterg:
Can we get this added to README or something? That is a developer quickstart guide on how to run tests

kc8apf:
Instructions for building tests and building with debug info added to README.md.

mcmasterg:
Hmm I don't see the changes? Looks like there are also some conflicts in part.cc

kc8apf:
I see them on https://github.com/SymbiFlow/prjxray/pull/42/files

PR #47, opened 2018-01-13
Bitstream writer utility class

mcmasterg:
Note: this needs some cleanup / review in at least the following areas:
-General formatting. I couldn't get make format to do anything (although it does run)
-type 0 packet handling: maybe we should put something in ConfigurationPacket to make this easier to add them? If we don't want InitWithWords to make a type 0 packet, maybe something like ConfigurationPacket::type0 static member, a function, or something of that sort

### DIFF
--- a/lib/include/prjxray/xilinx/xc7series/bitstream_writer.h
+++ b/lib/include/prjxray/xilinx/xc7series/bitstream_writer.h
@@ -1,0 +1,113 @@
+/*
+ * Takes in a collection of ConfigurationPacket and writes them to specified
+ * file This includes the following: -Bus auto detection -Sync Word -FPGA
+ * configuration
+ */
+#ifndef PRJXRAY_LIB_XILINX_XC7SERIES_BITSTREAM_WRITER_H
+#define PRJXRAY_LIB_XILINX_XC7SERIES_BITSTREAM_WRITER_H
+
+#include <algorithm>
+#include <memory>
+#include <vector>
+
+#include <absl/types/optional.h>
+#include <absl/types/span.h>
+
+#include <prjxray/big_endian_span.h>
+#include <prjxray/xilinx/xc7series/configuration_packet.h>
+
+namespace prjxray {
+namespace xilinx {
+namespace xc7series {
+
+class BitstreamWriter {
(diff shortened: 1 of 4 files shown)


mithro:
@kc8apf - Please take a look at this for John.

lib/include/prjxray/xilinx/xc7series/bitstream_writer.h
```diff
@@ -0,0 +1,114 @@
...
+		absl::optional<packet_iterator> op_itr_packet;
+	};
+
+	BitstreamWriter(const packets_t& packets);
+
+	iterator begin();
+	iterator end();
+
+       private:
+	static header_t header;
```
kc8apf: Member variables get suffixes with an underscore.

lib/xilinx/xc7series/bitstream_writer.cc
```diff
@@ -0,0 +1,224 @@
...
+/**************************************************
+ *
+ *************************************************/
+
+BitstreamWriter::packet_iterator BitstreamWriter::iterator::packet_begin() {
+	// itr_packets = packets.begin();
+	const ConfigurationPacket& packet = *itr_packets;
+
+	return BitstreamWriter::packet_iterator(
+	    &packet, BitstreamWriter::packet_iterator::STATE_BEGIN,
```
kc8apf: How does STATE_BEGIN get changed to STATE_HEADER?  I'm not seeing how this works.

mcmasterg: STATE_BEGIN = 1,
			STATE_HEADER = 1,
I can eliminate STATE_BEGIN if that is confusing

lib/include/prjxray/xilinx/xc7series/bitstream_writer.h
```diff
@@ -57,11 +57,11 @@ class BitstreamWriter {
 
 		// Data iterators
 		// First over the fixed header, then the configuration data
-		state_t state;
+		state_t _state;
```
kc8apf: Underscore should be a suffix, not a prefix.

PR #50, opened 2018-01-18
Fasm2frame

mcmasterg:
FPGA assembler tool with examples as unit test

Note: for now only specify included entries like "my.thing 1" and don't explicitly include excluded entities like "my.thing 0". This should get fixed at a later date


mcmasterg:
Something was weird with the commit. Should be fixed now

PR #54, opened 2018-01-20
Add simpleroute.py tool for routing signals through ROI

cliffordwolf:
Signed-off-by: Clifford Wolf <clifford@clifford.at>

### DIFF
--- a/tools/simpleroute.py
+++ b/tools/simpleroute.py
@@ -1,0 +1,153 @@
+#!/usr/bin/env python3
+
+import getopt, sys, os, json, re
+
+
+if len(sys.argv) == 1 or (len(sys.argv) % 2) != 1:
+    print()
+    print("Usage: %s src1 dst1 [src2 dst2 [...]]" % sys.argv[0])
+    print()
+    print("Example: %s VBRK_X29Y140/VBRK_ER1BEG2 VFRAME_X47Y113/VFRAME_EL1BEG2" % sys.argv[0])
+    print()
+    sys.exit(1)
+
+class MergeFind:
+    def __init__(self):
+        self.db = dict()
+
+    def merge(self, a, b):
+        a = self.find(a)
+        b = self.find(b)
+        if a != b:
+            self.db[a] = b
+
+    def find(self, a):
+        if a in self.db:
+            c = self.find(self.db[a])
+            self.db[a] = c
+            return c
+        return a
+
+
+print("Reading database..")
+
+with open("%s/%s/tilegrid.json" % (os.getenv("XRAY_DATABASE_DIR"),
+                                   os.getenv("XRAY_DATABASE")), "r") as f:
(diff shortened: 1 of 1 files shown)


mcmasterg:
Merge this in once you run "make format" at top to fix formatting to be pep8

cliffordwolf:
> "make format" ... to fix formatting to be pep8

Done. However, I must add that most of the changes introduced by this are not necessary for PEP 8 compliance. Furthermore, this removed extra blank lines that I used to group related blocks in a meaningful way, which is explicitly allowed by PEP 8.

If yapf formatting is required, fine. But then please lets not say it has anything to do with PEP 8. :) Afaict the only thing not PEP 8 compliant in the original code was long lines.

mcmasterg:
Hhe okay. I'm just going along with what was setup. If you have objections to auto formatting stuff, you'll need to discuss that with the others

cliffordwolf:
No, I'm not against using an autoformatter. I'm just saying that if this is what we do then let's not pretend its about PEP 8. Because then we should use a PEP 8 checker instead of an autoformatter.

For example, the current version of the coded uses multiple imports in one line, which is a violation of PEP 8. But apparently that's okay because we don't care about PEP 8, we care about yapfl.

On the other hand most of the changes performed by yapfl are stuff like this:

```
-for argidx in range((len(sys.argv)-1) // 2):
-    src_tile, src_wire = sys.argv[2*argidx+1].split("/")
-    dst_tile, dst_wire = sys.argv[2*argidx+2].split("/")
+for argidx in range((len(sys.argv) - 1) // 2):
+    src_tile, src_wire = sys.argv[2 * argidx + 1].split("/")
+    dst_tile, dst_wire = sys.argv[2 * argidx + 2].split("/")
```

But the original code here is perfectly fine with regard to PEP 8. So obviously PEP 8 is a red herring. That's all I wanted to point out.

kc8apf:
I think the confusion comes from YAPF's configuration file declaring that our style is based on YAPF's PEP8 style.  I'm hearing you, say that YAPF's PEP8 style isn't strictly the minimum set of changes required by PEP8.

cliffordwolf:
In addition to that: Other changes, that would be required by PEP 8 are not performed by the tool (such as multiple imports in one line). So "PEP 8" and "YAPF's PEP 8 style" are two different things, none is a subset of the other.

However, saying "this is because of PEP 8 compliance" gives something a certain ring of authority. But if in fact the changes requested have nothing to do with PEP 8 then it's just nonsensical namedropping. And that's sad because "we standardize on YAPF output" on it's own without even mentioning PEP 8 would of course be a perfectly sane position.

If however PEP 8 is the desired goal then YAPF is not the right tool imo. (I assume PEP 8 is not the goal and it's about canonical style and just about any autoformatter will do and I'm sure YAPF is a fine one.)

PR #55, opened 2018-01-22
FASM and xc7patch proof of concept using partial reconfig flow

kc8apf:


Makefile
```diff
@@ -8,6 +8,6 @@ go:
 	cd build; cmake ..; make -j$(JOBS)
 
 format:
-	find . -name \*.cc -and -not -path './third_party/*' -exec $(CLANG_FORMAT) -style=file -i {} \;
-	find . -name \*.h -and -not -path './third_party/*' -exec $(CLANG_FORMAT) -style=file -i {} \;
-	find . -name \*.py -and -not -path './third_party/*' -exec yapf -p -i {} \;
+	find . -name \*.cc -and -not -path './third_party/*' -and -not -path './.git/*' -exec $(CLANG_FORMAT) -style=file -i {} \;
```
mcmasterg: Did this corrupt something in .git?

minitests/partial_reconfig_flow/Makefile
```diff
@@ -0,0 +1,83 @@
+# Top-level target for generating a programmable bitstream.  Given a .fasm
+# file, calling make with the .fasm extension replaced with .hand_crafted.bit
+# will generate a bitstream that includes both the harness and the .fasm design
+# ready for programming to a board.  For example, 'make
+# roi_noninv.hand_crafted.bit' will generate a bitstream that includes the
+# design from roi_noninv.fasm. 
+%.hand_crafted.bit: init_sequence.bit %.no_headers.bit final_sequence.bit
+	cat $^ > $@
+
+%.no_headers.bit: %.patched.bit
```
mcmasterg: Doesn't this make a .bin instead of a .bit?

minitests/partial_reconfig_flow/Makefile
```diff
@@ -0,0 +1,83 @@
...
+		--part_file ${XRAY_PART_YAML} \
+		--bitstream_file harness_routed.bit \
+		--frm_file $< \
+		--output_file $@
+
+# xc7patch currently only generates the actual frame writes which is
+# insufficient to program a device.  Grab the initialization and finalization
+# sequences from the harness bitstream so they can be tacked on to the
+# xc7patch-generated bitstream to create a programmable bitstream.
+init_sequence.bit: harness_routed.bit
```
mcmasterg: This looks super fragile. I'm not entirely clear where some of these numbers are coming from. For example, if I want to target arty a7 instead of basys3 (changes harness pnr), do these all need to be changed?

Makefile
kc8apf: No, yapf would throw an exception though.  I hadn't anticipated a branch named simpleroute.py.

minitests/partial_reconfig_flow/Makefile
kc8apf: I doubt it.  I'll add an explanation of how I came up with these offsets.

minitests/partial_reconfig_flow/Makefile
kc8apf: Good point.  That probably needs to be fixed in a few places.

PR #59, opened 2018-01-23
OpenOCD board configuration for basys3

kc8apf:
Signed-off-by: Rick Altherr <kc8apf@kc8apf.net>


mcmasterg:
Might it be better to put this in utils/data or something like that?

Issue #58, opened 2018-01-23
"FPGA assembly": syntax


mcmasterg:
"FPGA assembly" (.fasm file) as its currently called, refers to an end user issuing low level instructions in a human readable text format to configure FPGA interconnect and logic elements such as FFs and LUTs. This could be hand written our dumped by a PNR tool. As an FPGA user should not generally be aware of segments or other internal bitstream details, this syntax was specifically created to not have any knowledge of segments or frames.

There have been some semi heated discussions about what syntax we should use. Today, we have a format that is very explicit but perhaps too verbose. There have been some suggestions to simplify it to match segprint closer, or to eliminate segments entirely and unify the two formats. I'm going to try to describe some of the background on this and then follow up with some more personal opinions.

Issue 27 describes proposals to change tilegrid: https://github.com/SymbiFlow/prjxray/issues/27 Now that I have more background on the bitstream format, I need to review this in more detail to better understand what the core issues are.

There was also some discussion on twitter: https://twitter.com/oe1cxw/status/954402448436756480

This was followed up with a brief chat with Clifford where he encouraged us to reference issue 27 regarding his thoughts on this matter. This implies to me that he considers this ticket and issue 27 the same thing (that is if we eliminate tiles segprint will produce something usable to end users).

So basically the questions are:
-How closely related should the format used to document the FPGAs be to the format our PNR tools spit out?
-Is tile based naming the right level for this?
-How closely do we want to follow Vivado naming?
-How verbose should we be? For example, some are advocating for implicit elements to be removed from the syntax.

mcmasterg:
From some quick discussions, it sounds like I should go ahead and change the way that I name slices to use relative names instead of global names. The reason why its like this right now is because originally I wanted names that look like this:

SLICE_X12Y102.AFF.DMUX.AX 1

However, Rick convinced me that to start we should start everything with a tile name to be more regular. I conceded to this with the comment that I might want something like above supported as an extension if we do a lot of FPGA assembly operations. So I did the most straightforward thing which was to simply prepend the tile name:

CLBLM_L_X10Y102.SLICE_X12Y102.AFF.DMUX.AX 1

But now this has two global coordinates in it. We probably want something with X0/X1 to make it relative to the tile. If we use segprint convention it might look something like this:

CLBLM_L_X10Y102.SLICEM_X0.AFF.DMUX.AX 1

[...]

PR #61, opened 2018-01-23
Fasm2frame oparg

mcmasterg:
Optional fasm argument for some entries as outlined here: https://github.com/SymbiFlow/prjxray/issues/58

Eventually this may need some additional error handling if we get anything that is a multi bit configuration that isn't interconnect


tools/fasm2frame.py
```diff
@@ -105,7 +117,9 @@ def frame_set(frame_addr, word_addr, bit_index):
         # Initiaize bitstream to 0
         frames_init()
 
+    line_number = 0
     for l in f_in:
+        line_number += 1
```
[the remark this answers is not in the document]
mcmasterg: Sure, but then its going to still be line_number + 1 since I would argue the first line is 1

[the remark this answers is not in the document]
kc8apf: enumerate() takes an optional start parameter since 2.6. 
 https://docs.python.org/2.7/library/functions.html#enumerate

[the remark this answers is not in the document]
cliffordwolf: jfyi: enumerate() has an optional 2nd parameter for the start value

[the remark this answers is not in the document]
mcmasterg: hunh didn't know that. fixed

tools/fasm2frame.py
```diff
@@ -16,25 +44,49 @@ def get_database(segtype):
...
+                    "Expect single bit DB entries to be set, got %s" % l)
+            segbitsdb[segtype][name] = {None: (seg_word_column, word_bit_n)}
+        else:
+            # An enumerated value
+            # Split the base name and selected key
+            m = re.match(r'(.+)[.](.+)', name)
+            name = m.group(1)
+            key = m.group(2)
+            # May or may not be the first key encountered
+            m = segbitsdb[segtype].get(name, {})
```
kc8apf: I'm getting confused as to what is happening with the shadowing of 'm' and 'l'.  I think you want setdefault() here instead of get() and the assignment below.

tools/fasm2frame.py
```diff
@@ -16,25 +44,49 @@ def get_database(segtype):
...
+            segbitsdb[segtype][name] = m
+            l = []
+            m[key] = l
+            #print("Add %s.%s" % (name, key))
+            for bit_val in bit_vals:
```
kc8apf: This loop feels ripe for being a list comprehensions:

m[key] = [parsebit(x) for x in bit_vals]

No need to unpack the results of parsebit() just to repack them into a tuple again.

tools/fasm2frame.py
```diff
@@ -140,53 +206,60 @@ def frame_set(frame_addr, word_addr, bit_index):
...
+                "Segment DB %s, key %s not found from line '%s'" %
+                (segj['type'], db_k, l))
+        '''
+        Creating the key depends on whether its an enumerated or a fixed value
+        If its enumerated, the value forms part of the key
+        We'd also like to ideally identify if there is a syntax error
+        Ultimately probably easier to separate out these two cases
+        '''
+        # A single bit value?
+        if None in db_vals:
```
kc8apf: I'd rather see 1 used as the key in db_vals and this test changed to len(db_vals) == 1.

kc8apf: Better yet, for single bit values, insert two keys for 0 and 1.  If no value is provided in the FASM line, assume they meant 1.  Then you can use the same code path for looking up values here.

tools/fasm2frame.py
mcmasterg: Changed in PR. Yes, didn't know that existed

tools/fasm2frame.py
mcmasterg: Changed in PR

tools/fasm2frame.py
mcmasterg: I don't like the first suggestion, but second sounds very reasonable to me. Changed in PR to that

PR #71, opened 2018-01-30
roi_harness Basys3 demo

mcmasterg:
See README for instructions how to use (basically ./runme.sh && python3 demo_sw_led.py ...)
simpleroute.py made more library friendly. Also has some misc .v tests for bin2fasm, but we are currently not planning to use these


minitests/partial_reconfig_flow/v2bit.sh
```diff
@@ -0,0 +1,39 @@
...
+    echo "bad prefix"
+    exit
+fi
+echo "$fin => $prefix"
+
+make
+mkdir -p $prefix
+
+vivado -mode batch -source /dev/stdin <<EOF
+read_verilog $prefix.v
```
kc8apf: FWIW, the Makefile has rules to go from .v to .dcp, .bit, and .fasm.  I think you can replace most of this with 'make ${prefix}_roi_partial.fasm', 'make ${prefix}.bit', and 'make ${prefix}_hand_crafted.bit'.  Those changes must have been submitted while you were working on this script.

mcmasterg:
I don't care about v2bit.sh as much at this time. I can remove it from the
PR if its blocking getting the rest merged in

kc8apf:
I don't think it will work as is.  Dropping it from this PR is probably the
best option.

PR #91, opened 2018-02-13
Renamed all README.txt files to README.md

unixb0y:
This was an open Issue so I did it.

Signed-off-by: Davide <davide.toldo@stud.tu-darmstadt.de>


mithro:
Hi @unixb0y,

Thanks for sending this pull request! I'm merging this now.

Once this has been merged, it would be good to go through each file and make sure that it is formatted correctly now. Would you have time to do that?

@mithro

unixb0y:
@mithro I'm pretty sure the txt files were plain text so the markdown version should look exactly the same and they're most probably "formatted correctly" as well.
Or do you mean adding the markdown tags that make headers appear bold / larger etc.?

unixb0y:
Actually, you're right - markdown files e.g. don't recognize a line break if there aren't 2 spaces at the end of the line and this sort of stuff. Working on it now.

PR #108, opened 2018-04-16
Merged terms from wiki glossary

sarahmaddox:
Added terms from wiki (https://github.com/SymbiFlow/prjxray/wiki/Glossary) and cleaned up some words.
Signed-off-by: Sarah Maddox <sarahmaddox@google.com>

### DIFF
--- a/docs/architecture/glossary.rst
+++ b/docs/architecture/glossary.rst
@@ -40,15 +40,14 @@ Glossary
     A CLB is a combination of basic logic elements (:term:`BELs <BEL>`).
 
   Database
-    A term used in *Project X-Ray* to denote
-    text files containing meaningful labels for bit positions within
+    Text files containing meaningful labels for bit positions within
     :term:`segments <segment>`.
 
   Frame
     The fundamental unit of :term:`bitstream` configuration data consisting of
     101 :term:`words <word>`.
     Each frame has a 32-bit frame address and 101 payload words, 32 bits each.
-    The 50th payload words is an EEC.
+    The 50th payload word is an EEC.
     The 7 LSB bits of the frame address are the frame index within the
     configuration :term:`column` (called *minor frame address* in the Xilinx
     documentation). The rest of the frame address identifies the configuration
@@ -74,7 +73,7 @@ Glossary
 
   Node
     A routing node on the device. A node is a collection of :term:`wires <wire>`
-    spanning one or more "tiles"term" `tiles <tile>`.
+    spanning one or more :term:`tiles <tile>`.
     Nodes that are local to a tile map 1:1 to a wire. A node that spans multiple
     tiles maps to multiple wires, one in each tile it spans.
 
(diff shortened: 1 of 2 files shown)


docs/architecture/glossary.rst
```diff
@@ -3,70 +3,131 @@ Glossary
 
 .. glossary::
 
-  basic element
   BEL
-  basic logic element
   BLE
-    For example a LUT5, LUT6, CARRY4, or MUX, but not PIPs.
+  Basic element (BEL) or basic logic element (BLE)
```
kc8apf: Having these on separate lines defines multiple terms that all have the same definition (see http://www.sphinx-doc.org/en/stable/markup/para.html#glossary).  As is, this breaks any :term references on other pages.

sarahmaddox: Thanks! Will do. I fixed the DCO check and put the BEL/BLE terms on separate lines. I haven't yet run the build to check errors - that's my next step.

sarahmaddox:
PTAL - ready for review. I ran the HTML and link checkers. The link checker picked up a previously-existing problem in the architecture overview: a link to a non-existent glossary term. I fixed that link.

docs/architecture/glossary.rst
```diff
@@ -7,66 +7,129 @@ Glossary
...
 
-  configurable logic block
   CLB
-    Basic building block of logic. 
-
-  frame
-    Fundamental unit of configuration data consisting of 101 :term:`words <word>`.
-
-  half
+  Configurable logic block (CLB)
```
kc8apf: This form makes the parenthesized initialism part of the glossary term used by other pages to link to this entry.  I'm unclear on what a good pattern is for this.  Do we restate the term and initialism in the definition?  Do we expect the reader to understand that the two term entries are equivalent?

docs/architecture/glossary.rst
```diff
@@ -7,66 +7,129 @@ Glossary
...
-    Portion of a tile where :term:`BELs <BEL>` can be placed.  :term:`Slices
-    <slice>` in a :term:`CLB` tile are sites.
-
-  slice
-    Portion of a :term:`CLB` tile that contains :term:`BELs <BEL>`.
-
-  tile
+    
+  ROI
+  Region of interest (ROI)
```
kc8apf: Same as above.

docs/architecture/glossary.rst
sarahmaddox: Good point. I've moved the abbreviation out of the term, so other pages don't need to include the abbreviation if they want to link to the full term. I ran the html and link checkers after making the change - all good.

On the pattern: I do think we should restate the term and the abbreviation in the definition when they're not obvious. We want to be as clear as possible, and also to tax the reader's internal parser as little as possible. It's not necessarily clear to the reader that a sequence of bold terms, each on a different row (BLE, Basic Logic Element, BEL, etc) are all the same thing, or that a particular phrase is the expansion of a given abbreviation, or that the next available definition applies to them all. Case in point: When I first came across this glossary, I thought we had a bunch of terms without definitions.

Another interesting point this thread raises, is the syntactical pattern of the definitions. If we repeat the term, we're forced to use a full sentence in the first part of the definition. Like: "A configurable logic block (CLB) is the configurable logic unit of an FPGA," rather than just a phrase: "The configurable logic unit of an FPGA."

I had a look at a few glossaries, especially those on Read the Docs sites, and they merrily mix and match full sentences versus phrases. I think it's OK for us to do that too, provided our goal is clarity and ease of comprehension.

docs/architecture/glossary.rst
sarahmaddox: Thanks! I've updated this one, and also the PIP entry above.

PR #118, opened 2018-09-17
fabric timing fuzzer

mcmasterg:
Beta release of the timing fuzzer. See README for description of how it works

On reviewing overall methodology / math stuffs: I talked to @cliffordwolf about methodology earlier and sounded like he thought it was reasonable. I'm hoping @litghost will do a more detailed methodology review at some point, but we did discuss things briefly. One thing he brought up is that we may be unfairly weighting variables that are orders of magnitude away (ex: carry chain 7 ps delay vs 200 ps PIP delay)

Known issues / limitations:
* fabric only. Need to do a follow up with BELs, although these are suspected to be much easier. This must be completed to get a full timing model
* investigate and remove ZERO delay elements (such as virtual PIPs). Removing these should significantly improve the timing solution. This should be fixed before I move onto other projects given changes may be more invasive, it requires somewhat specialized knowledge, and results may be useful for solving other bitstream elements
* Tweak project parameters. There are a lot of knobs on how much data to feed in but I haven't (recently) tweaked things. Ex: how many LUTs to feed into the fuzzer, how tightly packed, etc
* We could probably get better numbers by writing dedicated torture tests (ex: a design that fans out once, then twice, then three times, etc)
* Results are not deterministic. I think this is mostly from using dict() instead of OrderedDict() type stuff
* We will eventually want a workflow to combine multiple projects together. This should be a relatively small tweak to the makefiles to just aggregate timing3.csv files from multiple projects instead of just one. However, there are some more complicated things that might be possible to do for performance reasons (ex: solve rref per project and then combine rref results)
* There is no test suite. Maybe worth adding something?

Other:
* Vivado indicates there is a way to get intermediate path delays, but we've never been able to get that flag to work. Getting this to work would significantly increase the model accuracy and such


mcmasterg:
Most of the CI failures look unrelated to this PR (htlmgen.py, segdata.py, etc). Should I create another PR to fix these first?

Other: also should have mentioned there are a few suspicious results such as:
```
                "VBRK_EE2A0": [
                    null,
                    0,
                    null,
                    0
                ],
```
Where there are mixed null and 0 values. This means that the corner was partially solved for, which isn't supposed to happen. Likely a bug in the way min corners are getting merged in

mcmasterg:
Re: formatting issues, see https://github.com/SymbiFlow/prjxray/issues/120

PR #123, opened 2018-09-19
WIP: Add step to extract site pin to tile nodes.

litghost:


mithro:
Do you have a sample output file?

fuzzers/072-site_pins/generate.tcl
```diff
@@ -0,0 +1,69 @@
+# This script first finds a list of all tile types.  It then outputs all nodes
+# within each tile, and any connected site_pins.  It also outputs all sites
+# present within the tile, its site type, and the list of site pins.
+#
+# This information should be sufficent to map the tile routing resources to
+# the sites within a tile.
+#
+# The output format is JSON5, which can be converted to JSON with json5.tool:
+# python -mjson5.tool --json {input JSON5 file} > {output JSON file}.
```
mcmasterg: We've standardized on JSON. Can we add that to the Makefile?

litghost:
Sample output for CLBLL_L: https://gist.github.com/litghost/5258715b4bd696b724189559a9e9ced9

fuzzers/072-site_pins/generate.tcl
litghost: Done

litghost:
I've updated PR with sample output and conversion to JSON and more useful structure of data.

mcmasterg:
run "make format" resolve the formatting thing (runs a PEP8 formatter)

litghost:
Formatting update, should be good

mcmasterg:
Looking at the sample data, you are favoring lists over dictionaries. Wouldn't it be better to change the lists to dictionaries and key on name? Or do you have reason to believe this won't be necessary in intended applications?

PR #125, opened 2018-09-26
Generate tile types, site types, tilegrid, tileconn for entire part

litghost:
This also verifies that tileconn.json regenerates the same nodes.

This CL failing to generate some dead nodes, but I'm starting to clean up the branch.


litghost:
110 wires are still not getting including in tileconn, but this is better than it was.  I've added a mechanism to ignore wires.  Some of these wires are important, but they aren't important for MVP.

fuzzers/074-dump_all/create_node_tree.py
```diff
@@ -0,0 +1,261 @@
...
+import pickle
+import collections
+
+def build_node_index(fname):
+  node_index = {}
+  with open(fname, 'rb') as f:
+    f.seek(0, 2)
+    bytes = f.tell()
+    f.seek(0, 0)
+    with progressbar.ProgressBar(max_value=bytes) as bar:
```
mcmasterg: haven't seen this library before, will have to check it out. Fuzzers could use this instead of my diy solution

mcmasterg:
Can you post some example file outputs?

fuzzers/074-dump_all/reduce_site_types.py
```diff
@@ -0,0 +1,55 @@
+""" Reduce sites types to prototypes that are always correct.
```
mcmasterg: what is a prototype?

mcmasterg: I'm presuming a generic description of something as opposed to a specific instance

fuzzers/074-dump_all/ignored_wires.txt
```diff
@@ -0,0 +1,110 @@
+LIOI3_X0Y141/LIOI_I2GCLK_TOP1
```
mcmasterg: Is there a writeup somewhere on how you came up with this list?

litghost: It's literally just the wires that currently aren't defined via the tileconn.json.  I'm going to file an issue to figure out how to extract tileconn.json entries to describe them, but my goal right now is to start to integrate the updated database with VPR, rather than fix these 110 wires.

fuzzers/074-dump_all/reduce_site_types.py
litghost: Yes, exactly.

fuzzers/074-dump_all/ignored_wires.txt
litghost: Filed #129

litghost:
https://gist.github.com/litghost/b0288ed54f897ae67779a897d827e4fd

mcmasterg:
```
      "index_in_site": "4"
```
What is this? Why do we need it?

litghost:
> ```
>       "index_in_site": "4"
> ```
> What is this? Why do we need it?

We probably don't.  I was trying to extract different bits of data, and that was probably a less useful extraction.

mcmasterg:
Sort dictionaries to make canonical

mcmasterg:
```
    "AFFMUX:F7": {
      "from_pin": "F7",
      "to_pin": "OUT"
    },
    "AFFMUX:CARRY4_MUX": {
      "from_pin": "CARRY4_MUX",
      "to_pin": "OUT"
    },
```
IMHO it would be much more readable to aggregate muxes

mcmasterg:
sites are represented in a specific tile type. Any concern about duplicating them across types? Ex: SLICEL in CLBLM_L vs CLBLL_R

mcmasterg:
> > ```
> >       "index_in_site": "4"
> > ```
> > What is this? Why do we need it?
> 
> We probably don't. I was trying to extract different bits of data, and that was probably a less useful extraction.

Lets remove it then

mcmasterg:
```
"src_wire": "CLBLL_LL_CMUX",
"is_directional": "1",
"dst_wire": "CLBLL_LOGIC_OUTS22",
```
Are there non-directional in 7 series? Suggest sorting src/dst when this occurs to make the output canonical

litghost:
> sites are represented in a specific tile type. Any concern about duplicating them across types? Ex: SLICEL in CLBLM_L vs CLBLL_R

I've specifically written "reduce_site_types.py" to check if sites are common between tiles.  As far as I can tell from site pips and site pins, all sites of a specific type are common across all tiles.  We make this assumption when assembling CLBLM_* vs CLBLL_* tiles, and we now have strong evidence this is true.  It is possible that future parts won't behave this way, but I think that would be suprising.

litghost:
> ```
> "src_wire": "CLBLL_LL_CMUX",
> "is_directional": "1",
> "dst_wire": "CLBLL_LOGIC_OUTS22",
> ```
> Are there non-directional in 7 series? Suggest sorting src/dst when this occurs to make the output canonical

There are 66 pips that are not directional.

litghost:
> > ```
> > "src_wire": "CLBLL_LL_CMUX",
> > "is_directional": "1",
> > "dst_wire": "CLBLL_LOGIC_OUTS22",
> > ```
> > Are there non-directional in 7 series? Suggest sorting src/dst when this occurs to make the output canonical
> 
> There are 66 pips that are not directional.

It does appear that no PIP's are invertable, but that could be specific to artix7.

litghost:
> ```
>     "AFFMUX:F7": {
>       "from_pin": "F7",
>       "to_pin": "OUT"
>     },
>     "AFFMUX:CARRY4_MUX": {
>       "from_pin": "CARRY4_MUX",
>       "to_pin": "OUT"
>     },
> ```
> IMHO it would be much more readable to aggregate muxes

Humans shouldn't be working with these files anyways.  I don't believe any tool (vpr, vivado, nextpnr) are aware of mux's at the database level.  I see no reason to complicate data structure parsing to add muxes to then remove them again.

mithro:
Just FYI - if @mcmasterg is happy, then I'm happy. Please do merge without waiting / requiring input from me.

mcmasterg:
Sounds like index_in_site was removed. LGTM

PR #126, opened 2018-09-26
timing site fuzzer

mcmasterg:
This provides site timing models. In particular at this time I'm looking for comments if the output format is acceptable. Sample entry:

```
{
    "sites": {
        "SLICEL": {
            "AX:AFF:D": [
                75,
                59,
                205,
                164
            ]
    }
}
```
A single JSON file is produced for both site and interconnect based delays. See existing PR for details on interconnect delays. The process corner delay 4 tuple is the same format as in the interconnect section (fast_max fast_min slow_max slow_min in ps).

Site delays are organized by site type. Each key within a site type is a 3 tuple of site_pin:bel_name:bel_pin. No directionalality is implied. In the above example, its describing the delay going from the site AX input, through PRECYINIT, through PRECYINIT, and into PRECYINIT pin D. This is based on the assumption that there are not multiple feasible paths. I'm not sure if this is strictly true, but is it good enough for a first cut at a timing model?

If not, suggestions? More canonical would be to annotate it with a list of site wires that it passes through, but I'm not sure I can easily query these.

Of course, this is also a very simple timing model without factoring fanout, etc. At some point we also might want to consider those more carefully.

Known issues:
* Inter site delays (ex: LUT to FF within the same site) are not currently generated due to tcl script limitation (needs rejiggering). I will also have to rejigger the output JSON to allow these
* Think there was a performance regression during recent cleanup
* Other: tcl script may have bug selecting more wires than it should (see "#set wires [get_wires -of_objects $net -from $src_site_pin -to $dst_site_pin]")


mithro:
I'm not sure I like the `AX:AFF:D`. I think this is unclear what this means.

I would suggest something like `SLICEL.AX->SLICEL.AFF.D`. Then since this is inside a thing called `SLICEL` we can drop that leaving `AX->AFF.D`. This output kind of has a directionality however.

Maybe something like the following would be better? (Seems like its too verbose however?)
```json
{
  "sites": {
    "SLICEL": {
      "pins": { 
        "AX": {
          "bels": {
            "AFF": {
              "pins": {
                "D": {
                  "delays": {
                   "fast_max": 75,
                   "fast_min": 59,
                   "slow_min": 205,
                   "slow_max": 164,
                 }
               }
             }
           }
         }
       }
    }
}
```

I'm not overly fond of any of these suggestions.... Maybe just follow [SDF format](https://web.archive.org/web/20131104211358/http://www.eda.org/sdf/sdf_3.0.pdf) - that is what we will end up translating this to eventually.

mcmasterg:
Eh yeah I was considering the more verbose option, but I was afraid it would be too verbose. It is explicit though. Liking none of the options super well, I defaulted to something simple.

SLICEL.AX->SLICEL.AFF.D naming might be better. Its less specific on whether AX is a site or a pin, but seems easy enough to infer SITE.BEL.PIN vs SITE.PIN. I should be able to add directionality without too much effort. I think I'll go this direction unless I hear a better idea.

Another general issue I should mention. For tiles, I have a complete list of wires, and so know the entire possible delay space. For sites, I only know what has occurred so far. We'll have to be very careful importing these results to verify that all paths are covered. If someone has ideas though for a source of truth, I can add Nones when I don't have timing info.

mithro:
Happy to go with `SLICEL.AX->SLICEL.AFF.D`

@mcmasterg I think you should chat with @litghost about the path coverage issue? We don't want VPR preferring some paths because it thinks they have no delay...

mcmasterg:
How about slightly more terse
*External like (ie goes into our out of site): SLICEL:AX->AFF.D, SLICEL:AFF.Q->AQ
*Internal (ie BEL to BEL) like: SLICEL:LUT6.O2->AFF.D

mcmasterg:
This seems to pass small builds (oneblink) reasonably. I just launched a medium sized job (picorv32 N=1) and will then follow up with a full run (picorv32 N=8). Assuming this passes, timing fuzzer work is done for now. However, there are still two potential outstanding issues:
 * Excessive disk space usage: https://github.com/SymbiFlow/prjxray/issues/137
 * Excessive runtime (TBD)

Sample JSON from a simple project (ie poor coverage but shows the basic idea): https://gist.github.com/mcmasterg/920764c02853818ca3f357bccad0c54a

mcmasterg:
Run failed due to issue with passthrough LUTs. Made a small tweak, resuming the run. Hopefully will have a result EOD

Issue #128, opened 2018-09-27
Create small partial reconfiguration ROI for VPR testing


mcmasterg:
Current ROI is 50 by 8 CLBs (400 CLBs). We'd like something more in the ballpark of 10x10 for early testing

mcmasterg:
![small_roi](https://user-images.githubusercontent.com/18452276/46116665-0fe7f780-c1b2-11e8-9f80-a59591d69c6b.png)

@litghost : proposing 12 CLB rows by 8 CLB columns ROI => 96 CLB tiles. 2 inputs, 2 outputs, and clock. Is this good enough for a simple test? This can be done with pretty low work on my side. I might be able to increase the I/O with a small tweak if it would be helpful.

metadata from the attached image:

```
name node pin wire
clk CLK_HROW_TOP_R_X60Y130/CLK_HROW_CK_BUFHCLK_L0 F21 HCLK_VBRK_X34Y130/HCLK_VBRK_CK_BUFHCLK0
din[0] INT_R_X9Y100/NE2BEG3 G22 VBRK_X29Y106/VBRK_NE2A3
din[1] INT_R_X9Y103/NE2BEG3 G21 VBRK_X29Y109/VBRK_NE2A3
dout[0] INT_L_X10Y105/WW2BEG0 D21 VBRK_X29Y110/VBRK_WW2A0
dout[1] INT_L_X10Y108/WW2BEG0 E21 VBRK_X29Y113/VBRK_WW2A0
```

mcmasterg:
![small_roi](https://user-images.githubusercontent.com/18452276/46116849-e24f7e00-c1b2-11e8-9f9e-05f87597dd72.png)

@litghost : proposing 12 CLB rows by 8 CLB columns ROI => 96 CLB tiles. 2 inputs, 2 outputs, and clock. Is this good enough for a simple test? This can be done with pretty low work on my side. I might be able to increase the I/O with a small tweak if it would be helpful.

metadata from the attached image:

```
name node pin wire
clk CLK_HROW_TOP_R_X60Y130/CLK_HROW_CK_BUFHCLK_L0 F21 HCLK_VBRK_X34Y130/HCLK_VBRK_CK_BUFHCLK0
din[0] INT_R_X9Y100/NE2BEG3 G22 VBRK_X29Y106/VBRK_NE2A3
din[1] INT_R_X9Y103/NE2BEG3 G21 VBRK_X29Y109/VBRK_NE2A3
[...]

PR #130, opened 2018-09-27
roi_harness: small project option

mcmasterg:
For https://github.com/SymbiFlow/prjxray/issues/128

We could make this more flexible, but I think this is good enough. This code should become obsolete anyway once we are more established

Signed-off-by: John McMaster <johndmcmaster@gmail.com>

### DIFF
--- a/minitests/roi_harness/.gitignore
+++ b/minitests/roi_harness/.gitignore
@@ -11,3 +11,5 @@
 /*.bit
 /*.bin
 /*.frm
+
+build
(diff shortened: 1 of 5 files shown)


mcmasterg:
I've updated the default makefile target to build BASYS3-SWBUT, the configuration we did for the last demo (not arty 7 as I think I mistakenly said before).

See runme.sh if you want to rebuild yourself. You must setup your environment correctly to build for 35T and such.

PR #133, opened 2018-09-27
Add methods to prjxray library.

litghost:
All data from #125 is now exported from the prjxray library.  A bug was found in how site reduction was done, so some sites in some tile types don't round trip.  This can be fixed in a future PR.

### DIFF
--- a/tools/quick_test.py
+++ b/tools/quick_test.py
@@ -11,14 +11,46 @@ def quick_test(db_root):
     tile_types_in_grid = set(
         g.gridinfo_at_loc(loc).tile_type for loc in g.tile_locations())
     tile_types_in_db = set(db.get_tile_types())
+    site_types = set(db.get_site_types())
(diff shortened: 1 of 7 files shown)


tools/quick_test.py
```diff
@@ -11,14 +11,46 @@ def quick_test(db_root):
...
+
+    for loc in g.tile_locations():
+        gridinfo = g.gridinfo_at_loc(loc)
+        assert gridinfo.tile_type in db.get_tile_types()
+        for site_name, site_type in gridinfo.sites.items():
+            assert site_type in site_types
+
+        tile = db.get_tile_type(gridinfo.tile_type)
+
+        # FIXME: The way sites are named in Tile.get_instance_sites is broken
```
mcmasterg: Should there be an issue for this?

litghost: #134

PR #138, opened 2018-10-05
Additional library behaviors after integration.

litghost:
- Add caching to db.py for tile types
- Add wire info to tile types to allow easy pip/site access from wire names


mithro:
LGTM.

We should log a bug to add docstrings to db.py file....

litghost:
Ya, agreed, filed #141

PR #139, opened 2018-10-05
Hack to make new tilegrid play nice with non-updated utilities

litghost:


mithro:
LGTM.

Can you add an issue to remove this hack and under what conditions we can remove it?

PR #148, opened 2018-10-11
Changes from trying to generate updated database

mithro:
* `074-dump_all` - Don't fail if output dir doesn't exist.
 * Add Python virtualenv to make it easier to install all the Python requirements.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+/env/
 /build/
 .Xil
 # Ignore database directories _except_ for their settings
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,11 @@
 CLANG_FORMAT ?= clang-format
 
-.PHONY: database format clean
+.PHONY: database format clean env
+
+env:
+	virtualenv --python=python3 env
+	. env/bin/activate; pip install -r requirements.txt
+	ln -sf $(PWD)/prjxray env/lib/python3.*/site-packages/
 
 build:
 	git submodule update --init --recursive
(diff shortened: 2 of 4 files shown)


.gitignore
```diff
@@ -1,3 +1,4 @@
+env
```
galv: Change to `/env/` to avoid capturing any non-directory env files that aren't at the top level.

Makefile
```diff
@@ -2,6 +2,11 @@ CLANG_FORMAT ?= clang-format
 
 .PHONY: database format clean
 
+env:
```
galv: Make this rule .PHONY? Up to you. Note that it won't work if something breaks before the last line. Alternatively, make a directory like `env-in-progress` and `mv` it to a directory called `env`.

Makefile
```diff
@@ -2,6 +2,11 @@ CLANG_FORMAT ?= clang-format
...
+	virtualenv --python=python3 env
+	. env/bin/activate; pip install -r requirements.txt
+	ln -s $(PWD)/prjxray env/lib/python3.*/site-packages/
```
galv: Change to `ln -sf` to make this make rule idempotent.

.gitignore
mithro: Done.

Makefile
mithro: Done.

Makefile
mithro: Done.

mithro:
Hi @galv,

Thanks for helping review this code, but who are you?

Tim @mithro Ansell

galv:
Hi @mithro. I'm a friend of @mcmasterg. I saw that the currently uploaded database was months old recently, but then I saw your PR which sounds like it'll hopefully update, so I figured I'd pitch in a little bit.

mithro:
@galv Ahh cool! Yes, there should be a database update soon.

PR #149, opened 2018-10-11
Fixes to dump all

litghost:


mithro:
These were merged as part of #148

PR #158, opened 2018-10-18
lib: xilinx: catch exceptions by reference

mgielda:
Signed-off-by: Michael Gielda <mgielda@antmicro.com>


mcmasterg:
Reference: "...we've found small bug preventing the tools to be compiled with GCC 8"

mcmasterg:
Thanks for the PR!

PR #167, opened 2018-10-19
Refactor FASM handling to use new FASM library.

litghost:
Also isolated genenv calls to prjxray/utils.py function calls, and default arguments to utils tooling.  If env vars are not set, command line flags can be used instead.

### DIFF
--- a/experiments/clbpips/generate.py
+++ b/experiments/clbpips/generate.py
@@ -1,10 +1,9 @@
 #!/usr/bin/env python3
 
-import sys, re
-
 from prjxray.segmaker import Segmaker
+from prjxray import util
 
-segmk = Segmaker("design.bits")
+segmk = Segmaker(util.get_db_root(), "design.bits")
 
 print("Loading tags from design.txt.")
 with open("design.txt", "r") as f:
--- a/experiments/clbpips/generate.py
+++ b/experiments/clbpips/generate.py
@@ -1,9 +1,8 @@
 #!/usr/bin/env python3
 
 from prjxray.segmaker import Segmaker
-from prjxray import util
 
-segmk = Segmaker(util.get_db_root(), "design.bits")
+segmk = Segmaker("design.bits")
 
 print("Loading tags from design.txt.")
 with open("design.txt", "r") as f:
--- a/minitests/roi_harness/fasm2bit.sh
+++ b/minitests/roi_harness/fasm2bit.sh
@@ -27,7 +27,7 @@ echo "Design .fasm: $fasm_in"
 echo "Harness .bit: $bit_in"
 echo "Out .bit: $bit_out"
 
-${XRAY_DIR}/tools/fasm2frame.py $fasm_in roi_partial.frm
+${XRAY_DIR}/utils/fasm2frames.py --sparse $fasm_in roi_partial.frm
 
 ${XRAY_TOOLS_DIR}/xc7patch \
 	--part_name ${XRAY_PART} \
--- a/minitests/roi_harness/fasm2bit.sh
+++ b/minitests/roi_harness/fasm2bit.sh
@@ -27,7 +27,7 @@ echo "Design .fasm: $fasm_in"
 echo "Harness .bit: $bit_in"
 echo "Out .bit: $bit_out"
 
-${XRAY_DIR}/utils/fasm2frames.py --sparse $fasm_in roi_partial.frm
+${XRAY_FASM2FRAMES} --sparse $fasm_in roi_partial.frm
 
 ${XRAY_TOOLS_DIR}/xc7patch \
 	--part_name ${XRAY_PART} \
--- a/minitests/roi_harness/runme.sh
+++ b/minitests/roi_harness/runme.sh
@@ -57,8 +57,8 @@ test -z "$(fgrep CRITICAL vivado.log)"
 
 ${XRAY_BITREAD} -F $XRAY_ROI_FRAMES -o design.bits -z -y design.bit
 ${XRAY_SEGPRINT} -zd design.bits >design.segp
-${XRAY_DIR}/tools/segprint2fasm.py design.segp design.fasm
-${XRAY_DIR}/tools/fasm2frame.py design.fasm design.frm
+${XRAY_DIR}/utils/bits2fasm.py --verbose design.bits > design.fasm
+${XRAY_DIR}/utils/fasm2frames.py design.fasm design.frm
 # Hack to get around weird clock error related to clk net not found
 # Remove following lines:
 #set_property CLOCK_DEDICATED_ROUTE FALSE [get_nets clk_IBUF]
--- a/prjxray/fasm_assembler.py
+++ b/prjxray/fasm_assembler.py
@@ -1,0 +1,140 @@
+import fasm
+from prjxray import bitstream
+from prjxray import grid
+
+class FasmLookupError(Exception):
+    pass
+
+class FasmInconsistentBits(Exception):
+    pass
+
+# How many 32-bit words for frame in a 7-series bitstream?
+FRAME_WORD_COUNT = 101
+
+def init_frame_at_address(frames, addr):
+    '''Set given frame to 0 if not initialized '''
+    if not addr in frames:
+        frames[addr] = [0 for _i in range(FRAME_WORD_COUNT)]
+
+class FasmAssembler(object):
+    def __init__(self, db):
+        self.db = db
+        self.grid = db.grid()
+
+        self.frames = {}
+        self.frames_line = {}
+
+    def get_frames(self, sparse=False):
+        if not sparse:
+            frames = self.frames_init()
+        else:
+            frames = {}
+
+        for (frame_addr, word_addr, bit_index), is_set in self.frames.items():
+            init_frame_at_address(frames, frame_addr)
+
+            if is_set:
+                frames[frame_addr][word_addr] |= 1 << bit_index
+
+        return frames
+
+    def frames_init(self):
+        '''Set all frames to 0'''
+        frames = {}
+
+        for bits_info in self.grid.iter_all_frames():
+            for coli in range(bits_info.bits.frames):
+                init_frame_at_address(frames, bits_info.bits.base_address + coli)
+
+        return frames
+
+    def frame_set(self, frame_addr, word_addr, bit_index, line):
+        '''Set given bit in given frame address and word'''
+        assert bit_index is not None
+
+        key = (frame_addr, word_addr, bit_index)
+        if key in self.frames:
+            if self.frames[key] != 1:
+                raise FasmInconsistentBits(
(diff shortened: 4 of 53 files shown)


experiments/clbpips/generate.py
```diff
@@ -1,10 +1,9 @@
 #!/usr/bin/env python3
 
-import sys, re
-
 from prjxray.segmaker import Segmaker
+from prjxray import util
 
-segmk = Segmaker("design.bits")
+segmk = Segmaker(util.get_db_root(), "design.bits")
```
mcmasterg: every fuzzer is expected to pass "util.get_db_root()". This should be a settable default

minitests/roi_harness/fasm2bit.sh
```diff
@@ -27,7 +27,7 @@ echo "Design .fasm: $fasm_in"
 echo "Harness .bit: $bit_in"
 echo "Out .bit: $bit_out"
 
-${XRAY_DIR}/tools/fasm2frame.py $fasm_in roi_partial.frm
```
mcmasterg: there is an env var for this now: XRAY_FASM2FRAMES

litghost: Sure

minitests/roi_harness/runme.sh
```diff
@@ -57,8 +57,8 @@ test -z "$(fgrep CRITICAL vivado.log)"
 
 ${XRAY_BITREAD} -F $XRAY_ROI_FRAMES -o design.bits -z -y design.bit
 ${XRAY_SEGPRINT} -zd design.bits >design.segp
-${XRAY_DIR}/tools/segprint2fasm.py design.segp design.fasm
```
mcmasterg: convert to env vars XRAY_BITS2FASM, XRAY_FASM2FRAMES

prjxray/fasm_assembler.py
```diff
@@ -0,0 +1,152 @@
+import fasm
```
mcmasterg: wasn't fasm stuff getting moved to an external library that called prjxray lib functions?

prjxray/fasm_assembler.py
```diff
@@ -0,0 +1,152 @@
...
+class FasmLookupError(Exception):
+    pass
+
+
+class FasmInconsistentBits(Exception):
+    pass
+
+
+# How many 32-bit words for frame in a 7-series bitstream?
+FRAME_WORD_COUNT = 101
```
mcmasterg: we should probably move this constant to a central location

prjxray/fasm_assembler.py
litghost: I had planned it the other way.  The FASM library handles parsing and emission of FASM itself.  Each target handles how to take parsed FASM and do something with it.  I see a lot more variation in each target.  I guess we could move to a plugin model, but that is tricky to design, I don't think I'd want to try to get that right on the first go.

prjxray/fasm_assembler.py
litghost: Ya, I moved the word size constant, I'll move this too.

experiments/clbpips/generate.py
litghost: What is meant by a "settable default"?  You mean a default argument?

I'm not sure why making the database root implicit to Segmaker is buying us, besides saving some characters.  Environment variable usage should be a decision made by the program, rather than the library, and I think keeping the environment variable usage to as few places is important.

mcmasterg: Making the code more readable is important to me. Its a very reasonable default. The program still has the option to override if it wants to

prjxray/fasm_assembler.py
```diff
@@ -0,0 +1,148 @@
...
+                    ))
+            return
+
+        self.frames[key] = 0
+        self.frames_line[key] = line
+
+    def enable_feature(self, tile, feature, address, line):
+        gridinfo = self.grid.gridinfo_at_tilename(tile)
+
+        # TODO: How to determine if the feature targets BLOCK_RAM segment type?
```
mcmasterg: not sure what the specific issue is here, but feel free to ping me if you need additional info on BRAM

prjxray/grid.py
```diff
@@ -1,7 +1,20 @@
 from collections import namedtuple
+import enum
+from prjxray import segment_map
+
+
+class SegmentType(enum.Enum):
```
mcmasterg: These aren't segment types. Segment type has a very specific meaning in tilegrid. Rather these are "Block Type". See UG470 pg 110: https://www.xilinx.com/support/documentation/user_guides/ug470_7Series_Config.pdf

prjxray/fasm_assembler.py
litghost: The original fasm2frames.py didn't seem to handle this either.  In the current database format, how do we know if a FASM line is targeting the CLB_IO_CLK section or the BLOCK_RAM section?  Maybe I missed the database file or something.  We could special case BLOCK_RAM FASM features as simply
BRAM_X<x>_Y<y>.BLOCK_RAM
etc.

prjxray/grid.py
litghost: I was having trouble naming this enum, I'll update it.

prjxray/fasm_assembler.py
litghost: The library is pretty boring.  It's pretty much string -> AST, AST -> string, and partial canonization.

prjxray/fasm_assembler.py
mcmasterg: Ah yeah, this isn't fully hammered out yet, but I'm assuming something like that symbols don't overlap between the memory regions. ie something under the hood tags it as coming from a certain memory region, and fasm knows which region to map that to without the user specifying explicitly

litghost: Hence why I left that as a TODO.  We'll need to add FASM support for BRAM initialization eventually, but I think we can leave that out of scope for this PR.

prjxray/roi.py
```diff
@@ -0,0 +1,48 @@
+import json
+
+
+class Roi(object):
```
mcmasterg: would be good to clarify intent of using this class as a x-ray vs Xilinx ROI

prjxray/roi.py
```diff
@@ -0,0 +1,48 @@
...
+    def __init__(self, tilegrid_file, x1, x2, y1, y2):
+        self.tilegrid_file = tilegrid_file
+        self.tilegrid = None
+        self.x1 = x1
+        self.x2 = x2
+        self.y1 = y1
+        self.y2 = y2
+
+    def tile_in_roi(self, tilej):
```
mcmasterg: instead of using the JSON directly, is there a Tile object that should be used instead?

utils/bits2fasm.py
```diff
@@ -1,309 +1,53 @@
...
+            canonical=canonical))
 
 
 def main():
     import argparse
 
-    # XXX: tool still works, but not well
-    # need to eliminate segments entirely
     parser = argparse.ArgumentParser(
-        description='XXX: does not print all data?')
```
mcmasterg: this was actually a bug fix comment, but I think this was fixed...

utils/fasm2pips.py
```diff
@@ -0,0 +1,51 @@
+#!/usr/bin/env python3
```
mcmasterg: what is this utility for?

mcmasterg: oh for highlighting in vivado

utils/fasm2fasm.py
```diff
@@ -0,0 +1,87 @@
+#!/usr/bin/env python3
+'''
+Takes input FASM and outputs optionally canonical FASM.
```
mcmasterg: if you don't make it canonical, what is the tool doing? is this more testing that the API works? Seems if its to make it canonical, the tool should just do that, and do it well

mcmasterg: and API test should be a separate utility

utils/fasm2fasm.py
```diff
@@ -0,0 +1,87 @@
...
+
+            address = 0
+            if feature.start is not None:
+                address = feature.start
+
+            feature_name = '{}.{}'.format(
+                gridinfo.tile_type, '.'.join(parts[1:]))
+
+            # Convert feature to bits.  If no bits are set, feature is
+            # psuedo pip, and should not be output from canonical FASM.
```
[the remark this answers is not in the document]
litghost: For a couple reasons.  The first is that pseudo pip's in some cases better express the intent of the generating tool.  For example all connections to VCC_WIRE are ppips with the values of "default".  So the FASM line:

INT_L_X0Y0.BYP_ALT2.VCC_WIRE

is an explicit way to say "connect VCC_WIRE to BYP_ALT2", even if the absence of the FASM line would generate an equivalent bitstream.

The second is reason around consistency for symbiflow/VPR routing emission.  The routing network contains many pip's that are implicit (e.g. "always", example CLBLL_L.CLBLL_L_AX.CLBLL_BYP0) or default (for example INT_L.BYP_ALT2.VCC_WIRE).  When patching the rr graph, if psuedo pips are acceptable FASM, then we don't need to filter those annotations.  This is connected with the third reason.

The third reason is debugging, and why the fasm2pips.py tool exists.  fasm2pips.py provides a quick way to examine VPR routing output in Vivado.  Including psuedo pips in FASM allows a couple extra pips to be output in Vivado without requiring logic to follow the nets to enable them.

mcmasterg:
think I've added most of my comments

utils/fasm2pips.py
litghost: It converts a FASM file into a Vivado tcl command that highlights the enabled pips from the FASM file.  Useful for sanity checking VPR routing output, etc.  I'm considering adding more logic to highlight upstream and downstream nodes from the pips, along with potentially highlight routing within the CLB sites.  For now it just highlights enabled routing pips.

litghost: I'll add a docstring.

utils/fasm2fasm.py
litghost: Think of this tool as a pretty printer.  I'm planning on adding modes for bit fusion, etc.  Maybe fasm_pretty.py or fasm_pretty_print.py or fasm_pprint.py is a better name?

utils/fasm2fasm.py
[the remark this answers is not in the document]
mcmasterg: Okay, I'll mull over this. My main concern is that if we are defining FASM to be purely about the bitstream, if we add non-existent features in, we are diverging from that

litghost:
Comments addressed

utils/fasm2fasm.py
mcmasterg: fasm_pretty or fasm_pprint sounds good to me

mcmasterg:
Ran into a few minor issues during testing, they are fixed now

Force merging since Github checkers are broken

Issue #169, opened 2018-10-22
Rethink ROI usage


mcmasterg:
We have three things right now that the ROI are used for:
- The area we target for VPR PnR testing
- The area that a developer wants to run for quick testing
- The area that we want the database to contain, which is now arguably the full chip area (this is why 074 exists)

We should rethink these in order to clean up the code. Proposal is something like:
- Expand "official" ROI to contain all tile addresses. However, keep the SLICEs and such limited to just the VPR area for now. This will make it so the database contains all tiles, but only addresses for the VPR ROI
- Have a secondary ROI for VPR testing. This is also probably a good area to use for quick test development

litghost:
I think this makes sense to me.  This work is directly tied with the change of the design.txt output from the roi_harness to support configurable grid roi's for VPR.  Once design.txt (e.g. design.json) also describes what tiles VPR should use, we can totally decouple these 3 things.

JohnDMcMaster:
this was fixed in a recent PR

PR #184, opened 2018-10-23
Create design.json that describes roi harness boundry,

litghost:
This enables complete separation of the VPR ROI per #169.


minitests/roi_harness/runme.tcl
```diff
@@ -360,6 +360,48 @@ proc node2wire {node} {
     return $wire
 }
 
+proc write_grid_roi {fp} {
+    foreach {slices} [split "$::env(XRAY_ROI)" " "] {
```
mcmasterg: technically sites, not slices

minitests/roi_harness/runme.tcl
```diff
@@ -360,6 +360,48 @@ proc node2wire {node} {
     return $wire
 }
 
+proc write_grid_roi {fp} {
```
mcmasterg: This completely ignores XRAY_ROI_GRID in favor of XRAY_ROI. Wouldn't it be better to go the other direction and assume that we want all primitives within XRAY_ROI_GRID?

litghost: The current runme.tcl code doesn't obey XRAY_ROI_GRID at all, so I'm not sure how we would leverage XRAY_ROI_GRID.

mcmasterg: Hmm but isn't our environment guaranteeing they are related?

litghost: Sure!  Done

PR #188, opened 2018-10-24
Make 005-tilegrid emit entire part, and have 074-dump_all not overwri…

### LINKED ISSUE #170: 074 shouldn't override tilegrid.json
074 overrides 005's tilegrid.json, making build products ambiguous. Would be better to name it tilegrid-full.json or something like that

Fixing https://github.com/SymbiFlow/prjxray/issues/169 may make this issue obsolete

litghost:
…te 005-tilegrid.

Fixes #170


mcmasterg:
How does this interact with the fuzzers that iterate over tilegrid for their CLBs, BRAM, etc?

litghost:
I've run 010-lutinit, 011-ffconfig and 012-clbn5ffmux without error.  I'll start a complete run now, but any fuzzer in particular you think will be affected?

mcmasterg:
I'm mostly worried that it will generate .v for all of them, but only solve for things in the ROI. I'll try to run a local test to verify this

PR #203, opened 2018-10-29
WIP: Attempt to add IOB segments to tilegrid

litghost:


fuzzers/005-tilegrid/generate.py
```diff
@@ -491,6 +547,20 @@ def db_add_bits(database, segments):
                     ("DSP", "CLB_IO_CLK"): (28, 2, 10),
                     ("INT_INTERFACE", "CLB_IO_CLK"): (28, 2, None),
                     ("BRAM_INT_INTERFACE", "CLB_IO_CLK"): (28, 2, None),
+                    # TODO: These are all guesses copied from CLBLL.
```
mcmasterg: you should be able to use the scraper tool to confirm the number of frames

fuzzers/005-tilegrid/generate.py
```diff
@@ -110,25 +116,28 @@ def make_tile_baseaddrs(tiles, site_baseaddr, verbose=False):
     # Look up a base address by tile name
     tile_baseaddrs = dict()
 
-    verbose and print('')
-    verbose and print('%u tiles' % len(tiles))
+    verbose and print('', file=sys.stderr)
```
mcmasterg: I've never been a big fan of stderr personally.  FWIW there is a output file arg if you don't want the JSON to flood the screen

fuzzers/005-tilegrid/top.v
```diff
@@ -1,10 +1,13 @@
 //Need at least one LUT per frame base address we want
 `define N_LUT 100
 `define N_BRAM 8
+`define N_IOB 250
```
mcmasterg: are you trying to solve every IOB in our part? Why so many?

litghost: I originally misunderstood what was required, we only need one IOB in each 50 column to get the base address. I can simplify this.

fuzzers/005-tilegrid/generate.py
mcmasterg: Scraper tool indicates this might not be right (IOB_X0Y0 width: 42):

```
$ cat build/design_IOB_X0Y0.delta
+bit_00400027_000_29
$ $XRAY_BLOCKWIDTH build/design_IOB_X0Y0.bit |grep 0x00400000
0x00400000: 0x2A
```

But it might be just as well to commit for now and we'll sort it out later

mcmasterg:
FYI looking at IOBs more seriously now

PR #209, opened 2018-10-30
BRAM: SDP minitest

mcmasterg:


mcmasterg:
Also a small segprint cleanup

Issue #212, opened 2018-10-31
BRAM fuzzer placer error


mcmasterg:
When I moved from my laptop to my desktop I got:
```
ERROR: [Place 30-99] Placer failed with error: 'Failed to move lutram to legal location'
```

PR #213, opened 2018-11-01
fuzzers: don't use EXCLUDE_PLACEMENT on ROI

mcmasterg:
Fixes: https://github.com/SymbiFlow/prjxray/issues/212

A recent change got the ROI X/Y out of sync with the ROI primitives. This caused issues for many fuzzers which used EXCLUDE_PLACEMENT to create a sharp boundary between the ROI and the rest of the FPGA. This is problematic because the site ROI was expanded to include most of the chip, and this precluded placing harness test logic in the top module.

Really, we need to separate out the ROI that we are using for tilegrid generation vs fuzzers. Or, if the idea is to no longer use an ROI for tilegrid, we should just remove the ROI entirely from that fuzzer

### DIFF
--- a/fuzzers/007-timing/minitest/test_unique/generate.tcl
+++ b/fuzzers/007-timing/minitest/test_unique/generate.tcl
@@ -19,7 +19,6 @@ proc build_design {} {
     puts "pblocking"
     create_pblock roi
     set roipb [get_pblocks roi]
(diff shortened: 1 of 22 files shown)


mcmasterg:
FYI roi_harness still has this set, since it may be more critical for that

Issue #218, opened 2018-11-01
Segprint not printing all tiles


mcmasterg:
Originally segmaker used the segment info in the DB to print literal segments as defined by tilegrid.json. However, it was reworked recently to print on a tile by tile basis with the intention that it would print the selected tile plus info from other tiles intersecting with the current tile. However, only the current file seems to be printed right now.

Other: segprint is probably outdated name and should be renamed tileprint

PR #220, opened 2018-11-01
Segprint: print related tiles

mcmasterg:
Fixes https://github.com/SymbiFlow/prjxray/issues/218

### DIFF
--- a/utils/segprint.py
+++ b/utils/segprint.py
@@ -57,7 +57,7 @@ def get_database(db, tile_type, verbose=False):
             # second part forms a tuple refereced in sets
             return (isset, (int(frame, 10), int(word, 10)))
 
-        if parts[1] == 'always':
+        if parts[1] == 'always' or parts[1] == 'hint':
             tagbits = []
         else:
             tagbits = [parsetag(x) for x in parts[1:]]
--- a/utils/segprint.py
+++ b/utils/segprint.py
@@ -1,7 +1,9 @@
 #!/usr/bin/env python3
 '''
 Take raw .bits files and decode them to higher level functionality
-This output is intended for debugging and not directly related to FASM
+
+A segment as currently used is defined as a tile type + a memory region
+Ex: BRAM_L_X6Y100:CLB_IO_CLK
 '''
 
 import sys, os, json, re
@@ -20,7 +22,6 @@ segbitsdb = dict()
 # int and sites are loaded together so that bit coverage can be checked together
 # however, as currently written, each segment is essentially printed twice
 def process_db(db, tile_type, process, verbose):
-    print(db.get_tile_types())
     ttdb = db.get_tile_type(tile_type)
 
     fns = [ttdb.tile_dbs.segbits, ttdb.tile_dbs.ppips]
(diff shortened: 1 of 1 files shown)


utils/segprint.py
```diff
@@ -20,7 +22,6 @@ class NoDB(Exception):
 # int and sites are loaded together so that bit coverage can be checked together
 # however, as currently written, each segment is essentially printed twice
 def process_db(db, tile_type, process, verbose):
-    print(db.get_tile_types())
     ttdb = db.get_tile_type(tile_type)
 
     fns = [ttdb.tile_dbs.segbits, ttdb.tile_dbs.ppips]
```
litghost: Could use TileSegbits via db.get_tile_segbits? 
https://github.com/SymbiFlow/prjxray/blob/master/prjxray/db.py#L138

prjxray/db.py
```diff
@@ -25,14 +25,15 @@ def get_available_databases(prjxray_root):
 
 
 class Database(object):
-    def __init__(self, db_root):
```
litghost: The merge resolution is backwards.  Tools should always specify the db_root, rather than relying on implicit env vars.

mcmasterg: I disagree. I want convenient tools

mcmasterg: The alternate solution would be a util class to make the DB. Does that work for you?

litghost: The tools can default to the env vars, but need to work in the absence of those env vars.  Env vars need to be kept in the roots of tools as much as possible, especially for tools that will be used extensively outside of database generation.

utils/segprint.py
```diff
@@ -222,18 +398,25 @@ def tile_segnames(tiles):
...
         segnames,
         omit_empty_segs=False,
         flag_unknown_bits=False,
         flag_decode_emit=False,
         flag_decode_omit=False,
         verbose=False):
-    tiles, segments = mk_grid(db_root)
-    db = prjxraydb.Database(db_root)
-
+    db = prjxraydb.Database()
```
litghost: Both mk_grid and Database rely on a common db_root, this should be made explicit.

prjxray/db.py
litghost: I'm fine with a util function to make the DB, but in this case the tool needs to be able to take the db_root from the command line, so the util function is redundant.

mcmasterg: Ah I see. I misunderstood what the merge conflict was. Let me re-review

utils/segprint.py
```diff
@@ -231,9 +415,9 @@ def run(
         flag_decode_emit=False,
         flag_decode_omit=False,
         verbose=False):
-    tiles, segments = mk_grid(db_root)
     db = prjxraydb.Database(db_root)
-
+    tiles = load_tiles()
```
litghost: Have load_tiles take db_root, and I think we are good.

mcmasterg:
@litghost sorry, misunderstood what the issue was. I've more carefully reviewed the merge issues and submitted a new patch. Please review and let me know if this works for you

utils/segprint.py
litghost: Sounds good to me.

litghost:
@mcmasterg No problem, left a new comment, and I think we are good to go.

Issue #146, opened 2018-10-10
Fuzzers should diff entire bitstream


mcmasterg:
For size and/or simplicity we originally were filtering out to only dump a smallish area of the bitstream. However, this is causing issues for things like the BRAM fuzzer. Its also not saving us that much disk space anymore. For example, my test bitstream increased from 208 KB to 212 KB, basically negligible.

However, we may add additional aliased bits if we add this. So we'll have to re-run the fuzzers and verify everything is still good.

```
mcmaster@xc2064:~/prjxray/fuzzers/005-tilegrid/specimen_001$ /home/mcmaster/prjxray/build/tools/bitread --part_file /home/mcmaster/prjxray/database/kintex7/xc7k70tfbg676-2.yaml -F 0x00400100:0x004006ff -o design_SLICE_X9Y50.bits -z -y design_RAMB36_X0Y10.bit
Bitstream size: 3157171 bytes
Config size: 789255 words
Number of configuration frames: 7432
DONE
mcmaster@xc2064:~/prjxray/fuzzers/005-tilegrid/specimen_001$ /home/mcmaster/prjxray/build/tools/bitread --part_file /home/mcmaster/prjxray/database/kintex7/xc7k70tfbg676-2.yaml -F 0x00000000:0xffffffff -o design_SLICE_X9Y50_full.bits -z -y design_RAMB36_X0Y10.bit
Bitstream size: 3157171 bytes
Config size: 789255 words
Number of configuration frames: 7432
DONE
mcmaster@xc2064:~/prjxray/fuzzers/005-tilegrid/tmp$ du -hs design_SLICE_X9Y50*.bits
208K    design_SLICE_X9Y50.bits
212K    design_SLICE_X9Y50_full.bits
```

cliffordwolf:
> However, we may add additional aliased bits if we add this.

jfyi: This is the main reason why the current code is only looking at certain sections of the bit stream.

mcmasterg:
A7 is diffing entire bitstream now. We have some unrelated fuzzer regressions I'm working through, will look into this more as part of that

mcmasterg:
Filed https://github.com/SymbiFlow/prjxray/issues/230 Think issue is not related to this one

Issue #200, opened 2018-10-27
Fuzzers are broken with the Kintex 7 settings


mithro:
Changes to the fuzzers mean they no longer work with the [Kintex 7 settings](https://github.com/SymbiFlow/prjxray/blob/master/database/kintex7/settings.sh).

So far the following fuzzers work;

 - [ ] 001-part-yaml
 - [ ] 005-tilegrid
 - [ ] 007-timing
 - [ ] 010-lutinit
 - [ ] 011-ffconfig
 - [ ] 012-clbn5ffmux
 - [ ] 013-clbncy0
 - [ ] 014-ffsrcemux
 - [ ] 015-clbnffmux
 - [ ] 016-clbnoutmux
 - [ ] 017-clbprecyinit
 - [ ] 018-clbram
 - [ ] 019-ndi1mux
 - [ ] 050-intpips
 - [ ] 051-imuxlout
 - [ ] 052-clkin
 - [ ] 053-ctrlin
 - [ ] 054-gfan
 - [ ] 055-gnd
 - [ ] 056-rempips
 - [ ] 057-bipips
 - [ ] 058-hclkpips
 - [ ] 071-ppips
 - [ ] 072-ordered_wires
 - [ ] 073-get_counts
 - [ ] 074-dump_all
 - [ ] 100-dsp-mskpat
 - [ ] 102-bram-data

mithro:
The current failing fuzzer is [`005-tilegrid`](https://github.com/SymbiFlow/prjxray/tree/master/fuzzers/005-tilegrid) it fails with;
```
+ for x in design*.bit
+ /big/github/SymbiFlow/prjxray-kintex/build/tools/bitread --part_file /big/github/SymbiFlow/prjxray-kintex/database/kintex7/xc7k70tfbg676-2.yaml -F 0x00400100:0x004006ff -o design_SLICE_X6Y50.bits -z -y design_SLICE_X6Y50.bit
Bitstream size: 3157171 bytes
Config size: 789255 words
Number of configuration frames: 7432
DONE
+ for x in design*.bit
+ /big/github/SymbiFlow/prjxray-kintex/build/tools/bitread --part_file /big/github/SymbiFlow/prjxray-kintex/database/kintex7/xc7k70tfbg676-2.yaml -F 0x00400100:0x004006ff -o design_SLICE_X8Y50.bits -z -y design_SLICE_X8Y50.bit
Bitstream size: 3157171 bytes
Config size: 789255 words
[...]

PR #229, opened 2018-11-06
k7: use all bitstream bits, format settings.sh alike

mcmasterg:
Fixes part of https://github.com/SymbiFlow/prjxray/issues/200

Fixes https://github.com/SymbiFlow/prjxray/issues/146


mithro:
LGTM!

mcmasterg:
I was looking at the results closer and there is actually an issue with this. It doesn't technically error, but its not correct either. Will follow up.

```
Sample LUT line
Writing bitstream ./design_SLICE_X18Y50.bit...
The BRAM line:
Writing bitstream ./design_.bit...
```

PR #234, opened 2018-11-07
dbfixup: per project instead of globally

mcmasterg:
Changes:
- dbfixup: split into more functions, add comments
- dbfixup: allow taking an external zero bit definition. External masks are not supported at this time
- dbfixup: add --db_root arg
- fuzzers: use new dbfixup flags to fixup local files more precisely instead of trying to apply globally
- fuzzers: consolidate some build scripts, move products to build dir

Context:

dbfixup aggregates multi bit values after segmatch solves for them. Right now this is done globally after data is pushed into the global DB directory. This means that all of the fixups are smashed together and is particularly problematic as currently written because the same fixups are applied to all included tile types. This was fine when all the types needed were CLB like, but this is now conflicting with BRAM.

Rather than expand the current scheme, this restructures the code to work per fuzzer before pushing into the DB. Old interconnect behavior is preserved right now with --clb-int and called in all projects that relied on it


mcmasterg:
The following fuzzers were checked and seem to still run.

Checked:
 - 001-part-yaml
 - 005-tilegrid
 - 010-lutinit
 - 011-ffconfig
 - 012-clbn5ffmux
 - 013-clbncy0
 - 014-ffsrcemux
 - 015-clbnffmux
 - 016-clbnoutmux
 - 017-clbprecyinit
 - 018-clbram
 - 019-ndi1mux
 - 025-bram-config
 - 026-bram-data

mcmasterg:
We need to do a full DB run to verify this is fully working, but the tests I was able to run seems okay. I'm merging in the interest in moving things with BRAM forward

PR #268, opened 2018-11-15
Do not insert non-existent frames.

litghost:


JohnDMcMaster:
I'm not familiar with the core issue here, but I have no objection to merging this if you think its the right thing

JohnDMcMaster:
FYI make format is failing due to non-pinned clang-format version

PR #281, opened 2018-12-05
tilegrid misc IOB fixes

mcmasterg:
Misc fixes from IOB fuzzer testing

NOTE: this is probably still not working (fuzzer is giving tautology, addresses might be wrong). But generally in the right direction


mcmasterg:
the tautology is probably caused by IOB33S and IOB33M getting merged in segmaker. I need to make them separate symbols like we do for SLICEL vs SLICEM

PR #300, opened 2018-12-10
Improving assert error messages.

mithro:


mcmasterg:
FYI I'm making a lot of changes right now in the checkdb.py type stuff (adding a lot more validation). So best to hold off from fiddling too much there

PR #309, opened 2018-12-11
specn()

mcmasterg:
This is useful for debugging to fix normally random elements per run number


prjxray/util.py
```diff
@@ -163,3 +163,9 @@ def gen_tile_bits(db_root, tilej, strict=False, verbose=False):
                     # 31_06
                     _bit_inv, (bit_addroff, bit_bitoff) = parse_tagbit(bitstr)
                     yield (baseaddr + bit_addroff, bitbase + bit_bitoff, tag)
+
+
+def specn():
+    # ex: build/specimen_001
+    specdir = os.getenv("SPECDIR")
+    return int(re.match(".*specimen_([0-9]*)", specdir).group(1), 10)
```
galv: Change [0-9]* to [0-9]+.

Otherwise you'll match `build/specimen_`

Also, you can use a raw string for your regex for good practice: `r".*specimen_([0-9]*)"`

PR #329, opened 2018-12-14
`make format` improvements

mithro:
* Update to `clang-format-5.0`
 * Use `yapf` from the environment if it exists.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,8 +27,9 @@ matrix:
         apt:
           sources:
             - ubuntu-toolchain-r-test
+            - llvm-toolchain-trusty-5.0
           packages:
-            - clang-format-3.9
+            - clang-format-5.0
             - g++-6
       install:
         - export CC=gcc-6
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,10 @@
-CLANG_FORMAT ?= clang-format-3.9
+CLANG_FORMAT ?= clang-format-5.0
+PYTHON_FORMAT ?= yapf
+TCL_FORMAT ?= utils//tcl-reformat.sh
 
 .PHONY: database format clean env
 
+IN_ENV = if [ -e env/bin/activate ]; then . env/bin/activate; fi;
 env:
 	virtualenv --python=python3 env
 	. env/bin/activate; pip install -r requirements.txt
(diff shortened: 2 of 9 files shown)


Makefile
```diff
@@ -15,11 +18,13 @@ build:
...
+FORMAT_EXCLUDE = third_party git env build
+FIND_EXCLUDE = $(foreach x,$(FORMAT_EXCLUDE),-and -not -path './$(x)/*')
 format:
-	find . -name \*.cc -and -not -path './third_party/*' -and -not -path './.git/*' -print0 | xargs -0 -P $$(nproc) ${CLANG_FORMAT} -style=file -i
-	find . -name \*.h -and -not -path './third_party/*' -and -not -path './.git/*' -print0 | xargs -0 -P $$(nproc) ${CLANG_FORMAT} -style=file -i
-	find . -name \*.py -and -not -path './third_party/*' -and -not -path './.git/*' -print0 | xargs -0 -P $$(nproc) yapf -p -i
-	find . -name \*.tcl -and -not -path './third_party/*' -and -not -path './.git/*' -print0 | xargs -0 -P $$(nproc) -n 1 ${XRAY_TCL_REFORMAT} 2>/dev/null
+	find . -name \*.cc $(FIND_EXCLUDE) -print0 | xargs -0 -P $$(nproc) ${CLANG_FORMAT} -style=file -i
+	find . -name \*.h $(FIND_EXCLUDE) -print0 | xargs -0 -P $$(nproc) ${CLANG_FORMAT} -style=file -i
+	$(IN_ENV) find . -name \*.py $(FIND_EXCLUDE) -print0 | xargs -0 -P $$(nproc) yapf -p -i
```
JohnDMcMaster: Is this correct? At a quick glance this looks to me like find is executed in the environment instead of yapf

.travis.yml
```diff
@@ -27,8 +27,9 @@ matrix:
         apt:
           sources:
             - ubuntu-toolchain-r-test
+            - llvm-toolchain-trusty-5.0
           packages:
-            - clang-format-3.9
+            - clang-format-5.0
```
JohnDMcMaster: Out of curiosity why did you chose version 5.0 specifically? When .travis and CLANG_FORMAT got out of sync before I had issues with the newer version running on travis, which is why I rolled back to 3.9. But it seems like it accepted this, so guess its fine

mithro: clang-format-5.0 seems the latest available on both our local machines and on travis.

Makefile
mithro: Environment is inherit through the find into yapf. I have confirmed this works locally.

PR #353, opened 2018-12-18
xc7z support

kgugala:
This PR adds initial support for xc7z family

It is a WIP. Fuzzer 005 fails on calculating deltas for CLBs.

### DIFF
--- a/database/xc7z/settings.sh
+++ b/database/xc7z/settings.sh
@@ -1,0 +1,29 @@
+# Unsetting all previous XRAY env variables
+unset ${!XRAY_@}
+
+export XRAY_DATABASE="xc7z"
+export XRAY_PART="xc7z010clg400-1"
+export XRAY_ROI_FRAMES="0x00000000:0xffffffff"
+export XRAY_XC7Z="1"
+
+# All CLB's in part, all BRAM's in part, all DSP's in part.
+export XRAY_ROI_TILEGRID="SLICE_X0Y0:SLICE_X43Y99 RAMB18_X0Y0:RAMB18_X2Y39 RAMB36_X0Y0:RAMB36_X2Y19 DSP48_X0Y0:DSP48_X1Y39"
+
+# These settings must remain in sync
+export XRAY_ROI="SLICE_X34Y0:SLICE_X43Y99 RAMB18_X2Y0:RAMB18_X2Y39 RAMB36_X2Y0:RAMB36_X2Y19 IOB_X0Y0:IOB_X0Y99"
+# Most of CMT X0Y2.
+export XRAY_ROI_GRID_X1="-1"
+export XRAY_ROI_GRID_X2="-1"
+# Include VBRK / VTERM
+export XRAY_ROI_GRID_Y1="-1"
+export XRAY_ROI_GRID_Y2="-1"
+
+export XRAY_PIN_00="T11"
+export XRAY_PIN_01="T10"
+export XRAY_PIN_02="T12"
+export XRAY_PIN_03="U12"
+export XRAY_PIN_04="U13"
+export XRAY_PIN_05="V13"
+export XRAY_PIN_06="V12"
+
+source $(dirname ${BASH_SOURCE[0]})/../../utils/environment.sh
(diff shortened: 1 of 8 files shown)


tools/gen_part_base_yaml.cc
```diff
@@ -8,16 +8,25 @@
```
JohnDMcMaster: This looks like a standalone patch. Can you break this out into its own PR?

mithro:
Can you please add `make format`?

fuzzers/001-part-yaml/Makefile.specimen
```diff
@@ -1,5 +1,11 @@
+ifeq (${XRAY_XC7Z}, 1)
```
JohnDMcMaster: Why doesn't the existing py work for xc7z?

database/xc7z/settings.sh
```diff
@@ -0,0 +1,29 @@
+# Unsetting all previous XRAY env variables
```
JohnDMcMaster: Other files are named such that this should be called database/zynq7/settings.sh

database/xc7z/settings.sh
```diff
@@ -0,0 +1,29 @@
...
+export XRAY_ROI_FRAMES="0x00000000:0xffffffff"
+export XRAY_XC7Z="1"
+
+# All CLB's in part, all BRAM's in part, all DSP's in part.
+export XRAY_ROI_TILEGRID="SLICE_X0Y0:SLICE_X43Y99 RAMB18_X0Y0:RAMB18_X2Y39 RAMB36_X0Y0:RAMB36_X2Y19 DSP48_X0Y0:DSP48_X1Y39"
+
+# These settings must remain in sync
+export XRAY_ROI="SLICE_X34Y0:SLICE_X43Y99 RAMB18_X2Y0:RAMB18_X2Y39 RAMB36_X2Y0:RAMB36_X2Y19 IOB_X0Y0:IOB_X0Y99"
+# Most of CMT X0Y2.
+export XRAY_ROI_GRID_X1="-1"
```
JohnDMcMaster: This is set explicitly in other projects. We should also set this explicitly

database/xc7z/settings.sh
```diff
@@ -0,0 +1,29 @@
+# Unsetting all previous XRAY env variables
+unset ${!XRAY_@}
+
+export XRAY_DATABASE="xc7z"
```
JohnDMcMaster: Similarly this should be zynq7

database/xc7z/settings.sh
```diff
@@ -0,0 +1,29 @@
...
+export XRAY_PART="xc7z010clg400-1"
+export XRAY_ROI_FRAMES="0x00000000:0xffffffff"
+export XRAY_XC7Z="1"
```
JohnDMcMaster: Do we need this? Usually we check value of $XRAY_DATABASE == zynq7

kgugala:
Right now fuzzer 001 uses debug bitstream which is not supported for xc7z family.

BTW it seems we've found a better way to handle the xc7z devices. I'll write more details soon.

tools/gen_part_base_yaml.cc
kc8apf: This new option doesn't seem to have anything to do with generating part YAMLs and duplicates bittool's get_device_id action.

JohnDMcMaster: Oh, I didn't read that carefully enough. Yeah, I'd agree this is probably out of scope

kc8apf:
> 
> 
> Right now fuzzer 001 uses debug bitstream which is not supported for xc7z family.
> 
> BTW it seems we've found a better way to handle the xc7z devices. I'll write more details soon.

FWIW, if BITSTREAM.GENERAL.DEBUGBITSTREAM isn't available, BITSTREAM.GENERAL.PERFRAMECRC works as well.  I preferred DEBUGBITSTREAM because it has more documentation available. Both cause frames to be written to FDRI individually which is all that is necessary to determine the config memory layout of the device.

kgugala:
Yep, we switched to PERFRAMECRC and it works better. I'll clean the code and issue a new PR

Issue #301, opened 2018-12-10
Review mask files


mcmasterg:
These are only processed by 050-intpips and 058-hclkpips? Was int_loop.mk supposed to be doing something with them?

What is the point of this file? Seems like its supposed to be a pre-computed mask the tile/segment occupies, but its not being consistently computed. At a minimum, we only have this for CLB and not for say BRAM

PR #370, opened 2018-12-20
masks: tidy up

mcmasterg:
Fixes https://github.com/SymbiFlow/prjxray/issues/301


mcmasterg:
Local quick build passed which means at least mask generation is working. Quick skips a lot of pushdb, but means that at least decent coverage was reached

PR #373, opened 2018-12-20
Added settings for Zynq FPGA SoC family

kgugala:
This PR adds settings file for xc7z family.

### DIFF
--- a/database/zynq7/settings.sh
+++ b/database/zynq7/settings.sh
@@ -1,0 +1,26 @@
+export XRAY_DATABASE="zynq7"
+export XRAY_PART="xc7z010clg400-1"
+export XRAY_ROI_FRAMES="0x00000000:0xffffffff"
+
+# All CLB's in part, all BRAM's in part, all DSP's in part.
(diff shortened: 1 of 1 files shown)


database/xc7z/settings.sh
```diff
@@ -0,0 +1,29 @@
+# Unsetting all previous XRAY env variables
+unset ${!XRAY_@}
+
+export XRAY_DATABASE="xc7z"
+export XRAY_PART="xc7z010clg400-1"
+export XRAY_ROI_FRAMES="0x00000000:0xffffffff"
+export XRAY_XC7Z="1"
```
JohnDMcMaster: Earlier comment on this still stands. This is redundant with XRAY_DATABASE

kgugala: Actually, if we're about to drop DEBUG bitstream and use PERFRAMECRC for all the families this is not necessary.

kgugala: The name has been changed to zynq. XRAY_XC7Z is removed.

PR #378, opened 2018-12-20
xyaml: fix import error

mcmasterg:


mcmasterg:
this is also breaking tilegrid

mithro:
Your running this without `env`?

mithro:
The line https://github.com/SymbiFlow/prjxray/blob/master/utils/environment.sh#L19 should mean you can import `xjson` / `xyaml`. Can you check what your `$PYTHONPATH` is?

mcmasterg:
Not a bug. Caused by my local screen command dropping PYTHONPATH

PR #386, opened 2018-12-21
reseg: segmatch only build dir

mcmasterg:
generate.py is build dir only, but segmatch was greedy match on everything


mcmasterg:
Context: this a dev only script. The current implementation is problematic because sometimes I'll have multiple build dirs for different runs

Issue #415, opened 2018-12-28
Wrong CMake version is used


GuzTech:
I'm running Arch Linux and I have CMake 3.13 installed. In the build step where we call `cmake ..` I get an error that you need at least version 3.5. The problem is that in the quick start guide, it is mentioned that you need to source the Vivado settings64.sh script first.

Apparently, Vivado 2017.2 has its own CMake version (3.3.2) which takes priority over the system-wide installed one. This is because sourcing the settings64.sh scripts **prepends** the Vivado paths to `$PATH`. I tried changing the order, but the script still prepends its own paths first.

The solution for me was to first run `cmake ..` and then source the settings64.sh script.

JohnDMcMaster:
We actually have someone working on this right now. I believe they are about to make a PR to make settings.sh sourced automatically in the scripts

JohnDMcMaster:
PR was merged. @GuzTech can you see if this is still an issue? Please see updated README

PR #416, opened 2018-12-28
introduce vivado wrapper

felixheld:
This fixes the problem that when sourcing the vivado settings file the
library search path is modified resulting in non-vivado binaries not working
due to being dynamically linked against the vivado libraries instead of the
system ones.

Signed-off-by: Felix Held <felix-github@felixheld.de>


felixheld:
this should fix https://github.com/SymbiFlow/prjxray/issues/415

PR #420, opened 2018-12-29
download-latest: clean files

mcmasterg:
This streamlines getting a cached DB while still doing fuzzer development

### DIFF
--- a/download-latest-db.sh
+++ b/download-latest-db.sh
@@ -1,6 +1,14 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
 GITHUB_PROTO=${1:-https}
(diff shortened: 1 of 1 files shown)


mcmasterg:
We are deleting the misc files, not the .db files

mcmasterg:
I should also add this has bothered me for a while, but is causing more serious problems with people trying to get started

mithro:
I think the correct solution is to move `settings.sh` files out of the database directory and then `.gitignore` the whole `database` directory?

mithro:
It does seem a bit silly to have the `settings.sh` inside the database?

mcmasterg:
you are probably correct in that is the proper fix. But I'd rather not mess with that right now while we have people to work with

mcmasterg:
Filed https://github.com/SymbiFlow/prjxray/issues/421

PR #424, opened 2018-12-29
Update README.md

GuzTech:
Add individual steps and explain how to fix pyjson5 build errors on Arch and Fedora

Signed-off-by: Oguz Meteer <info@guztech.nl>

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,25 +12,29 @@ More documentation can be found published on [prjxray ReadTheDocs site](http://p
 # Quickstart Guide
 Instructions were originally written for Ubuntu 16.04. Please let us know if you have information on other distributions.
 
+### Step 1: ###
 Install Vivado 2017.2. If you did not install to /opt/Xilinx default, then set the environment variable
 XRAY_VIVADO_SETTINGS to point to the settings64.sh file of the installed vivado version, ie
 
     export XRAY_VIVADO_SETTINGS=/opt/Xilinx/Vivado/2017.2/settings64.sh
 
+### Step 2: ###
 Pull submodules:
 
     git submodule update --init --recursive
 
-Install CMake and build the C++ tools:
+### Step 3: ###
+Install CMake:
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -41,33 +45,52 @@ Install CMake and build the C++ tools:
...
     ./download-latest-db.sh
 
+### Step 8: ###
 (Re-)creating the entire database (this will take a very long time!):
 
     cd fuzzers
     make -j$(nproc)
 
-(Re-)creating parts of the database, for example LUT init bits:
+If you want to (re-)create parts of the database, you first need to run:
```
mcmasterg: This is incomplete. Tilegrid is definitely an important dependency. FWIW we just fixed some of the issues that made download-latest-db not good for general use, so maybe we can simplify the instructions to recommend everyone to do that. Maybe remove this small bit right now and address that in a separate PR?

GuzTech: Ok, so then the steps would be to always download the latest database (or build everything yourself) and then you can choose to build parts yourself?

README.md
```diff
@@ -64,23 +64,19 @@ running any other commands:
     source database/artix7/settings.sh
 
 ### Step 7: ###
-**If you are a prjxray developer, skip this step.** But, if you just want to use the Python API with a pre-generated database, just download a current stable version and skip the rest of the instructions: 
+(Option 1) - Download a current stable version (you can use the Python API with
+a pre-generated database)
 
     # WARNING: causes git issues, overrides environment
```
JohnDMcMaster: This has been fixed.v can you delete this line

GuzTech: Which line do you mean? The line with the "#WARNING:"?

JohnDMcMaster: Yeah. If we are making this part of the main instructions now especially we should remove that

README.md
JohnDMcMaster: Yeah. I think given recent fixes, general quick start should now always download latest db

PR #425, opened 2018-12-29
add a basic test of prjxray.util.get_row().gen_sites()

jeremy-rutman:
Hi John (or anyone else),

 This is a basic test of what we discussed at 35C3, testing from prjxray.util import get_roi, get_roi().gen_sites() for two trivial cases: a database with a single tile and either no sites (empty sites dictionary) or a single site.

 I made it so it's easy to add more tests.

HTH
Alon

### DIFF
--- a/tests/test_util.py
+++ b/tests/test_util.py
@@ -1,0 +1,63 @@
+#!/usr/bin/env python3
+
+from os import environ, getcwd, chdir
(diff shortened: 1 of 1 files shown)


JohnDMcMaster:
Thanks! Really appreciate the contribution!

PR #427, opened 2018-12-31
Move settings out of database directory into new settings directory.

mithro:
The fuzzers are dependent on the contents of `settings.sh` and hence the file needs to be kept in sync with the fuzzers. The database is not necessarily kept in sync with the fuzzers, hence the settings.sh files are in the wrong place.

### DIFF
--- a/database/artix7/settings.sh
+++ b/database/artix7/settings.sh
@@ -1,0 +1,1 @@
+source $(dirname ${BASH_SOURCE[0]})/../../settings/artix7.sh
--- a/.gitignore
+++ b/.gitignore
@@ -1,9 +1,7 @@
 /env/
 /build/
+/database/
 .Xil
-# Ignore database directories _except_ for their settings
-database/*/*
-!database/*/settings.sh
 **/specimen_*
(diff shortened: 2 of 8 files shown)


database/artix7/settings.sh
```diff
@@ -1,26 +1 @@
-export XRAY_DATABASE="artix7"
```
JohnDMcMaster: Why are we keeping this file? Are you afraid there will be a compatibility issue?

mithro: Until I push a new rebuild of the database without `settings.sh` there will confusingly be a `database/xxx/settings.sh` and `settings/xxx.sh`. Having these files still exist means that your addition to the `download-database.sh` script will clobber them with the redirect.

Plus I thought it might confuse some people....

mithro:
Rebased.

database/artix7/settings.sh
JohnDMcMaster: Okay, we can leave for now, but lets trash this once we start pushing new DBs

PR #441, opened 2019-01-05
Add all CLKOUT1_DIVIDE bits

emka:
Some values must be present in order to successfully correlate. I increased their probability to appear in `top.py` and hopefully set the number of specimens large enough. Is there a better way to make sure specific configurations appear at least once?


fuzzers/031-mmcm/Makefile
```diff
@@ -1,13 +1,13 @@
...
 # Even slower with multi bit masks...
-N := 8
+N := 14
 include ../fuzzer.mk
 
 SEGDATAS=$(addsuffix /segdata_cmt_top_r_lower_b.txt,$(SPECIMENS))
 
 database: build/segbits_cmt_top_r_lower_b.db
 
-build/segbits_cmt_top_r_lower_b.rdb: $(SPECIMENS_OK)
```
JohnDMcMaster: Generate creates the segbit files, not the rdb files. Unfortunately this is buried in generate.sh, so you'd need to put this on the specimen ok file

fuzzers/031-mmcm/top.py
```diff
@@ -36,18 +37,34 @@ def gen_sites():
...
-        "CLKOUT0_USE_FINE_PS": random.choice(["\"TRUE\"", "\"FALSE\""]),
-        "CLKOUT1_USE_FINE_PS": random.choice(["\"TRUE\"", "\"FALSE\""]),
-        "CLKOUT2_USE_FINE_PS": random.choice(["\"TRUE\"", "\"FALSE\""]),
-        "CLKOUT3_USE_FINE_PS": random.choice(["\"TRUE\"", "\"FALSE\""]),
-        "CLKOUT4_USE_FINE_PS": random.choice(["\"TRUE\"", "\"FALSE\""]),
-        "CLKOUT5_USE_FINE_PS": random.choice(["\"TRUE\"", "\"FALSE\""]),
-        "CLKOUT6_USE_FINE_PS": random.choice(["\"TRUE\"", "\"FALSE\""]),
+        "CLKOUT1_DIVIDE":
+        int(
+            np.random.choice(
```
JohnDMcMaster: Random. Choice would be more consistent with the rest of the code

JohnDMcMaster:
Wow, great work on this! Way harder than I thought it was going to be. Requested some minor changes and then we can merge in. Thanks!

fuzzers/031-mmcm/Makefile
emka: Removed. I am not sure how to add this in `fuzzer.mk`, but it would be a separate commit anyway.

fuzzers/031-mmcm/generate.py
```diff
@@ -7,22 +7,40 @@
...
+                if i == 0:
+                    mybit = 0
+                    mybit2 = 0
+                elif i == 1:
+                    mybit = 1
+            elif value == 128:
+                mybit = 0
+                mybit2 = 0
+
+            segmk.add_site_tag(site, '%s_CNT0_[%u]' % (param, i), mybit)
```
JohnDMcMaster: Just noticed this should maybe be CNT0[n] instead of CNT0_[n]. Regardless, we will probably have to do a whole project style clean up at some point

Issue #438, opened 2019-01-03
Clean up fuzzer names


JohnDMcMaster:
Make them all contain the tile name. Standardize - vs _

mithro:
Thing it would be good to break them into "pip" (IE muxes and such) verses "configuration" (IE lut bits, ff async verse sync configuration, etc).

PR #448, opened 2019-01-07
fuzzers: name with tile type

mcmasterg:
Fixes https://github.com/SymbiFlow/prjxray/issues/438


mithro:
Should the tile type be `int` rather than `pip`?

Issue #171, opened 2018-10-22
All output products should go into "build" dir


mcmasterg:
Right now we dump misc output products into source directories. It would be nice to instead consistently dump all products into a unified output directory

mcmasterg:
Following are left:
  * 051-055
  * 057, 058
  * 071-074
  * 100

mcmasterg:
Following are left:
  * 055
  * 057, 058
  * 071-074
  * 100

mcmasterg:
Following are left:
* 058
* 071-074

JohnDMcMaster:
Following are left: 071-074

mcmasterg:
nats looking at this, adding some details discussed over IRC:

```
I'd like there to only be two generated files in fuzzer directories: build, run.ok
(with the eventual goal to eliminate run.ok, but we can do that later)
This is the case on most fuzzers today, except for the three mentioned in the ticket
You will need to primarily edit the makefile, but might also need to adjust some paths in say tcl or python scripts in that directory
the idea is that if everything is moved to build, simply deleting the build dir will clean it
```

natsfr:
Following left 072-074, I have patches but because of RAM use problem in 072 I can't test them for 072 **AND** 074 (dependent on 072).
Discussing that at the moment to figure out what to do next.

natsfr:
Should be ok to close now

JohnDMcMaster:
Looks like part-yaml still needs this

JohnDMcMaster:
fixed

PR #464, opened 2019-01-08
Bug #171: Fixing build directory for Fuzzers 071

natsfr:
Correcting Bug #171 for Fuzzer 071.
Someone can verify it before merging ?

Signed-off-by: Mehdi Khairy <mehdi.khairy@more-magic.org>


JohnDMcMaster:
Looks right to me and if you tested it I'm going to call it good. I'm going to merge and if build servers complain we can fix it

PR #474, opened 2019-01-09
05x fuzzers rework

kgugala:
This PR refactors the 05x series fuzzers:

* generic generate.py script is introduced. For now only 051 and 056 use it. The rest has to be tested, but it looks like all the 05x fuzzers can use this script
* fuzzer 051 is now todo.txt driven. Beside the test logic it also instantiates picorv and randluts to deliver negative examples for segmaker. With this approach every fuzzer step can solve 10 pips from todo file:

```
wc -l build/todo/*
288 build/todo/1_all.txt
278 build/todo/2_all.txt
268 build/todo/3_all.txt
258 build/todo/4_all.txt
248 build/todo/5_all.txt
```

However, because of the design size each step takes really long time. I assume we can use smaller design with less negative examples to speed up the process and still be able to generate the DB. This has to be investigated.


mcmasterg:
When this is ready for final review, please review WIP tag

kgugala:
With 50 DB entries calculation per iteration the 051 fuzzer finishes in ~30 minutes on 10 core (20 thread) machine.

@JohnDMcMaster please review this PR

mcmasterg:
We should do a follow up PR to exclude imux from 056. I'll file a ticket

mcmasterg:
Merging as this is a good improvement on the current state. Please follow up to exclude imux from rempips though: https://github.com/SymbiFlow/prjxray/issues/485

PR #475, opened 2019-01-09
tilegrid: fix "multi" regression

mcmasterg:


mcmasterg:
looks like travis died. Forcing merge

PR #478, opened 2019-01-10
Add ppips for bram

litghost:


mcmasterg:
Not familiar with ppips, but I have no objections to merging this

Issue #481, opened 2019-01-10
solve FIXME in add_tdb.py of 005-tilegrid fuzzer


acomodi:
There is a FIXME to be solved in the https://github.com/SymbiFlow/prjxray/pull/462 PR (https://github.com/SymbiFlow/prjxray/pull/462/files#diff-558ab8b139f412dbaddc7136214af59d).
In the add_tdb.py file, the script takes only the first bit of a tag, even though there could be more bits. There must be a check that verifies that all the bits relative to a tag belong to the same base address

PR #482, opened 2019-01-10
005-tilegrid/add_tdb.py: added frame check

acomodi:
This PR solves issue #481. add_tdb.py now checks if all the bits in a same tag have the same base address and, if that is the case, the first of them is selected.

### DIFF
--- a/fuzzers/005-tilegrid/add_tdb.py
+++ b/fuzzers/005-tilegrid/add_tdb.py
@@ -55,7 +55,7 @@ def parse_addr(line):
 def check_frames(frames):
     baseaddr = set()
     for frame in frames:
-        baseaddr.add(round(frame / 128))
+        baseaddr.add(frame // 128)
     assert len(baseaddr) == 1, "Multiple base addresses for the same tag"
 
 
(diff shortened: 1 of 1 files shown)


fuzzers/005-tilegrid/add_tdb.py
```diff
@@ -52,18 +52,33 @@ def parse_addr(line):
     return frame, wordidx, bitidx
 
 
+def check_frames(frames):
+    baseaddr = set()
+    for frame in frames:
+        baseaddr.add(round(frame / 128))
```
mcmasterg: use //

fuzzers/005-tilegrid/add_tdb.py
```diff
@@ -52,18 +52,33 @@ def parse_addr(line):
...
-        addrstr = parts[1]
+        addrlist = parts[1:]
+
+        frames = list()
+
+        for addrstr in addrlist:
+            frame, wordidx, bitidx = parse_addr(addrstr)
+            frames.append(frame)
+
+        # Check the frames base addresses
```
mcmasterg: ...is the same

fuzzers/005-tilegrid/add_tdb.py
```diff
@@ -52,18 +52,33 @@ def parse_addr(line):
...
         parts = l.split(' ')
-        # FIXME: need to check that all bits in part have same baseaddr, for now only the first bit is taken
-        #assert len(parts) == 2, "Unresolved bit: %s" % l
         tagstr = parts[0]
-        addrstr = parts[1]
+        addrlist = parts[1:]
+
+        frames = list()
+
+        for addrstr in addrlist:
```
mcmasterg: we are implicitly taking the last one here? should probably add a comment about that

fuzzers/005-tilegrid/add_tdb.py
acomodi: done

fuzzers/005-tilegrid/add_tdb.py
acomodi: done

Issue #341, opened 2018-12-18
iob tilegrid: support LIOB33_SING_X0Y100/LIOB33_SING_X0Y149


mcmasterg:
These tiles are a little special and so were omitted from the base implementation (singled ended half sized LIOB33_SING tiles instead of full sized differential LIOB33)

PR #488, opened 2019-01-10
Support IOB tiles with only one site (IOB_SING)

mcmasterg:
Fixes https://github.com/SymbiFlow/prjxray/issues/341

These were held off because they generate some weird special cases I didn't want to think about at the time. So here's a proposed solution.

First, the problem: in a7, IOBs are typically in LIOB33 tiles. These have two IOB sites, just like a SLICE has several CLBs. However, unlike CLBs, these IOBs are sometimes split into LIOB33_SING tiles, which only have a single IOB33. The problem is that in some cases the single instance is the Y=1 case, which basically starts at word 2 in our LIOB33_SING tile (if you follow the pattern in LIOB33 or want a Y=0 and a Y=1 LIOB33_SING to have a consistent address space). The issue is that this actually occurs at FDRI word 0, meaning we essentially have a negative tile address (-2) if we want the reference bits, starting at address 2, to be in frame 0.

Sample tilegrid output from running this fuzzer:

```
//Special bottom entry
//its missing the Y=0 site
"LIOB33_SING_X0Y100": {
    "bits": {
        "CLB_IO_CLK": {
            "baseaddr": "0x00020000",
            "frames": 42,
            "height": 4,
            "offset": -2,
            "words": 4
        }
    },
//A typical entry
"LIOB33_X0Y101": {
    "bits": {
        "CLB_IO_CLK": {
            "baseaddr": "0x00020000",
            "frames": 42,
            "height": 4,
            "offset": 2,
            "words": 4
        }
    },
...
//Special bottom entry
//its missing the Y=1 site
"LIOB33_SING_X0Y149": {
    "bits": {
        "CLB_IO_CLK": {
            "baseaddr": "0x00020000",
            "frames": 42,
            "height": 2,
            "offset": 99,
            "words": 2
        }
    },
```

I ran segprint and the 030-iob fuzzer. Both seemed happy with this, so I'm inclined to commit this as is and move forward.

### DIFF
--- a/fuzzers/005-tilegrid/add_tdb.py
+++ b/fuzzers/005-tilegrid/add_tdb.py
@@ -25,10 +25,12 @@ def add_tile_bits(
     bits = tile_db['bits']
     block_type = util.addr2btype(baseaddr)
 
-    assert 0 <= offset <= 100, offset
+    assert offset <= 100, (tile_name, offset)
+    # Few rare cases at X=0 for double width tiles split in half
+    assert offset >= 0 or "IOB" in tile_name, (tile_name, offset)
     assert 1 <= words <= 101
     assert offset + words <= 101, (
-        tile_db, offset + words, offset, words, block_type)
+        tile_name, offset + words, offset, words, block_type)
 
     baseaddr_str = '0x%08X' % baseaddr
 
@@ -148,8 +150,19 @@ def run(fn_in, fn_out, verbose=False):
             bitsj = tilej['bits']
             bt = util.addr2btype(frame)
             verbose and print("Add %s %08X_%03u" % (tile, frame, wordidx))
+            # Special case for half height IOB
+            if tile == "LIOB33_SING_X0Y149":
+                tile_words = 2
+            else:
+                tile_words = words
             add_tile_bits(
-                tile, tilej, frame, wordidx, frames, words, verbose=verbose)
+                tile,
+                tilej,
+                frame,
+                wordidx,
+                frames,
+                tile_words,
+                verbose=verbose)
 
     # Save
     json.dump(
--- a/fuzzers/005-tilegrid/add_tdb.py
+++ b/fuzzers/005-tilegrid/add_tdb.py
@@ -6,7 +6,15 @@ import os
 
 
 # Copied from generate_full.py
-def add_tile_bits(tile_db, baseaddr, offset, frames, words, height=None):
+def add_tile_bits(
+        tile_name,
+        tile_db,
+        baseaddr,
+        offset,
+        frames,
+        words,
+        height=None,
+        verbose=False):
     '''
     Record data structure geometry for the given tile baseaddr
     For most tiles there is only one baseaddr, but some like BRAM have multiple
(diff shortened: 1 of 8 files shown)


fuzzers/005-tilegrid/add_tdb.py
```diff
@@ -17,16 +25,29 @@ def add_tile_bits(tile_db, baseaddr, offset, frames, words, height=None):
     bits = tile_db['bits']
     block_type = util.addr2btype(baseaddr)
 
-    assert 0 <= offset <= 100, offset
+    assert offset <= 100, (tile_name, offset)
+    # Few rare cases at X=0 for double width tiles split in half
+    assert offset >= 0 or "IOB" in tile_name, (tile_name, offset)
```
litghost: Make explicit that this allows for negative word addresses

mcmasterg:
No, there is not a negative segbits file. Only the upper half of the file is used

mcmasterg:
the latter

fuzzers/005-tilegrid/iob/top.py
```diff
@@ -11,24 +11,19 @@
 
 
 def gen_iobs():
-    '''
-    IOB33S: main IOB of a diff pair
-    IOB33M: secondary IOB of a diff pair
```
mithro: `IOB33S` was actually the main? I would have guess the `S` and `M` stood for `Master`, `Slave` or `Main`, `Secondary`.

mithro:
If I understand correctly, what you are saying is that;

 * Like CLBs, an IOBs has two sites, `Y0` and `Y1`.
 * However, `LIOB33_SING` is a special case of an `LIOB33` were either the `Y0` or the `Y1` site is missing.

How do we deal with this in `CLBL` verse `CLBM`. In one case we have two `SLICEL` and the other case we have `SLICEL` and a `SLICEM`? IE The table below;

| Tile | Site Y0 | Site Y1 |
|------|-------|------|
| `CLBL` | `SLICEL` | `SLICEL` |
| `CLBM` | `SLICEL` | `SLICEM` |

Isn't the IOB just three cases like below?

| Tile | Site Y0 | Site Y1 |
|------|-------|------|
| `IOB33` | `IOB` | `IOB` |
| `IOB33S_TOP` | `<empty>` | `IOB` |
| `IOB33S_BOT` | `IOB` | `<empty>` |

Is it really hard to make the tools able to handle the "`<empty>`" in the first site of a tile?

JohnDMcMaster:
Tim: is that question for me or @litghost ?

litghost:
> 
> 
> Tim: is that question for me or @litghost ?
> 
> Maybe to answer: the problem is that the tile_type for "IOB33S_TOP" and "IOB33S_BOT" is the same

Exactly, IOB33S_TOP and IOB33S_BOT are both just IOB33S_SING.  The ideal case would've been for these tiles to shares the same segbits.  I assume we've confirmed this is not the case?

mithro:
Can we just special case `IOBS` and generate the `IOBS_TOP` / `IOBS_BOT` ourselves?

litghost:
> 
> 
> Can we just special case `IOBS` and generate the `IOBS_TOP` / `IOBS_BOT` ourselves?

Sure, we just need to agree on an approach.  If we go with iob_top and iob_bot, it changes how the fuzzers will output database data.  The original approach was to have both _top and _bot in the same segbit.  It's not obvious which requires less rework overall, but I'd say that iob_top and iob_bot is probably most consistent with how the other tiles work.

Issue #465, opened 2019-01-08
High RAM usage in fuzzer 072-ordered-wires


natsfr:
When running the 072 Fuzzer the process stop outputting message quickly on that last one for long time:

**INFO: [Common 17-14] Message 'Vivado 12-2683' appears 100 times and further instances of the messages will be disabled. Use the Tcl command set_msg_config to change the current settings.**

After a long time the RAM use is at about 16GB on my machine and the process get killed:
**/home/nats/Xilinx/Vivado/2017.2/bin/loader: line 179: 14890 Killed                  "$RDI_PROG" "$@"
Makefile:13: recipe for target 'specimen_001/OK' failed
make[1]: *** [specimen_001/OK] Error 137
make[1]: Leaving directory '/home/nats/project/symbiflow/prjxray/fuzzers/072-ordered_wires'
Makefile:17: recipe for target 'run' failed
make: *** [run] Error 2**

Maybe my computer is too RAM limited but I would expect 16GB to be enough.

mithro:
I believe 128gb RAM is needed for some at the moment..

natsfr:
Working on a solution and may have found a way to mitigate the issue.

natsfr:
I found a way to mitigate the problem. Let's try to explain here what is going on from my point of view.
1) in some way the vivado **get_nodes** tcl wrapper is leaky. It could be a bad implementation of their tcl interpreter or like litghost pointed out it could be something in the binary behind. To prove that I tried a smaller code:
```
set pips [get_pips]
 
for { set i 0 } { $i < [llength $pips] } { incr i } {
    set dwn_nodes [get_nodes -downhill -of_object [lindex $pips $i]]
    for { set j 0 } { $j < [llength $dwn_nodes] } { incr j } {
        #puts [lindex $dwn_nodes $j]
    }
    unset dwn_nodes
}
```
Even with unsetting the returned list and making sure it doesn't stay linked to global namespace vars the leak is here (about 9MB per seconds on my computer).

2) I tried something more elaborate using slave interpreter because the spec of TCL says that a slave interpreter must be flushed away with all his own variable after delete (not necessarily right after, the GC is free to choose when) but it doesn't happens in the following code, it still eats RAM until being killed by my Linux.
```
proc process_block { start stop } {
    puts "inside interpreter $start $stop"
    set downhill_fp [open downhill_wires.txt a]
    set uphill_fp [open uphill_wires.txt a]
   
    set pips [get_pips]
   
[...]

PR #489, opened 2019-01-10
Fixing Bug #171 and #465

natsfr:
This pull request aim two linked problems.
In the first place it closes the Bug #171 by correcting last build path in Fuzzers 072-074.
At the same time the commit 40738b7 is changing the way Fuzzer 072 works, I basically split jobs in different small blocks, avoiding the huge RAM use of that Fuzzer and it allows to parallelize them.
For now it'll be necessary to change the number of threads in "run_fuzzer.py". See Bug #465 for more explanation.
This PR needs a testing by someone who can generate or already has the uphill and downhill wires files ! I can't do it myself because of the RAM use of original code, but size match. For information on my computer High-End quad core i7 and 16GB ram it takes 30 minutes to generate the files.


fuzzers/072-ordered_wires/run_fuzzer.py
```diff
@@ -0,0 +1,49 @@
+import threading
+import os
+import shutil
+
+def start_vivado(blockID, start, stop):
+    print ("Running instance :" + str(blockID))
+    os.system("${XRAY_VIVADO} -mode batch -source $FUZDIR/job.tcl -tclargs " + str(blockID) + " " + str(start) + " " + str(stop))
+
```
litghost: Use subprocess.check_call

litghost:
This same technique could be applied to fuzzer 074.  Can to take a crack at that in the mean time?  I'll compare your 072 and comment on deltas.

fuzzers/072-ordered_wires/run_fuzzer.py
natsfr: I'll take a look in about 20 hours. Sorry I'm a bit rusty in python.

natsfr:
> This same technique could be applied to fuzzer 074. Care to take a crack at that in the mean time? I'll compare your 072 and comment on deltas.

Sure I'll take a look just after correcting python warning you and "travis" noticed.

litghost:
I got the same sha1sum with the old code as you did:

# sha1sum downhill_wires.txt 
ce520ef646faf050651189290e1128818c19d426  downhill_wires.txt
# sha1sum uphill_wires.txt 
2ade66e56811d9753e48d51aa0fdcc1692e5d1ef  uphill_wires.txt

Given that the sha1sum is the same, once you fix the system -> subprocess changes, this is good to merge.  In other good news, your new code runs ~3 times faster than the old approach!

fuzzers/072-ordered_wires/run_fuzzer.py
```diff
@@ -0,0 +1,49 @@
...
+def get_nb_pips():
+    print ("Fetching nb pips")
+    os.system("${XRAY_VIVADO} -mode batch -source $FUZDIR/get_pipscount.tcl")
+    countfile = open("nb_pips.txt", "r")
+    return int(countfile.readline())
+
+if __name__ == "__main__":
```
mithro: You would normally do something like the following;
```python
import sys

def main(argv):
   print("My code goes here")
   if len(argv) > 2:
      return 0
   else:
      return 1

if __name__ == "__main__":
  sys.exit(main(sys.argv))

Comment (natsfr) [reply fuzzers/072-ordered_wires/run_fuzzer.py]: Oky will add that to the evening cleanup of that code :) Again I’m little rusty in python so next PR should be easier.

**Comment by natsfr** [comment]
Thanks for the test litghost. We should be able to merge that soon and I’ll do the same for 074 :)

**Comment by natsfr** [comment]
I fixed python, can I do an other review on the code ?
Thanks :)

**Comment by natsfr** [comment]
ohhhh thanks @mithro you just saved me certainly a lot of hours :)

File: fuzzers/072-ordered_wires/run_fuzzer.py [hunk 3]
```diff
@@ -0,0 +1,75 @@
...
+def get_nb_pips():
+    print("Fetching nb pips")
+    try:
+        subprocess.check_call(
+            "${XRAY_VIVADO} -mode batch -source $FUZDIR/get_pipscount.tcl",
+            shell=True)
+    except subprocess.CalledProcessError:
+        print("Couldn't start vivado instance")
+        sys.exit(1)
+
```
litghost: Don't swallow the exception.  Either don't catch the exception, or append to the exception

PR #490, opened 2019-01-11
Add utility to convert from bit to FASM in one program.

litghost:


mcmasterg:
do we want to eliminate bits2fasm? or do you think its helpful to have both for now

Issue #495, opened 2019-01-11
Getting INT_L baseaddrs connected to PSS in Zynq


acomodi:
For the xc7z parts, the Processing System is connected to the fabric through the first column of INT tiles. The processing of the INT_L/R base addresses as it is right now is done by propagating the base address of the corresponding CLB/BRAM which they belong to.
As the INT tiles connected to the PSS do not have a belonging site (it should actually be the PSS hard block), we need to find another way to populate the INT tiles base addresses for what regards the xc7z parts.

kgugala:
@acomodi can we close this one?

acomodi:
Yes

PR #503, opened 2019-01-11
005-tilegrid/generate_full: propagating INT_L base addrs

acomodi:
This PR is to solve #495.
I have added a way to propagate INT_L base addresses which do not have an adjacent block (In this case the INT_L_X0Y* interconnects) of the top fabric row. 
There is also the need to do the same for the bottom CMT column, given that the PS7 is attached to all the left interconnects of the FPGA fabric.

### DIFF
--- a/fuzzers/005-tilegrid/generate_full.py
+++ b/fuzzers/005-tilegrid/generate_full.py
@@ -551,7 +551,7 @@ def db_int_fixup(database, tiles, tiles_by_grid):
                 tile_name, tile, baseaddr, offset, frames, words, height)
 
 
-def run(json_in_fn, json_out_fn, tiles_fn, deltas_fns, int_tdb, verbose=False):
+def run(json_in_fn, json_out_fn, tiles_fn, deltas_fns, int_tdb=None, verbose=False):
     # Load input files
     tiles = load_tiles(tiles_fn)
     site_baseaddr = load_baseaddrs(deltas_fns)
@@ -570,8 +570,9 @@ def run(json_in_fn, json_out_fn, tiles_fn, deltas_fns, int_tdb, verbose=False):
     db_add_bits(database, segments)
     db_add_segments(database, segments)
 
-    tile_baseaddrs_fixup = load_tdb_baseaddr(database, int_tdb)
-    db_int_fixup(database, tile_baseaddrs_fixup, tiles_by_grid)
+    if int_tdb is not None:
+        tile_baseaddrs_fixup = load_tdb_baseaddr(database, int_tdb)
+        db_int_fixup(database, tile_baseaddrs_fixup, tiles_by_grid)
 
     # Save
     json.dump(
@@ -602,6 +603,7 @@ def main():
         "deltas", nargs="*", help=".bit diffs to create base addresses from")
     parser.add_argument(
         "--int-tdb",
+        default=None,
         help=".tdb diffs to fill the interconnects without any adjacent CLB")
     args = parser.parse_args()
 
--- a/fuzzers/005-tilegrid/generate_full.py
+++ b/fuzzers/005-tilegrid/generate_full.py
@@ -14,6 +14,7 @@ from generate import load_tiles
 from prjxray import util
 import util as localutil
 
+
 def nolr(tile_type):
     '''
     Remove _L or _R suffix tile_type suffix, if present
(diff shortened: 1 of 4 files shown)


fuzzers/005-tilegrid/generate_full.py
```diff
@@ -224,6 +224,13 @@ def process_bram_dsp():
                     segtype=tile_type.lower().replace("_", "%d_" % k, 1),
                     baseaddr=baseaddr)
 
+        def process_int():
+            if "INT" not in tile_type:
```
JohnDMcMaster: simplify this if else assert to a single assert line

fuzzers/005-tilegrid/generate_full.py
```diff
@@ -499,13 +504,12 @@ def add_tile_bits(tile_db, baseaddr, offset, frames, words, height=None):
...
     bits = tile_db['bits']
     block_type = util.addr2btype(baseaddr)
-
     assert 0 <= offset <= 100, offset
     assert 1 <= words <= 101
     assert offset + words <= 101, (
         tile_db, offset + words, offset, words, block_type)
 
-    assert block_type not in bits
+    #assert block_type not in bits:
```
JohnDMcMaster: We should handle more explicitly

JohnDMcMaster: Take a look at my IOB PR. I had some logic there for doing a merge check on these. We should maybe also move this utility function to somewhere shared since its in generate_full.py and add_tdb.py. Or maybe just import from one of the existing .py files instead of duplicating

acomodi: Done

fuzzers/005-tilegrid/generate_full.py
acomodi: Done

fuzzers/005-tilegrid/util.py
```diff
@@ -0,0 +1,63 @@
...
+    baseaddr_str = '0x%08X' % baseaddr
+    block = bits.get(block_type, None)
+    if block is not None:
+        verbose and print(
+            "%s: existing defintion for %s" % (tile_name, block_type))
+        assert block["baseaddr"] == baseaddr_str
+        assert block["frames"] == frames
+        assert block["offset"] == offset, "%s; orig offset %s, new %s" % (
+            tile_name, block["offset"], offset)
+        assert block["words"] == words
```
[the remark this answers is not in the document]
acomodi: corrected

fuzzers/005-tilegrid/generate_full.py
```diff
@@ -562,18 +529,27 @@ def db_add_bits(database, segments):
...
-                    ("HCLK", "CLB_IO_CLK"): (26, 1, 1),
-                    ("INT", "CLB_IO_CLK"): (28, 2, 2),
-                    ("BRAM", "CLB_IO_CLK"): (28, 10, None),
-                    ("BRAM", "BLOCK_RAM"): (128, 10, None),
-                    ("DSP", "CLB_IO_CLK"): (28, 2, 10),
-                    ("INT_INTERFACE", "CLB_IO_CLK"): (28, 2, None),
-                    ("BRAM_INT_INTERFACE", "CLB_IO_CLK"): (28, 2, None),
-                }.get((nolr(tile_type), block_type), None)
+
+                is_not_segint = "SEG_INT" not in segment_name
```
mcmasterg: Why is this special case needed? Why not just use the tile type like the rest of the code?

fuzzers/005-tilegrid/generate_full.py
```diff
@@ -300,8 +306,6 @@ def create_segment_for_int_lr(
             segtype=database[tile]["type"],
             verbose=verbose,
         )
-    else:
```
mcmasterg: Why was this assertion removed?

fuzzers/005-tilegrid/generate_full.py
acomodi: Mainly because there already is an "INT" in the previous entry with a 28 as frames value. I wanted to separate the two different cases

mcmasterg: Isn't 28 the correct value? INT tiles don't use more than that. Its possible the PS7 configuration is larger, but all of it may not apply to INT

fuzzers/005-tilegrid/generate_full.py
acomodi: Mainly because the else branch did not affect the behaviour given that no segment is added if the if condition is not met.

fuzzers/005-tilegrid/generate_full.py
acomodi: Ok, I'll keep 28 then and delete the special case

fuzzers/005-tilegrid/generate_full.py
mcmasterg: We discussed this and I think more fundamental rework is needed on this. I'm changing title to WIP, please change back once ready for review again

acomodi:
I have checked and confirmed that the changes do not break the fuzzer when using artix7/kintex7

PR #504, opened 2019-01-11
Improve clb READMEs

mcmasterg:
import from wiki and clean up


mcmasterg:
FYI I went significantly beyond this, will update PR with more READMEs in the near future

PR #510, opened 2019-01-14
utils: add cleandb tool

kgugala:
This PR adds a tool for removing partial DB from a global one. This is useful when a certain fuzzer is rerun during development

### DIFF
--- a/utils/cleandb.py
+++ b/utils/cleandb.py
@@ -1,0 +1,54 @@
+#!/usr/bin/env python3
+
+import sys, re
+import os
+from prjxray import util
+
+
+def run(globaldb, localdb, verbose=False):
+
+    local_db_files = list()
+    work_db_files = list()
+
+    # get DB files
+
+    global_entries = {}
+    local_entries = {}
+    final_entries = {}
+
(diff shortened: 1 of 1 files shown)


utils/cleandb.py
```diff
@@ -0,0 +1,54 @@
...
+
+    verbose and print("removing %s from %s" % (localdb, globaldb))
+    # parse global db
+    for line, (tag, bits, mode) in util.parse_db_lines(globaldb):
+        global_entries[tag] = bits
+    # parse local db
+    for line, (tag, bits, mode) in util.parse_db_lines(localdb):
+        local_entries[tag] = bits
+
+    for entry in global_entries:
```
litghost: Why not just do:

```python
final_entries = global_entries
for entry in local_entries:
  if entry in final_entries:
    verbose and print("Removing entry %s" % entry)
    del final_entries[entry]
```

mithro:
Does this tool still make sense? If so, we should merge it.

tmichalak:
This tool comes in handy when experimenting with fuzzers. It can be used to revert the changes in the database introduced by the tested fuzzer in order to run several iterations of it.

PR #512, opened 2019-01-15
Increase N for NFFMUX and NOUTMUX to prevent false matches.

litghost:
Some INT bits were getting matched into the NFFMUX and NOUTMUX fuzzers.  Increasing N to 3 appears to prevent this.


kgugala:
lgtm

PR #514, opened 2019-01-15
fuzzers: 056: do not attempt to solve IMUX*LOGIC_OUTS

### LINKED ISSUE #485: Exclude imux from 056 rempips
Follow up to https://github.com/SymbiFlow/prjxray/pull/474

056 can't solve imux. We should explicitly exclude it and/or document this in the README

kgugala:
IMUX*LOGIC_OUTS should be solved by 051
fixes #485


litghost:
Can you update the documentation or add a comment about this change?

kgugala:
actually this PR may not be needed anymore. Since #520 is merged, the 056 fuzzer should newer need to solve IMUX*LOGIC_OUTS. Before the merge, those PIPs were solved twice - in 051 and 056 and in 056 it significantly increased the build time. #520 fixes this and now at the point when build process reaches fuzzer 056 all those PIPs should be in the database

litghost:
Okay, I'll try a build without this PR once I verify the current run.

litghost:
Agree, this PR doesn't seem required anymore.  Closing.

PR #520, opened 2019-01-15
fuzzers: Copy all the db files to build/database

### LINKED ISSUE #519: 05x fuzzers todo files contain already solved PIPs
e.g 056 todo contain PIPs solved by 050. This should not happen and it slows down the whole build

kgugala:
fixes #519

### DIFF
--- a/fuzzers/int_loop.mk
+++ b/fuzzers/int_loop.mk
@@ -42,7 +42,7 @@ ifneq ($(QUICK),Y)
 	${XRAY_DBFIXUP} --db-root build --clb-int
 	# https://github.com/SymbiFlow/prjxray/issues/399
 	# Clobber existing .db to eliminate potential conflicts
-	if [ -f ${XRAY_DATABASE_DIR}/${XRAY_DATABASE}/segbits_clb*.db ] ; then cp ${XRAY_DATABASE_DIR}/${XRAY_DATABASE}/segbits_clb*.db build/database/${XRAY_DATABASE}; else bash ${XRAY_DIR}/fuzzers/int_create_empty_db.sh; fi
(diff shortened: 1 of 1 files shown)


fuzzers/int_loop.mk
```diff
@@ -42,7 +42,7 @@ ifneq ($(QUICK),Y)
 	${XRAY_DBFIXUP} --db-root build --clb-int
 	# https://github.com/SymbiFlow/prjxray/issues/399
 	# Clobber existing .db to eliminate potential conflicts
-	if [ -f ${XRAY_DATABASE_DIR}/${XRAY_DATABASE}/segbits_clb*.db ] ; then cp ${XRAY_DATABASE_DIR}/${XRAY_DATABASE}/segbits_clb*.db build/database/${XRAY_DATABASE}; else bash ${XRAY_DIR}/fuzzers/int_create_empty_db.sh; fi
+	if [ -f ${XRAY_DATABASE_DIR}/${XRAY_DATABASE}/segbits*.db ] ; then cp ${XRAY_DATABASE_DIR}/${XRAY_DATABASE}/segbits*.db build/database/${XRAY_DATABASE}; else bash ${XRAY_DIR}/fuzzers/int_create_empty_db.sh; fi
```
litghost: This if statement is always false :(

https://stackoverflow.com/questions/2937407/test-whether-a-glob-has-any-matches-in-bash

kgugala: I think this should be even more intelligent. We have quite a few db files and we may hit any combination of them existing or not. IMO we should have a list of db files, iterate over it, copy existing and touch unexisting.

kgugala: or simply create empty db (if none present) in settings script and do not care about checks in fuzzers makefiles

kgugala: I have opened followup PR #531

PR #531, opened 2019-01-16
create empty db files if they don't exist

kgugala:
this PR is a follow up to #520

### DIFF
--- a/utils/environment.sh
+++ b/utils/environment.sh
@@ -44,3 +44,41 @@ if [ $(${XRAY_VIVADO} -h |grep Vivado |cut -d\  -f 2) != "v2017.2" ] ; then
     export XRAY_DIR="/bad/vivado/version"
     return
 fi
+
+function init_db
+{
+  # Check if database files exist. If don't, create empty ones
+  db_files="mask_clbll_l.db \
+            mask_clbll_r.db \
+            mask_clblm_l.db \
+            mask_clblm_r.db \
+            mask_hclk_l.db \
+            mask_hclk_r.db \
+            ppips_clbll_l.db \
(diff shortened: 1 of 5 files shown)


utils/environment.sh
```diff
@@ -44,3 +44,38 @@ if [ $(${XRAY_VIVADO} -h |grep Vivado |cut -d\  -f 2) != "v2017.2" ] ; then
     export XRAY_DIR="/bad/vivado/version"
     return
 fi
+
+# Check if database files exist, if don't create empty ones
```
litghost: This doesn't belong in environment.sh.  Maybe a call in ${XRAY_PART}/settings.sh?

kgugala: sure, we can move it there but we'd have to call it in every settings file (for that I'd pack this into another script and call the script)
ROI harness minitest introduces it's own settings file. The only common part between all of them is the environment.sh, that is why I put it there.
Shall I add the call every setting file in the project?

litghost: enviroment.sh might be common, but it should have side affects.

Have environment.sh define a function is fine, just call it in each settings.sh

kgugala: sure, I'll do it that way then

kgugala: done

experiments/clbpips/generate.sh
```diff
@@ -1,6 +1,7 @@
 #!/bin/bash
 
 source ../../utils/environment.sh
+init_db
```
litghost: Only call init_db for settings.sh

kgugala: fixed

PR #535, opened 2019-01-16
Increase solved PIPs per loop to 50.

litghost:


kgugala:
with this we can remove explicit TODO_N = 50 setting in 051

Issue #534, opened 2019-01-16
fuzzer 018-clbram fails with duplicate bits


kgugala:
```
AssertionError: strict: got duplicate bits frozenset({'31_16', '31_17'}): CLB.SLICE_X0.BLUT.RAM CLB.SLICE_X0.ALUT.RAM
...
make[2]: *** [pushdb] Error 1
make[1]: *** [run] Error 2
make: *** [018-clb-ram/run.ok] Error 2
```
tested with zynq7 settings

acomodi:
tested with artix7 settings and there are no errors

kgugala:
I rerun this fuzzer alone on zynq also with no errors. It may be interference with some other fuzzer, or some kind of a race condition.

litghost:
This could be a side effect of using randomness for CLB RAM.  One way to solve this is to ensure that all 16 combination (or so) 64X1S's are generated.  So the permutation starts with 4 LUT6's in a SLICE, and the permutation ends with 4 64X1S.  There are more than enough CLB's to do this for both Y0 and Y1.

litghost:
Note that the DLUT must be a 64X1S if any 64X1S are present.

litghost:
#543 fixes this.

PR #536, opened 2019-01-16
Increase CLB match requirement.

litghost:
Still seeing false matches, seeing if increase match requirement drives false rate to 0.

### DIFF
--- a/fuzzers/clb.mk
+++ b/fuzzers/clb.mk
@@ -16,7 +16,7 @@ endif
 
 
 build/segbits_clbx.rdb: $(SPECIMENS_OK)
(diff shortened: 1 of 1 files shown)


acomodi:
I have tried to run the 018-clb-bram fuzzer and got the error reported in https://github.com/SymbiFlow/prjxray/issues/534. 

Even by applying the changes in this PR the error is still present because the basically, the bits related to `CLB.SLICE_X0.BLUT.RAM` and `CLB.SLICE_X0.ALUT.RAM` change several times, but they change always together in the `segdata_clbl[lm]_[lr].txt` files of the various specimens.

litghost:
Which targets?  I'm mostly testing on artix7

acomodi:
zynq7 actually, I tried it for the artix7 and it doesn't fail

PR #538, opened 2019-01-16
Change int termination criteria to progress

litghost:


litghost:
Unfortunately 056-pip-rem cannot solve some pips, and as a result 050 needs to run multiple iterations to solve some PIPs, specifically PIPs matching the regexp "(LOGIC_OUTS)|(IMUX)".  It is unclear how many iterations over at reasonable termination criteria are required to solve all INT PIPs matching "(LOGIC_OUTS)|(IMUX)".  A better solution is to change 050 to use a new termination condition, and then make a version of 056-pip-rem (e.g. a new 055) that can solve for "(LOGIC_OUTS)|(IMUX)".

kgugala:
apparently 050 can solve them

litghost:
> IMUX*.LOGIC_OUTS* are solved by 051. I noticed that 056 cannot solve BYP_ALT*.LOGIC_OUTS. I'll check if 51 can solve them

Good catch.  This the is the list "left over" from 056 with the early termination:

INT_L.BYP_ALT6.LOGIC_OUTS_L7
INT_L.BYP_ALT7.LOGIC_OUTS_L3
INT_L.IMUX_L19.GFAN0
INT_R.BYP_ALT6.LOGIC_OUTS7
INT_R.BYP_ALT7.LOGIC_OUTS3

litghost:
This PR combined with #544 should work.  Test is running now.

litghost:
With only forcing 050 to solve IMUX.*GFAN and minimum progress of 50 pips per iteration, 050 terminates in 5 iterations.  Test is running to see if all PIPs are solved.

PR #543, opened 2019-01-17
01x CLB rerun if pushdb fails

acomodi:
This PR solves https://github.com/SymbiFlow/prjxray/issues/534 issue and issues related to the CLB fuzzers for which the segmatch solver is unable to find a solution.

Basically the CLB fuzzers are now re-run whenever the pushdb fails because of a bad db file generation.
The top.py of the CLB fuzzers now adds to the CLBN variable an increment calculated in the clb.mk file.
The increment allows to have an higher probability of getting a suitable design for the fuzzer.

Mainly this was done to ensure the correct clb .db file generation with targets that do require more CLBs to be instantiated.

### DIFF
--- a/fuzzers/fuzzer.mk
+++ b/fuzzers/fuzzer.mk
@@ -2,26 +2,19 @@ N ?= 1
 SPECIMENS := $(addprefix build/specimen_,$(shell seq -f '%03.0f' $(N)))
 SPECIMENS_OK := $(addsuffix /OK,$(SPECIMENS))
 ENV_VAR ?=
-VAR ?=
-ITER ?= 0
-MAX_ITER ?= 10
+FUZDIR ?= ${PWD}
 
 all: database
 
 $(SPECIMENS_OK):
+	echo ${ENV_VAR}
 	mkdir -p build
-	if [ -f `pwd`/generate.sh ]; then export $(ENV_VAR); bash `pwd`/generate.sh $(subst /OK,,$@); else bash ${XRAY_DIR}/utils/top_generate.sh $(subst /OK,,$@); fi
+	if [ -f $(FUZDIR)/generate.sh ]; then export $(ENV_VAR); bash $(FUZDIR)/generate.sh $(subst /OK,,$@); else bash ${XRAY_DIR}/utils/top_generate.sh $(subst /OK,,$@); fi
 
 run:
-	# If the database presents errors or is incomplete, the fuzzer is rerun.
-	# When it reaches the maximum number of iterations it fails.
-	@if [ $(ITER) -gt $(MAX_ITER) ]; then \
-		echo "Max Iterations reached. Fuzzer unsolvable."; \
-		exit 1; \
-	fi
 	$(MAKE) clean
 	$(MAKE) database
-	$(MAKE) pushdb || $(MAKE) $(VAR) ITER=$$(($(ITER) + 1)) run
+	$(MAKE) pushdb
 	touch run.ok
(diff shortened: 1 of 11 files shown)


fuzzers/fuzzer.mk
```diff
@@ -1,17 +1,27 @@
...
+VAR ?=
+ITER ?= 0
+MAX_ITER ?= 10
 
 all: database
 
 $(SPECIMENS_OK):
 	mkdir -p build
-	bash ${XRAY_DIR}/utils/top_generate.sh $(subst /OK,,$@)
+	if [ -f `pwd`/generate.sh ]; then export $(ENV_VAR); bash `pwd`/generate.sh $(subst /OK,,$@); else bash ${XRAY_DIR}/utils/top_generate.sh $(subst /OK,,$@); fi
```
litghost: FUZDIR over pwd?

litghost:
This is probably the wrong solution.  pushdb can fail for other reasons, and this may cause an infinite loop. A narrow check should be added to fix #534, specifically that the local db is good to push.  This moves the retry to a step before pushdb.

Also consider adding a max iteration count to prevent infinite loops.

fuzzers/fuzzer.mk
acomodi: They should behave in the same way, probably it's better to stick with FUZDIR for coherency

fuzzers/fuzzer.mk
```diff
@@ -1,17 +1,27 @@
...
+	# If the database presents errors or is incomplete, the fuzzer is rerun.
+	# When it reaches the maximum number of iterations it fails.
+	@if [ $(ITER) -gt $(MAX_ITER) ]; then \
+		echo "Max Iterations reached. Fuzzer unsolvable."; \
+		exit 1; \
+	fi
 	$(MAKE) clean
 	$(MAKE) database
-	$(MAKE) pushdb
+	$(MAKE) pushdb || $(MAKE) $(VAR) ITER=$$(($(ITER) + 1)) run
```
acomodi: @litghost the max_iteration check is performed here in the run target

acomodi:
Should it be worth then to add a way to get error codes when checking the db consistency so to act accordingly with respect to the error received?
I agree that there are other reasons for which the pushdb fails, but the problem is that I couldn't find a simpler and cleaner way to perform this fix that could be used by all 01x fuzzers.

litghost:
I think the key is to separate the error checking in mergedb.  Maybe add a checkdb.py and call that on the local version of the database?

litghost:
It looks like running mergedb.py with only the input will work.

acomodi:
> I think the key is to separate the error checking in mergedb. Maybe add a checkdb.py and call that on the local version of the database?

Yeah, I am doing something similar, but directly in the clb.mk, If it compiles correctly I'll push the new version

fuzzers/fuzzer.mk
```diff
@@ -1,12 +1,15 @@
 N ?= 1
 SPECIMENS := $(addprefix build/specimen_,$(shell seq -f '%03.0f' $(N)))
 SPECIMENS_OK := $(addsuffix /OK,$(SPECIMENS))
+ENV_VAR ?=
+FUZDIR ?= ${PWD}
 
 all: database
 
 $(SPECIMENS_OK):
+	echo ${ENV_VAR}
```
litghost: Remove debug echo?  If you want to keep it, provide context for what is being echo'd

PR #552, opened 2019-01-18
Add utility to compare JSON from two database's.

litghost:


utils/diff_db_json.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash -e
+
+# diff_db_json.sh <database a> <database b>
```
mithro: Again, this feels like it would be better as a Python script?

I don't really care "that" much however.

litghost: Not sure I agree.  The only advantage that a Python script would give us is better temp cleanup in the error case.  Most of this is file redirection and calling subprocesses, which is what bash scripts are good at.

litghost: I'll try a python conversion and see what it looks like

PR #553, opened 2019-01-18
Add utility for diffing segbits from two databases.

litghost:


utils/diff_db_bits.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash -e
+
+# diff_db_bits.sh <database a> <database b>
```
mithro: This feels better as a Python script?

PR #555, opened 2019-01-18
Adding modular arguments for parallel Fuzzer 072

natsfr:
This commit update the Fuzzer 072, adding arguments to control the number of parallel process from Makefile.
It should be close to 074.
For information the Zynq version of this fuzzer run in  15 to 20 minutes on 4 core. Need to be tested before merging.


natsfr:
Pushed a fix, the fuzzer now works correctly when total number of item to iterate is an integer multiple of the number of block.

Issue #558, opened 2019-01-21
Zynq ROI does not include IO column


kgugala:
while this https://github.com/SymbiFlow/prjxray/pull/557 fixed 058 it broke 005. This is because the new ROI does not include any IO. In the zynq part we use all the IOs are on the right side so there is no option to include them in ROI without including hclks after hclk_brk. Including those hclk's breaks 058.

litghost:
For now, does it make sense to have 058 to exclude tiles?

Issue #469, opened 2019-01-09
tilegrid: IOB INT fuzzer


JohnDMcMaster:
we need to create the base address for the INT that's near IOB. Since there's no CLB to reference to, we must do something special. @acomodi is experimenting right now with doing something similar for Zynq that we can probably use as an example

Issue #219, opened 2018-11-01
Should BRAM_INT_INTERFACE have address info?


mcmasterg:
These appear to be completely passive with just wires

Ex entry:
```
"BRAM_INT_INTERFACE_L_X6Y102": {
    "bits": {
        "CLB_IO_CLK": {
            "baseaddr": "0x00020300",
            "frames": 28,
            "height": 2,
            "offset": 4,
            "words": 2
        }
    },
    "grid_x": 20,
    "grid_y": 49,
    "segment": "SEG_BRAM2_L_X6Y100",
    "segment_type": "bram2_l",
    "sites": {},
    "type": "BRAM_INT_INTERFACE_L"
},
```

Issue #342, opened 2018-12-18
tilegrid: BRAM breaks on full part


mcmasterg:
There are some missing tiles for GTX, PMV or something like that. This breaks the current tilegrid algorithm. Suggest we switch to the new solve technique that IOB is using to eliminate this entirely.

JohnDMcMaster:
More details...

Background (old method): The traditional method works by creating a reference bitstream and then toggling one bit in an IP block we want a base address for. Then we propagate those base addresses for IP blocks we can easily guess the base address for. For example, we toggle a single LUT bit in a CMT column in the CLB tile at the bottom of the column. We diff this new bitstream against the original btistream before the bit was toggled. This should give a single bit changed, exactly locating this bit. However, this only solves 1 / 50 CLB tile base addresses in a CMT column. A second fixup step does some fragile logic to guess addresses of other CLBs in that tile, taking into account things like a jump around HCLK. While this logic is fragile, subsequent fuzzers will probably fail if bad addresses are generated.

Background (new method): the new technique works by segmatching (statistical constraint solver) base addresses. This is similar to how most fuzzers solve for configuration bits. Unlike the original method, all IP blocks are changed in parallel. This means that log(N, 2) bitstreams are required instead of N to get the same number of base addresses. However, as part of this conversion, address propagation is also generally discouraged. So it is also recommended to toggle bits in all IP blocks in a column, not just one. In the CLB case, this means that every single CLB tile gets one bit set to a random value. If there are 4 CLB CMT columns in the ROI, this means we'd randomly set 4 * 50 bits in every bitstream. With 200 bits, it takes minimum floor(log(200, 2)) => 8 bitstreams (specimens) to solve all of them.

Okay, so what would this mean mechanically?
- Find a tilegrid fuzzer to copy. Suggest "monitor"
- Enter your copied directory
- Edit top.py
  - Consult the Xilinx 7 Series Library guide and/or Vivado layout to understand the primitive you need to instantiate
  - Find a single bit parameter that can be easily toggled, such as a clock inverter or a bulk configuration bit
  - Find the correct site type in gen_sites()
  - Instantiate the correct verilog library macro in top
  - LOC it, if necessary. Its necessary to LOC it if there is more than one
  - mcmaster: I usually base my instantiation on yosys/techlibs/xilinx and fpgaperf wrapper generator 
[...]

PR #562, opened 2019-01-24
Expand tilegrid to entire grid.

litghost:
With this PR, tilegrid now solves 100% for the following tiles on a7 and z7 settings:

- CLBL[LM]_[LR]
- HCLK_[LR]
- INT_[LR]
- BRAM_[LR]
- DSP_[LR]

On k7, most INT tiles are solved, expect for column X24 and X25 (#566).


JohnDMcMaster:
are you aware of https://github.com/SymbiFlow/prjxray/issues/342

litghost:
Yes, I'm already thinking about converting BRAM over to the fuzzer approach.

JohnDMcMaster:
1) I would suggest merging that before this to avoid breaking BRAM 2) IMHO team needs to have a higher level discussion about current goals. This PR hurts what I consider to be MVP

fuzzers/005-tilegrid/generate_full.py
```diff
@@ -468,6 +472,9 @@ def process_BLOCK_RAM(wordbase):
 
                     # FIXME: PCIE block cuts out some BRAM
                     # this messes up algorithm  as is and may cause this to fail
+                    if grid_y < 0:
```
JohnDMcMaster: IIRC this isn't the right fix. It will generate a few tiles with bad addresses

litghost:
I've successfully moved CLB and INT tiles for CLB, BRAM and IOB over to the fuzzer approach.  I need to add DSP, HCLK and DSP INT fuzzing to match what is currently in prxray-db.  The fuzzer approach is working nicely for things other than INT tiles.  INT tiles require some fairly routing specific logic, but it does demostrate that INT tiles appear to use the same bits regardless of whether it is attached to an CLB, IOB or BRAM.

litghost:
@acomodi - The output of this PR populates all INT tiles I believe for zynq7.  In particular, it adds BRAM and DSP INT fuzzers which appears to add enough known INT tiles to solve the whole grid.  Kintex7 needs more work, as two columns remain unsolved.

litghost:
Fixes #558.

litghost:
Fixes #469.

litghost:
Fixes #219.

litghost:
This new tilegrid works for most fuzzers, but 100-dsp-makpat stopped working.  Investigating.

litghost:
> I also see that the ps7_int fuzzer is actually not needed anymore because the INT_L base addresses are calculated from the adjacent INT_R. Therefore, I think the ps7_int can now be removed

I like the fact that there are more fuzzers than strictly speaking required.  It validates the the inputs are correct.   For example, I found a bug during development when the propagation found conflicting offsets in two instances.  Both cases were bugs in the fuzzers, rather than irregular INT columns.

litghost:
Found and fixed issue where "height" was being used instead of "words".

PR #565, opened 2019-01-26
New `make formatdb` target

mithro:
Adding a new `make formatdb` target to the Makefile and needed utils. The new target will,
 * Canonicalize the database by running `./utils/sort_db.py`
 * Regenerate the [`Info.md` file](database/Info.md)

### DIFF
--- a/utils/sort_db.py
+++ b/utils/sort_db.py
@@ -1,0 +1,352 @@
+#!/usr/bin/env python3
+"""
+Canonicalize the Project X-Ray database files by sorting. The aim is to reduce
+the diff output between runs to make it clearer what has changed.
+
+DB Files
+--------
+
+DB files are sorted into "natural" ordering. This is generally the order that a
+human would sort them in rather than how they sort as ASCII strings.
+
+For example with tags, a sequence of ABC1 to ABC12 would have the ASCII sort
+order of;
+
+    ABC1
+    ABC10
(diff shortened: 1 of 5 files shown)


utils/sort_db.py
```diff
@@ -0,0 +1,352 @@
...
+    """Sort the bit section of a segbit line.
+
+    >>> segbit_line_sort_bits("A !28_35 !27_39 27_37")
+    'A 27_37 !27_39 !28_35'
+
+    >>> segbit_line_sort_bits("B !28_35 !27_39 !27_37")
+    'B !27_37 !27_39 !28_35'
+
+    >>> segbit_line_sort_bits("C 28_35 00_00 !27_37")
+    'C 00_00 !27_37 28_35'
```
litghost: No, put negative bits first or last.  Do not intermix.

mithro: See reason in docstring of whole module.

litghost: But 1 hot encoding are relatively rare.  The bits that are set are more important than the bits that are cleared.

mithro: It also makes it hard to see when a diff has a bit flip.

mithro: Also, when you sort this way, a number of physical structures become much clearer (specially around chained muxes).

PR #576, opened 2019-01-29
Fuzzers: 071: add missing run.ok target

kgugala:
071 was missing run.ok target so it was not run from the top level makefile


litghost:
Root level makefile calls "$$(MAKE) -C $(1) run", so this target is not required?
https://github.com/SymbiFlow/prjxray/blob/master/fuzzers/Makefile#L16

kgugala:
yes, I must have miss-looked that

PR #577, opened 2019-01-29
Set bit2fasm.py as executable and increase context on JSON diff.

litghost:


utils/diff_db_json.py
```diff
@@ -40,7 +40,7 @@ def main():
 
             print("Comparing {}".format(a_json_base))
             subprocess.call(
-                "diff -U 3 {} {}".format(a_json_out.name, b_json_out.name),
+                "diff -U 10 {} {}".format(a_json_out.name, b_json_out.name),
```
mithro: We should consider if there is a way to use `--show-function-line=RE` argument to extract useful information;
```
       -F, --show-function-line=RE
              show the most recent line matching RE
```

litghost: Example?

mithro: BTW Not saying we should block merging this pull request on anything here.

PR #581, opened 2019-01-30
BRAM Cascade pip fuzzer.

### LINKED ISSUE #477: BRAM interconnect muxes are missing
There are pips that switch between the RAM18 and RAM36 that are not documented.  Example:
BRAM_L.BRAM_LOGIC_OUTS_B10_3.BRAM_RAMB18_DOADO3

Example pip junction: BRAM_L_X6Y110/BRAM_LOGIC_OUTS_B10_3

Which has the following pips:
BRAM_L_X6Y110/BRAM_L.BRAM_RAMB18_DOADO3->BRAM_LOGIC_OUTS_B10_3
BRAM_L_X6Y110/BRAM_L.BRAM_FIFO36_DOADOU3->BRAM_LOGIC_OUTS_B10_3

litghost:
PIP fuzzer that solves BRAM cascade junction box.  Fixes #477


litghost:
Fuzzer is easily solving everything except the bottom two address lines.  The bottom address lines are fairly correlated, and may just need a high sample count to resolve.

fuzzers/060-bram-cascades/generate.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
```
JohnDMcMaster: Can this use the general utils generate.sh? I'm not sure if I see anything special here other than the grep? And that looks like just for logging

JohnDMcMaster: ah I see theres actually a different print scheme, nvm

fuzzers/060-bram-cascades/Makefile
```diff
@@ -0,0 +1,38 @@
...
+# Specimens from current run must complete, but previous iterations may exist
+database: $(SPECIMENS_OK)
+	${XRAY_SEGMATCH} $(SEGMATCH_FLAGS) -o build/segbits_bram_l.rdb $(shell find build -name segdata_bram_l.txt)
+	${XRAY_SEGMATCH} $(SEGMATCH_FLAGS) -o build/segbits_bram_r.rdb $(shell find build -name segdata_bram_r.txt)
+
+	# Keep a copy to track iter progress
+	# Also is pre-fixup, which drops and converts
+	cp build/segbits_bram_l.rdb build/$(ITER)/segbits_bram_l.rdb
+	cp build/segbits_bram_r.rdb build/$(ITER)/segbits_bram_r.rdb
+	${XRAY_DBFIXUP} --db-root build \
```
JohnDMcMaster: Are the bits in BRAM_L different than BRAM_R? Traditionally pip fuzzers separate l vs r for solving, but configuration fuzzers solve them together. I'm not sure why this is the case though, this convention was inherited from Clifford

fuzzers/060-bram-cascades/top.py
```diff
@@ -0,0 +1,221 @@
...
+
+
+def max_address_bits(width):
+    return math.floor(math.log(float(MAX_BRAM * 36 * 1024) / width, 2))
+
+
+def random_sdp_bram(luts, name, modules, lines):
+    sdp_choices = set()
+
+    for width in (1, 2, 4, 8, 16, 18, 32, 36):
```
JohnDMcMaster: SDP has width 72 i think?

fuzzers/pip_loop.mk
```diff
@@ -0,0 +1,75 @@
+TODO_N ?= 50
```
JohnDMcMaster: What does this do differently vs int_loop?

utils/dbfixup.py
```diff
@@ -154,6 +153,12 @@ def add_zero_bits(
                     zero_range(tag, bits, 22, 25)
                 zero_groups(tag, bits, zero_db, strict=strict, verbose=verbose)
 
+                if len(bits) == 0:
+                    verbose and print(
```
JohnDMcMaster: Should this be ERROR, not WARNING? A WARNING to me is something that wasn't given enough data. In this case the database is actually oversolved and thus invalid

JohnDMcMaster:
Re: why it can't solve certain PIPs, can you give some more details on the ones it can't solve? Do you know anything about them? For example, if you run segprint after solving everything you can. are there a lot of bits left? Can you make a reasonable guess as to the remaining pips based on that?

fuzzers/pip_loop.mk
litghost: int_loop database and pushdb targets are for int_l and int_r tiles only.  This pip fuzzer is for bram_l and bram_r, so it needs to define different database and pushdb targets.  I refactored the common parts of both into pip_loop.mk and int_loop.mk includes pip_loop.mk

litghost:
> Re: why it can't solve certain PIPs, can you give some more details on the ones it can't solve? Do you know anything about them? For example, if you run segprint after solving everything you can. are there a lot of bits left? Can you make a reasonable guess as to the remaining pips based on that?

It is now solving all PIPs.  There was a bug earlier where I was only assign address lines 9:0 instead of <address_bits>-1:0.  Once that was fixed, the upper bits were solving.  The low address lines are hard to solve because they are rarely toggled.

fuzzers/060-bram-cascades/Makefile
litghost: I answered this question in another message, but the bits are the same between BRAM_L and BRAM_R, however the pip names are different.  E.g


26_56  26_57  26_59 ,BRAM.BRAM_ADDRBWRADDRL1.BRAM_IMUX_ADDRBWRADDRL1
vs
26_56  26_57  26_59 ,BRAM.BRAM_ADDRBWRADDRL1.BRAM_R_IMUX_ADDRBWRADDRL1

With some string mangling I should be able to unify the fuzzing which will improve the numbers of samples required to solve all pips.

utils/dbfixup.py
litghost: It can be an ERROR on strict, I agree.  However, when the zero db is missing entries, treating this like an error makes it hard to boot strap a fuzzer.

fuzzers/060-bram-cascades/top.py
litghost: The loop below starts to gets to those values

fuzzers/060-bram-cascades/Makefile
litghost: Refactored generate.py to mangle the pips, and added unmanagle to regenerated the L and R versions.

PR #586, opened 2019-01-31
fuzzers: Makefile: add explicit all target

kgugala:


fuzzers/Makefile
```diff
@@ -8,6 +8,8 @@ FUZZONLY=N
 # Allow building without it
 BITONLY=N
 
+all:
```
litghost: Also need a .PHONY for all?

kgugala: I don't think so. In my tests this works fine without PHONY

litghost: Please add it, see https://www.gnu.org/software/make/manual/html_node/Phony-Targets.html

The reason it is working is there is not a file named all.

Try this:
```
make clean
touch all
make all
```
It should probably do nothing

litghost: Though I guess all has dependencies, so it might work out.

kgugala: actually all the fuzzer's run.ok files are `all` target dependencies.

mithro: You probably want to make the `all` target the default? 
```makefile
.DEFAULT_GOAL := all
```

PR #596, opened 2019-02-01
Adding wrapper to get understandable output.

mithro:
* Saves stdout / stderr to log files.
 * Prefixes all messages with fuzzer name.
 * Gives regular status updates.

### DIFF
--- a/database/Makefile
+++ b/database/Makefile
@@ -1,3 +1,32 @@
-clean:
-	rm -f $$(ls */* |grep -v settings.sh)
 
+DATABASE_FILES = *.csv *.db *.json *.yaml
+
+clean-artix7-db:
+	rm -f $(addprefix artix7/,$(DATABASE_FILES))
+
+clean-kintex7-db:
+	rm -f $(addprefix kintex7/,$(DATABASE_FILES))
+
+clean-zynq7-db:
+	rm -f $(addprefix zynq7/,$(DATABASE_FILES))
+
+clean-db: clean-artix7-db clean-kintex7-db clean-zynq7-db
+	@true
+
+clean: clean-db
+	@true
+
+.PHONY: clean-artix7-db clean-kintex7-db clean-zynq7-db clean-db clean
+
+reset:
+	git reset --hard
+
+.PHONY: reset
+
+update:
+	git stash
+	git fetch origin
+	git merge origin/master
+	git stash pop
+
+.PHONY: update
--- a/fuzzers/Makefile
+++ b/fuzzers/Makefile
@@ -1,3 +1,5 @@
+SELF_DIR := $(dir $(lastword $(MAKEFILE_LIST)))
+
 # Developer options for quicker runs
 # Run algorithm, but only a small number of iterations
 QUICK=N
@@ -11,11 +13,24 @@ BITONLY=N
 all:
 
 define fuzzer
+
+# $(1) - Fuzzer name/directory
+# $(2) - Space seperated list of dependencies for the fuzzer.
+
+# Make the all target depend on the run.ok file for the fuzzer.
 all: $(1)/run.ok
+
+# Make the clean target run `make clean` in the fuzzer's directory.
 clean::
 	$$(MAKE) -C $(1) clean
+	rm -f $(1)/*.log
+
+# Describe how to create the fuzzer's run.ok file.
+# This command must start with a + to tell make to pass the jobserver
+# parameters downwards.
 $(1)/run.ok: $(addsuffix /run.ok,$(2))
-	$$(MAKE) -C $(1) run
+	+@$(SELF_DIR)/run_fuzzer.py $(1)
+
 endef
 
 $(eval $(call fuzzer,000-init-db,))
--- a/fuzzers/Makefile
+++ b/fuzzers/Makefile
@@ -32,7 +32,7 @@ ifeq ($(VERBOSE),Y)
 
 # When verbose we just call make directory
 $(1)/run.ok: $(addsuffix /run.ok,$(2))
-	$(MAKE) $(1)
+	$$(MAKE) -C $(1) run
 
 else
 
--- a/fuzzers/Makefile
+++ b/fuzzers/Makefile
@@ -23,13 +23,25 @@ all: $(1)/run.ok
 # Make the clean target run `make clean` in the fuzzer's directory.
 clean::
 	$$(MAKE) -C $(1) clean
-	rm -f $(1)/*.log
+	rm -rf $(1)/logs
 
 # Describe how to create the fuzzer's run.ok file.
 # This command must start with a + to tell make to pass the jobserver
 # parameters downwards.
+ifeq ($(VERBOSE),Y)
+
+# When verbose we just call make directory
+$(1)/run.ok: $(addsuffix /run.ok,$(2))
+	$(MAKE) $(1)
+
+else
+
+# When not verbose, we use the run_fuzzer wrapper which will save the results
+# to log files.
 $(1)/run.ok: $(addsuffix /run.ok,$(2))
 	+@$(SELF_DIR)/run_fuzzer.py $(1)
+
+endif
 
 endef
 
--- a/fuzzers/run_fuzzer.py
+++ b/fuzzers/run_fuzzer.py
@@ -124,8 +124,11 @@ class Logger:
         running_for = time_log - self.time_start
         msg = msg.format(*args, **kw)
 
-        log_prefix = "{:s} - {:>5s}: ".format(
-            self.fuzzer, pretty_timedelta_str(running_for))
+        log_prefix = "{:s} - {:s} - {:>5s}: ".format(
+            time_log.isoformat(),
+            self.fuzzer,
+            pretty_timedelta_str(running_for),
+        )
 
         msg = "\n".join(log_prefix + x for x in msg.splitlines())
         print(msg, flush=True)
--- a/fuzzers/run_fuzzer.py
+++ b/fuzzers/run_fuzzer.py
@@ -365,39 +365,17 @@ def main(argv):
     assert os.path.exists(fuzzer_logdir)
 
     exit_code = -1
-    test_cases = []
     for retry_count in range(0, args.retries):
         logger.log('Running fuzzer attempt: {}', [retry_count])
-        exit_code, test_case = run_fuzzer(
+        exit_code = run_fuzzer(
             args.fuzzer,
             fuzzer_dir,
             fuzzer_logdir,
             logger,
         )
-        if test_case:
-            test_cases.append(test_case)
         if exit_code <= 0:
             break
         logger.log('WARNING: Fuzzer failed!')
-
-    if test_cases:
-        pretty = True
-        for i, tc in enumerate(test_cases):
-            tc.name = "Run {}".format(i)
-            if tc.stdout.startswith('file://'):
-                pretty = False
-                tc.stdout = "\n".join(
-                    last_lines(open(tc.stdout[7:]), 1000))
-
-            if tc.stderr.startswith('file://'):
-                pretty = False
-                tc.stderr = "\n".join(
-                    last_lines(open(tc.stderr[7:]), 1000))
-
-        ts = junit_xml.TestSuite(args.fuzzer, test_cases)
-        tsfilename = os.path.join(fuzzer_logdir, 'sponge_log.xml')
-        with open(tsfilename, 'w') as f:
-            junit_xml.TestSuite.to_file(f, [ts], prettyprint=pretty)
 
     return exit_code
 
@@ -410,7 +388,7 @@ def run_fuzzer(fuzzer_name, fuzzer_dir, fuzzer_logdir, logger):
     make_flags = os.environ.get('MAKEFLAGS', '')
     # Should run things?
     if not should_run_submake(make_flags):
-        return 0, None
+        return 0
 
     fuzzer_runok = os.path.join(fuzzer_dir, "run.ok")
     if os.path.exists(fuzzer_runok):
@@ -420,7 +398,7 @@ def run_fuzzer(fuzzer_name, fuzzer_dir, fuzzer_logdir, logger):
             "Skipping as run.ok exists (updated @ {})",
             last_modified.isoformat())
 
-        return 0, None
+        return 0
 
     time_start = datetime.utcnow()
     log("Starting @ {}", time_start.isoformat())
@@ -534,20 +512,36 @@ def run_fuzzer(fuzzer_name, fuzzer_dir, fuzzer_logdir, logger):
 
     time_end = datetime.utcnow()
 
+    error_log = "\n".join(last_lines(open(fuzzer_stderr), 10000))
+    success_log = "\n".join(last_lines(open(fuzzer_stdout), 100))
+
+    # Find the next X_sponge_log.xml file name...
+    for i in range(0, 100):
+        tsfilename = os.path.join(fuzzer_logdir, '{}_sponge_log.xml'.format(i))
+        if not os.path.exists(tsfilename):
+            break
+
     test_case = junit_xml.TestCase(
         name=fuzzer_name,
         timestamp=time_start.timestamp(),
         elapsed_sec=(time_end - time_start).total_seconds(),
-        stdout='file://' + fuzzer_stdout,
-        stderr='file://' + fuzzer_stderr,
+        stdout=success_log,
+        stderr=error_log,
     )
+
+    if retcode != 0:
+        test_case.add_failure_info(
+            'Fuzzer failed with exit code: {}'.format(retcode))
+
+    with open(tsfilename, 'w') as f:
+        ts = junit_xml.TestSuite(fuzzer_name, [test_case])
+        junit_xml.TestSuite.to_file(f, [ts])
 
     if retcode != 0:
         test_case.add_failure_info(
             'Fuzzer failed with exit code: {}'.format(retcode), )
 
         # Log the last 10,000 lines of stderr on a failure
-        error_log = "\n".join(last_lines(open(fuzzer_stderr), 10000))
         log(
             """\
 Failed @ {time_end} with exit code: {retcode}
@@ -563,7 +557,6 @@ Failed @ {time_end} with exit code: {retcode}
     else:
 
         # Log the last 100 lines of a successful run
-        success_log = "\n".join(last_lines(open(fuzzer_stdout), 100))
         log(
             """\
 Succeeded! @ {}
@@ -575,7 +568,7 @@ Succeeded! @ {}
 
     logger.flush()
     signal.signal(signal.SIGINT, old_sigint_handler)
-    return retcode, test_case
+    return retcode
 
 
 if __name__ == "__main__":
--- a/fuzzers/run_fuzzer.py
+++ b/fuzzers/run_fuzzer.py
@@ -2,6 +2,8 @@
 
 from datetime import datetime, timedelta
 
+import argparse
+import collections
 import io
 import os
 import os.path
@@ -93,7 +95,7 @@ class Logger:
     def __init__(self, fuzzer, time_start, padding):
         self.fuzzer = fuzzer + ' ' * (padding - len(fuzzer))
         self.time_start = time_start
-        self.queue = []
+        self.queue = collections.deque()
 
     def _now(self):
         return datetime.utcnow()
@@ -115,7 +117,7 @@ class Logger:
 
     def flush(self):
         while len(self.queue) > 0:
-            time_log, msg, args, kw = self.queue.pop(0)
+            time_log, msg, args, kw = self.queue.popleft()
             self._output(time_log, msg, args, kw)
 
     def _output(self, time_log, msg, args, kw):
@@ -126,7 +128,7 @@ class Logger:
             self.fuzzer, pretty_timedelta_str(running_for))
 
         msg = "\n".join(log_prefix + x for x in msg.splitlines())
-        print(msg)
+        print(msg, flush=True)
 
 
 def last_lines(f, num, blocksize=100):
@@ -169,7 +171,7 @@ def last_lines(f, num, blocksize=100):
     ['1', '22', '3333333', '']
 
     """
-    lines = ['']
+    lines = collections.deque([''])
     # Seek to the end of the file
     f.seek(0, os.SEEK_END)
     fpos = f.tell()
@@ -191,7 +193,7 @@ def last_lines(f, num, blocksize=100):
     if fpos == 0:
         lines.insert(0, '')
 
-    return lines[1:]
+    return list(lines)[1:]
 
 
 def get_usage():
@@ -261,7 +263,7 @@ def get_pstree(pid):
     # Special case segfaults.
     if 'Segmentation fault' in stderr:
         stderr = ''
-        stdout = '(result to big, pstree segfaulted.)'
+        stdout = '(result too big, pstree segfaulted.)'
 
     # If no error, just return stdout
     if not stderr:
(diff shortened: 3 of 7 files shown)


fuzzers/run_fuzzer.py
```diff
@@ -0,0 +1,369 @@
...
+    def log(msg, *a, **k):
+        logger.log(msg, a, k, flush=True)
+
+    # Need a signal handler on SIGCHLD otherwise get_resource doesn't return
+    # anything.
+    signal.signal(signal.SIGCHLD, lambda sig, frame: None)
+
+    fuzzer_dir = os.path.join(fuzzers, fuzzer)
+    assert os.path.exists(fuzzer_dir), fuzzer_dir
+    # FIXME: The fuzzer output directory should be different from the top level
```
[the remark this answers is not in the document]
mithro: The bigger problem is that various fuzzers delete and then recreate the build directory.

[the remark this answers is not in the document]
kgugala: maybe the script can create a `log` folder in each fuzzer?

fuzzers/run_fuzzer.py
```diff
@@ -0,0 +1,369 @@
...
+    os.fsync(stderr_fd)
+    stderr_fd.close()
+
+    stderr_fd = open(fuzzer_stderr, "w+")
+    stderr_fd.seek(0, os.SEEK_END)
+
+    p = None
+    try:
+        p = subprocess.Popen(
+            "make -C {} run".format(fuzzer_dir),
```
litghost: Use the make enviroment variable, not make directly.  This ensures flags (like -j and VERBOSE=1, etc) are passed down.

litghost: https://www.gnu.org/software/make/manual/html_node/Options_002fRecursion.html

fuzzers/run_fuzzer.py
```diff
@@ -0,0 +1,369 @@
...
+            stdin=subprocess.DEVNULL,
+            stdout=stdout_fd,
```
litghost: Why not use StringIO and not both with the filesystem?  If logs are less than 100 MB, then who cares about the in memory usage?

fuzzers/run_fuzzer.py
```diff
@@ -0,0 +1,369 @@
...
+            close_fds=True)
+
+        while True:
+            try:
+                retcode = p.wait(timeout=10)
+                p = None
+            except subprocess.TimeoutExpired:
+                retcode = None
+
+            if retcode is not None:
```
litghost: Why not just use Popen.communicate or Popen.wait?

litghost:
This seems more complicated than required?  Is there a simpler implementation?

mithro:
FYI This is what the output looks like;
```
012-clb-n5ffmux          - 3m21s: make---make---sh---bash---make---vivado.sh---vivado---loader---vivado
019-clb-ndi1mux          - 1m20s: Still running.                               
019-clb-ndi1mux          - 1m20s: make---make---sh---bash---make---vivado.sh---vivado---loader---vivado
018-clb-ram              - 1m30s: Still running.
018-clb-ram              - 1m30s: make---make---sh---bash---make---vivado.sh---vivado---loader---vivado
015-clb-nffmux           - 3m31s: Still running.                   
015-clb-nffmux           - 3m31s: make---make---sh---bash---make---vivado.sh---vivado---loader---vivado
016-clb-noutmux          - 3m31s: Still running.                                                            
016-clb-noutmux          - 3m31s: make---make---sh---bash---make---vivado.sh---vivado---loader---vivado         
012-clb-n5ffmux          - 3m31s: Still running.                                 
012-clb-n5ffmux          - 3m31s: make---make---bash---python3                                                                               
019-clb-ndi1mux          - 1m30s: Still running.                         
019-clb-ndi1mux          - 1m30s: make---make---sh---bash---make---vivado.sh---vivado---loader---vivado     
012-clb-n5ffmux          - 3m31s: Finishing (User:154s System:7s). 
012-clb-n5ffmux          - 3m31s: Succeeded! @ 2019-02-01T16:01:12.034116
012-clb-n5ffmux          - 3m31s: --------------------------------------------------------------------------
012-clb-n5ffmux          - 3m31s: r/local/google/home/tansell/github/SymbiFlow/prjxray/fuzzers/012-clb-n5ffmux'
012-clb-n5ffmux          - 3m31s: touch run.ok           
012-clb-n5ffmux          - 3m31s: make[1]: Leaving directory '/usr/local/google/home/tansell/github/SymbiFlow/prjxray/fuzzers/012-clb-n5ffmux'
012-clb-n5ffmux          - 3m31s:                                    
012-clb-n5ffmux          - 3m31s: --------------------------------------------------------------------------
018-clb-ram              - 1m40s: Still running.                                      
018-clb-ram              - 1m40s: make---make---sh---bash---make---vivado.sh---vivado---loader---vivado
015-clb-nffmux           - 3m41s: Still running.
015-clb-nffmux           - 3m41s: make---make---sh---bash---make---sh---python3
016-clb-noutmux          - 3m41s: Still running.
016-clb-noutmux          - 3m41s: make---make---sh---bash---make---sh---python3
019-clb-ndi1mux          - 1m40s: Still running.
019-clb-ndi1mux          - 1m40s: make---make---sh---bash---make---vivado.sh---vivado---loader---vivado
018-clb-ram              - 1m50s: Still running.
018-clb-ram              - 1m50s: make---make---sh---bash---make---sh---python3
015-clb-nffmux           - 3m51s: Still running.
015-clb-nffmux           - 3m51s: make---make---sh---bash---make---vivado.sh---vivado---loader---vivado---vivado---loader---parallel_synth_
016-clb-noutmux          - 3m51s: Still running.
016-clb-noutmux          - 3m51s: make---make---sh---bash---make---vivado.sh---vivado---loader---vivado---vivado---loader---parallel_synth_
019-clb-ndi1mux          - 1m50s: Still running.
019-clb-ndi1mux          - 1m50s: make---make---sh---bash---make---vivado.sh---vivado---loader---vivado
018-clb-ram              - 1m53s: Finishing (User:105s System:4s).
018-clb-ram              - 1m53s: Succeeded! @ 2019-02-01T16:01:33.604123
018-clb-ram              - 1m53s: --------------------------------------------------------------------------
018-clb-ram              - 1m53s: ory '/usr/local/google/home/tansell/github/SymbiFlow/prjxray/fuzzers/018-clb-ram'
018-clb-ram              - 1m53s: touch run.ok
018-clb-ram              - 1m53s: make[1]: Leaving directory '/usr/local/google/home/tansell/github/SymbiFlow/prjxray/fuzzers/018-clb-ram'
018-clb-ram              - 1m53s:
018-clb-ram              - 1m53s: --------------------------------------------------------------------------
015-clb-nffmux           - 4m01s: Still running.
015-clb-nffmux           - 4m01s: make---make---sh---bash---make---vivado.sh---vivado---loader---vivado---vivado---loader---parallel_synth_
```

fuzzers/run_fuzzer.py
mithro: Reading the `stdout` / `stderr` from a subprocess is *highly* complicated under Python due to the nature of how they block. IE If the stderr buffer gets full while reading from stdout your program will deadline.

fuzzers/run_fuzzer.py
mithro: You are unable to print out the fact the process is still running. This is an important feature as you have multiple fuzzers running at a time you want to understand which ones are still running.

fuzzers/run_fuzzer.py
mithro: (The only safe way is to use `Popen.communicate` but that doesn't allow the current status printing.

mithro:
Output for a single fuzzer run;
```
(env) tansell@tansell:~/github/SymbiFlow/prjxray/fuzzers$ rm -f 001-part-yaml/run.ok 001-part-yaml/*.log && make 001-part-yaml/run.ok
001-part-yaml            -    0s: Starting @ 2019-02-03T06:00:59.679062
001-part-yaml            -   10s: Still running.
001-part-yaml            -   10s: make---make---make---vivado.sh---vivado---loader---vivado---vivado---loader---parallel_synth_
001-part-yaml            -   20s: Still running.
001-part-yaml            -   20s: make---make---make---vivado.sh---vivado---loader---vivado
001-part-yaml            -   30s: Still running.
001-part-yaml            -   30s: make---make---make---vivado.sh---vivado---loader---vivado
001-part-yaml            -   40s: Still running.
001-part-yaml            -   40s: make---make---make---vivado.sh---vivado---loader---vivado---sh---xilcurl
001-part-yaml            -   47s: Finishing (User:49s System:2s).
001-part-yaml            -   47s: Succeeded! @ 2019-02-03T06:01:46.790568
001-part-yaml            -   47s: --------------------------------------------------------------------------
001-part-yaml            -   47s: touch run.ok
001-part-yaml            -   47s: make[1]: Leaving directory '/usr/local/google/home/tansell/github/SymbiFlow/prjxray/fuzzers/001-part-yaml'
001-part-yaml            -   47s: 
001-part-yaml            -   47s: --------------------------------------------------------------------------
(env) tansell@tansell:~/github/SymbiFlow/prjxray/fuzzers$ 
```

database/Makefile
```diff
@@ -1,3 +1,32 @@
-clean:
-	rm -f $$(ls */* |grep -v settings.sh)
 
+DATABASE_FILES = *.csv *.db *.json *.yaml
+
+clean-artix7-db:
+	rm -f $(addprefix artix7/,$(DATABASE_FILES))
+
+clean-kintex7-db:
+	rm -f $(addprefix artix7/,$(DATABASE_FILES))
```
litghost: Copy/paste error?

fuzzers/run_fuzzer.py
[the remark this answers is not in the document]
litghost: +1 a directory to contain these files

database/Makefile
mithro: Correct. Fixed.

mithro: Fixed.

mithro:
This change should be ready for another review now. It is working well on the Kokoro CI system and it is much easier to understand what is going on.

fuzzers/Makefile
```diff
@@ -11,11 +13,36 @@ BITONLY=N
...
+
+# Describe how to create the fuzzer's run.ok file.
+# This command must start with a + to tell make to pass the jobserver
+# parameters downwards.
+ifeq ($(VERBOSE),Y)
+
+# When verbose we just call make directory
 $(1)/run.ok: $(addsuffix /run.ok,$(2))
-	$$(MAKE) -C $(1) run
+	$(MAKE) $(1)
```
litghost: You do need double $$ here

fuzzers/Makefile
```diff
@@ -11,11 +13,36 @@ BITONLY=N
...
+# $(1) - Fuzzer name/directory
+# $(2) - Space seperated list of dependencies for the fuzzer.
+
+# Make the all target depend on the run.ok file for the fuzzer.
 all: $(1)/run.ok
+
+# Make the clean target run `make clean` in the fuzzer's directory.
 clean::
 	$$(MAKE) -C $(1) clean
+	rm -f $(1)/*.log
```
litghost: Should be writing logs to a log directory

fuzzers/run_fuzzer.py
```diff
@@ -0,0 +1,528 @@
...
+    >>> f = io.StringIO("11\n22\n33\n44\n55")
+    >>> last_lines(f, 2, 4)
+    ['44', '55']
+
+    >>> f = io.StringIO("1\n22\n3333333\n")
+    >>> last_lines(f, 100, 3)
+    ['1', '22', '3333333', '']
+
+    """
+    lines = ['']
```
litghost: Use a deque rather than a python list?
https://docs.python.org/2/library/collections.html#collections.deque

fuzzers/run_fuzzer.py
```diff
@@ -0,0 +1,528 @@
...
+    ... ''')
+    fuzz   -   10s: Line 1
+    fuzz   -   10s: Line 2
+    fuzz   -   10s: Line 3
+    """
+
+    def __init__(self, fuzzer, time_start, padding):
+        self.fuzzer = fuzzer + ' ' * (padding - len(fuzzer))
+        self.time_start = time_start
+        self.queue = []
```
litghost: deque?

fuzzers/run_fuzzer.py
```diff
@@ -0,0 +1,528 @@
...
+    a1, a5, a15 = os.getloadavg()
+    cpus = os.cpu_count()
+    return a1 / cpus * 100.0, a5 / cpus * 100.0, a15 / cpus * 100.0
+
+
+_pstree_features = None
+
+
+def get_pstree_features():
+    global _pstree_features
```
litghost: No globals.  Make into class

fuzzers/run_fuzzer.py
```diff
@@ -0,0 +1,528 @@
...
+        stderr=subprocess.PIPE,
+        stdin=subprocess.DEVNULL,
+        universal_newlines=True,
+    )
+    stdout, stderr = p.communicate()
+
+    # Special case segfaults.
+    if 'Segmentation fault' in stderr:
+        stderr = ''
+        stdout = '(result to big, pstree segfaulted.)'
```
litghost: too

fuzzers/run_fuzzer.py
```diff
@@ -0,0 +1,528 @@
...
+
+    # If no error, just return stdout
+    if not stderr:
+        return stdout
+
+    return "{}\n{}".format(stdout, stderr)
+
+
+def run_submake(make_flags):
```
litghost: Poorly named, maybe "should_run_submake"?

fuzzers/run_fuzzer.py
```diff
@@ -0,0 +1,528 @@
...
+    True
+    """
+    r = re.search(r'(?:^|\s)[^-]*(n|q)[^\s]*(\s|$)', make_flags)
+    if not r:
+        return True
+    return not bool(r.groups()[0])
+
+
+def main(argv):
+    fuzzers = os.path.abspath(os.path.dirname(__file__))
```
litghost: Do not rely on the file path of this file, instead take an argument

fuzzers/run_fuzzer.py
```diff
@@ -0,0 +1,528 @@
...
+
+    assert len(argv) == 2, argv
```
litghost: Use argparse and add documentation for tool and it's arguments

fuzzers/run_fuzzer.py
```diff
@@ -0,0 +1,528 @@
...
+            last_modified.isoformat())
+    else:
+        log("Starting @ {}", time_start.isoformat())
+
+    log_suffix = ".{}.log".format(time_start.isoformat())
+    fuzzer_stdout = os.path.join(fuzzer_out, "stdout" + log_suffix)
+    fuzzer_stderr = os.path.join(fuzzer_out, "stderr" + log_suffix)
+
+    # Make sure the stdout file exists
+    stdout_fd = open(fuzzer_stdout, "w")
```
litghost: Just use a context manager here

with open(fuzzer_stdout, "w") as stdout_fd:
   stdout_fd.write("Build started {}\n".format(time_start.isoformat()))
    stdout_fd.write("-" * 75)
    stdout_fd.write("\n")

fuzzers/run_fuzzer.py
```diff
@@ -0,0 +1,528 @@
...
+    stdout_fd.write("Build started {}\n".format(time_start.isoformat()))
+    stdout_fd.write("-" * 75)
+    stdout_fd.write("\n")
+    stdout_fd.flush()
+    os.fsync(stdout_fd)
+    stdout_fd.close()
+
+    stdout_fd = open(fuzzer_stdout, "w+")
```
litghost: Why not use a+?

fuzzers/run_fuzzer.py
```diff
@@ -0,0 +1,528 @@
...
+    stdout_fd.seek(0, os.SEEK_END)
+
+    # Make sure the stderr file exists
+    stderr_fd = open(fuzzer_stderr, "w")
+    stderr_fd.write("Build started {}\n".format(time_start.isoformat()))
+    stderr_fd.write("-" * 75)
+    stderr_fd.write("\n")
```
litghost: Don't write anything to the stderr file.  Eventually we should consider check that this file is empty?

fuzzers/run_fuzzer.py
```diff
@@ -0,0 +1,528 @@
...
+
+    # Prevent Ctrl-C so we exit properly...
+    signal.signal(signal.SIGINT, lambda sig, frame: logger.log("Dieing!"))
+
+    try:
+        if p is not None:
+            try:
+                retcode = p.wait(1)
+            except subprocess.TimeoutExpired:
+                retcode = -1
```
litghost: Log that a kill was required?

fuzzers/run_fuzzer.py
```diff
@@ -0,0 +1,528 @@
...
+            bits.append("{:02d}s".format(int(ts)))
+        else:
+            bits.append("{:d}s".format(int(ts)))
+
+    assert len(bits) > 0, d
+
+    return "".join(bits)
+
+
+class Logger:
```
litghost: Why have an additional output buffer at all?  print() is to a buffered writer?

fuzzers/Makefile
mithro: Fixed. (Also fixed the missing `-C $(1) run` bit).

fuzzers/run_fuzzer.py
mithro: Done.

fuzzers/run_fuzzer.py
mithro: Fixed.

fuzzers/run_fuzzer.py
mithro: Fixed.

fuzzers/run_fuzzer.py
mithro: I disagree. The path of the tool is significantly more likely to be right than any given argument. You can't get it wrong currently.

fuzzers/run_fuzzer.py
mithro: You can't use `print()` (or any io function) in a signal handler.

fuzzers/run_fuzzer.py
mithro: Done.

fuzzers/Makefile
mithro: The `run.ok` should probably be in the same directory as the logs?

fuzzers/run_fuzzer.py
mithro: Done.

fuzzers/run_fuzzer.py
litghost: And you can't call malloc either, so what's you point?  This code is doubly wrong then?

fuzzers/run_fuzzer.py
mithro: Done.

fuzzers/run_fuzzer.py
mithro: I forgot the `a` open mode existed... Fixed.

fuzzers/run_fuzzer.py
mithro: Done.

fuzzers/Makefile
litghost: I'm fine with that.

fuzzers/run_fuzzer.py
mithro: The goal with the header was to make sure you have the matching stderr / stdout files.

fuzzers/Makefile
mithro: FYI - Done, will think about `run.ok` in the future...

fuzzers/run_fuzzer.py
mithro: I reworked this code to make that much clear.

fuzzers/run_fuzzer.py
litghost: Putting it on a non-preallocated queue doesn't solve the problem.  List,append can result in a memory allocation, which is not safe in a signal handler either.

mithro: > A Python signal handler does not get executed inside the low-level (C) signal handler. Instead, the low-level signal handler sets a flag which tells the virtual machine to execute the corresponding Python signal handler at a later point (for example at the next bytecode instruction).

This means that memory allocation is fine in a Python signal handler. At first glance this also makes it seem like IO should be also be fine. However, there are a bunch of complicated reasons why it is very *occasionally* not.

fuzzers/run_fuzzer.py
mithro: Done.

fuzzers/run_fuzzer.py
```diff
@@ -0,0 +1,528 @@
...
+    Values is average perecntage of CPU used over 1 minute, 5 minutes, 15
+    minutes.
+    """
+    a1, a5, a15 = os.getloadavg()
+    cpus = os.cpu_count()
+    return a1 / cpus * 100.0, a5 / cpus * 100.0, a15 / cpus * 100.0
+
+
+class PsTree:
+    _pstree_features = None
```
litghost: Why would you do it this way instead of regular methods and an instance object?

mithro: This is a pretty normal way to do a singleton style object in Python. 

Making this class needing to be instantiated adds a whole extra level of complexity.

litghost: But it isn't a singleton, and there is no need to it to be a singleton.

mithro: The pstree feature detection should only run once and is being cached. Caches like that are generally singletons.

litghost: Consider using functools.lru_cache of manual caching then?  This is more common and is thread safe in python 3.6 and above.  Alternatively make it an instance class, which makes clear that there is instance data.

mithro: lru_cache is not really designed for making a single call have permanent results;
> Decorator to wrap a function with a memoizing callable that saves up to the maxsize most recent calls. It can save time when an expensive or I/O bound function is periodically called with the same arguments.

I guess Python really should have something in this category included but it does not seem to...

litghost: @lru_cache(maxsize=None) should work here.

litghost:
```
import functools

@functools.lru_cache(maxsize=None)
def func():
    print('test')
    return 5

for _ in range(5):
    print(func())
```

Prints:
```
test
5
5
5
5
5
```

fuzzers/run_fuzzer.py
```diff
@@ -333,52 +333,70 @@ def should_run_submake(make_flags):
 
 
 def main(argv):
-    fuzzers = os.path.abspath(os.path.dirname(__file__))
+    fuzzers_dir = os.path.abspath(os.path.dirname(__file__))
 
     parser = argparse.ArgumentParser(description=__doc__)
     parser.add_argument("fuzzer", help="fuzzer to run")
+    parser.add_argument(
```
litghost: Probably shouldn't do this?  Flaky fuzzers should be fixed at the fuzzer level.

mithro: I generally agree, until all the fuzzers have been fixed this is a work around. I will log a bug to remove it. I am also hoping to mark the fuzzer as flaky in the test results output.

litghost: At least default it to no retries.  Then maybe have CI run it will higher retry count, but other users would see the flakyness?

Issue #12, opened 2017-12-22
Document the bitstream for the BRAM blocks in Artix-7


mithro:
The blockram parts are the last part needed to be documented to able to do a useful CPU design on the Artix-7.

mcmasterg:
I've started on this and made significant progress getting tilegrid and INIT fuzzer setup. More to come

mcmasterg:
just ran the full fuzzer for the first time and...there is an interleave issue. Only have of the bits fell out of the fuzzer. Probably related to BRAM18 vs BRAM36...will look into it tomorrow

JohnDMcMaster:
Yes. Basically:
- All of the bits that are needed for configuring core BRAM modes (ie excluding FIFO) should be decoded. This includes both bitstream sections
- Multibit issues should be fixed now (per mergedb changes last week)
- Interconnect fuzzer includes BRAM, but doesn't include all paths. I think it just needs a small tweak to instantiate additional primitives, but I need to poke at it

litghost:
The following bits are toggled by use of BRAM's, but are unknown bits:

```
# In frame 0x00400f1b 30 bits were not converted.
{ unknown_bit = "00400f1b_0_8" }
{ unknown_bit = "00400f1b_0_11" }
{ unknown_bit = "00400f1b_0_12" }
{ unknown_bit = "00400f1b_0_13" }
{ unknown_bit = "00400f1b_0_16" }
{ unknown_bit = "00400f1b_0_19" }
{ unknown_bit = "00400f1b_0_20" }
{ unknown_bit = "00400f1b_0_21" }
{ unknown_bit = "00400f1b_0_24" }
{ unknown_bit = "00400f1b_0_27" }
{ unknown_bit = "00400f1b_0_28" }
{ unknown_bit = "00400f1b_0_29" }
{ unknown_bit = "00400f1b_1_0" }
{ unknown_bit = "00400f1b_3_3" }
{ unknown_bit = "00400f1b_3_4" }
{ unknown_bit = "00400f1b_6_28" }
{ unknown_bit = "00400f1b_6_29" }
{ unknown_bit = "00400f1b_9_0" }
{ unknown_bit = "00400f1b_9_3" }
{ unknown_bit = "00400f1b_9_4" }
{ unknown_bit = "00400f1b_9_5" }
{ unknown_bit = "00400f1b_9_8" }
{ unknown_bit = "00400f1b_9_11" }
{ unknown_bit = "00400f1b_9_12" }
[...]

PR #597, opened 2019-02-01
Add FIFO bits and fix BRAM pips output

litghost:
Solves 26 of the 30 unknown bits in #12.

Also adds other new BRAM fuzzers to root makefile.


mithro:
LGTM but someone like @kgugala should review properly.

PR #600, opened 2019-02-02
Add block ram segbits support to prjxray FASM

litghost:


mithro:
Generally LGTM from a high level. Could you get @kgugala or @acomodi to review?

litghost:
BLOCK_RAM is actually specified in the bitsream manual, see table 5-24 in https://www.xilinx.com/support/documentation/user_guides/ug470_7Series_Config.pdf

PR #606, opened 2019-02-04
roi_harness: added zybo support

acomodi:
Fixed also basys3 settings to produce a correct harness for the artix


minitests/roi_harness/runme.tcl
```diff
@@ -433,15 +446,19 @@ if {$fixed_xdc eq ""} {
...
+    set y_right [expr {$Y_DIN_BASE + $Y_OFFSET}]
     for {set i 0} {$i < $DIN_N} {incr i} {
         # needed to force routes away to avoid looping into ROI
         if {[net_bank_left "din[$i]"]} {
             set node "INT_L_X0Y${y_left}/EE2BEG2"
             route_via2 "din_IBUF[$i]" "$node"
             set y_left [expr {$y_left + $PITCH}]
         } else {
-            set node "INT_R_X25Y${y_right}/WW2BEG1"
+            if {$part eq "xc7z010clg400-1"} {
```
litghost: Rather than a part driven if, just make a new env var?

minitests/roi_harness/basys3.sh
```diff
@@ -1,6 +1,8 @@
 # XC7A35T-1CPG236C
 export XRAY_PART=xc7a35tcpg236-1
 export XRAY_PINCFG=BASYS3-SWBUT
+export XRAY_DIN_N_LARGE=17
+export XRAY_DOUT_N_LARGE=17
```
kgugala: why 16? Originally the large target was using 8 ins and 8 outs

litghost: basys harness uses 16 LED, 16 switches, and a UART RX/TX (hence 17)

kgugala: ok, this seems to be correct. We have 16 LEDs/Swiches + UART

minitests/roi_harness/runme.tcl
acomodi: Done. Actually, I have tried to make it the more general as possible by adding more than an env var. I double checked it and it works fine

minitests/roi_harness/arty.sh
```diff
@@ -11,4 +11,10 @@ export XRAY_PIN_04="D13"
 export XRAY_PIN_05="J17"
 export XRAY_PIN_06="U14"
 
+# HCLK Tile
```
litghost: Arty should have some _INT_LR defines?

acomodi: Yep, forgot about that, done

PR #608, opened 2019-02-04
Add fuzzer for BRAM/FIFO enable bits.

### LINKED ISSUE #12: Document the bitstream for the BRAM blocks in Artix-7

litghost:
Fixes #12


mithro:
I don't know enough to *actually* review this... Best to get either @kgugala, @JohnDMcMaster or @acomodi to review?

fuzzers/029-bram-fifo-config/generate.py
```diff
@@ -0,0 +1,41 @@
+#!/usr/bin/env python3
+
+import json
+
+from prjxray.segmaker import Segmaker
+
+
+def bitfilter(frame, bit):
+    # Filter out interconnect bits.
+    if frame in [0, 1, 17, 23, 25]:
```
acomodi: Are these reference frame numbers fixed or only artix7 specific? More in general, where do they come from?

litghost: So far, the bits for a7, k7 and z7 have been the same, so this is likely not a7 specific, but only testing will give us confidence.

As for where the frame numbers come from:
https://github.com/SymbiFlow/prjxray-db/blob/master/artix7/segbits_int_l.db
https://github.com/SymbiFlow/prjxray-db/blob/master/kintex7/segbits_int_l.db

PR #612, opened 2019-02-05
Refactor INT fuzzers.

litghost:
The old tcl loop was actually quiet slow.  Just iterating over all nets and outputting pips outputs more data in less time (~15 min => ~5 min per sample).


litghost:
At -j56:
050 run time 47 min -> 25 mins
057 run time 38 min -> 9 mins

PR #614, opened 2019-02-06
iob: enhance fuzzer with bits for SLEW and DRIVE for LVCMOS25

tmichalak:


fuzzers/030-iob/generate.py
```diff
@@ -7,27 +7,42 @@
 
 print("Loading tags")
 '''
-port,site,tile,pin,val
+port,site,tile,pin,slew,drive,pulltype
```
litghost: Should have both iostandard and drive?

fuzzers/030-iob/generate.py
```diff
@@ -7,27 +7,42 @@
...
-        val = "NONE"
+    if pulltype == "":
+        pulltype = "NONE"
     segmaker.add_site_group_zero(
         segmk, site, "PULLTYPE.", ("NONE", "KEEPER", "PULLDOWN", "PULLUP"),
-        "PULLDOWN", val)
+        "PULLDOWN", pulltype)
 
+    segmaker.add_site_group_zero(
+        segmk, site, "LVCMOS25.DRIVE.", ("4", "8", "12", "16"), "12", drive)
```
litghost: Use the IOSTANDARD from the CSV?

fuzzers/030-iob/generate.tcl
```diff
@@ -53,20 +53,26 @@ proc loc_pins {} {
         set tile [get_tiles -of_objects $site]
         set pin [dict get $io_pin_sites $site]
 
-        set_property -dict "PACKAGE_PIN $pin IOSTANDARD LVCMOS33" $port
+        set_property -dict "PACKAGE_PIN $pin IOSTANDARD LVCMOS25" $port
```
litghost: Toggle between 2.5 and 3.3?

fuzzers/030-iob/bits.dbf
```diff
@@ -1 +1,3 @@
 38_98 39_97 39_99,LIOB33.IOB_Y1.PULLTYPE.PULLDOWN
```
litghost: What happened to IOB_Y0?

tmichalak: I think IOB_Y1 is used for inputs whereas IOB_Y0 for outputs. Drive and slew are properties for outputs only. Do you think we should have a separate entry in the dbf file for IOB_Y0 that is exactly as for IOB_Y1?

fuzzers/030-iob/generate.tcl
tmichalak: Well, the thing is that for LVCMOS33 we will solve the pulltype and slew bits, but not the drive - no clear zero value group problem.

fuzzers/030-iob/bits.dbf
litghost: That is incorrect.  IOB_Y1 and IOB_Y0 can be both inputs and outputs.

fuzzers/030-iob/generate.tcl
litghost: A zero value group is not required if it doesn't exist.

litghost:
Please rebase and re-test this PR on master.

fuzzers/030-iob/bits.dbf
tmichalak: I state corrected, I rebased the PR and now we have IOB_Y0 as well as IOB_Y1

PR #616, opened 2019-02-07
Non-canonical sorting for bit2fasm

litghost:


Makefile
```diff
@@ -7,7 +7,7 @@ env:
 	$(IN_ENV) pip install -r requirements.txt
 	$(IN_ENV) pip install -r docs/requirements.txt
 	ln -sf $(PWD)/prjxray env/lib/python3.*/site-packages/
-	ln -sf $(PWD)/third_party/fasm/fasm.py env/lib/python3.*/site-packages/
+	ln -sf $(PWD)/third_party/fasm/ env/lib/python3.*/site-packages/
```
mithro: Lets merge this, but we should probably be using something like `$(IN_ENV) pip install --develop third_party/fasm`.

litghost: It still is not passing, different symptom, similiar failure:

```
==================================== ERRORS ====================================
______________________ ERROR collecting utils/bit2fasm.py ______________________
utils/bit2fasm.py:8: in <module>
    import fasm.output
E   ModuleNotFoundError: No module named 'fasm.output'
 generated xml file: /home/travis/build/SymbiFlow/prjxray/build/py_test_results.xml 
!!!!!!!!!!!!!!!!!!! Interrupted: 1 errors during collection !!!!!!!!!!!!!!!!!!!!
=========================== 1 error in 0.50 seconds ============================
make: *** [test-py] Error 2
The command "make test-py" exited with 2.
```

PR #619, opened 2019-02-07
Fix frame initialization regression.

litghost:


mithro:
Needs `make format`, otherwise LGTM.

Issue #566, opened 2019-01-28
Kintex7 tilegrid missing some INT tiles


litghost:
INT column X24 and X25 are not current solved in 005-tilegrid for kintex7.

Issue #611, opened 2019-02-05
Collision between INT_L_X24Y112.INT_L.FAN_ALT7.GFAN1 and CLBLL_L_X24Y112.CLBLL_L.SLICEL_X0.CEUSEDMUX


mithro:
Latest build in the artix7 has the following issue;
```
Checking database/artix7
============================
Checking individual .db...
segbits_*.db: 14 okay
mask_*.db: 12 okay
Checking aggregate dir...
ERROR: 1 collisions
  00020c01_025_07: had INT_L_X24Y112.INT_L.FAN_ALT7.GFAN1, got CLBLL_L_X24Y112.CLBLL_L.SLICEL_X0.CEUSEDMUX
Traceback (most recent call last):
  File "utils/checkdb.py", line 136, in <module>
    main()
  File "utils/checkdb.py", line 132, in main
    run(args.db_root, strict=args.strict, verbose=args.verbose)
  File "utils/checkdb.py", line 118, in run
    check_tile_overlap(db, db_root, strict=strict, verbose=verbose)
  File "utils/checkdb.py", line 94, in check_tile_overlap
    raise ValueError("%s collisions" % len(collisions))
ValueError: 1 collisions
```

litghost:
Is this just another duplicate of #567?

mithro:
Seem like it might be...

litghost:
If CLBLL_L.SLICEL_X0.CEUSEDMUX is same as before, then this is just a dup of #567

acomodi:
As said in https://github.com/SymbiFlow/prjxray/issues/567, would it be worth re-testing with 054-pips-gfan fuzzer and see if the problem persists?

Issue #567, opened 2019-01-28
INT FAN_ALT[0-9].GFAN0 solutions are still unstable


litghost:
At some probability, the INT 056-rem fuzzer is generating bad solutions.  I've seen the following error at least once with the latest master:
```patch
-INT_R.FAN_ALT4.GFAN0 !22_08 !23_08 !24_08 21_08 25_08
+INT_R.FAN_ALT4.GFAN0 !00_10 !00_11 !01_09 !01_10 !22_08 !23_08 !24_08 01_14 21_08 25_08
```
01_14 is a bit for the GFAN0 input PIPs, not the FAN_ALT4 input PIPs.  Because 01_14 is very correlated to INT_R.FAN_ALT4.GFAN0, a dedicated fuzzer (rather than 056-rem) is a good idea.

litghost:
Should test to see if reverting https://github.com/SymbiFlow/prjxray/pull/518/files solves this?

kgugala:
054 was using the same algorithm for creating the test cases. Restoring it may help in a way that there will be more fuzzers trying to solve the same problem, so they may be more lucky. This, IMHO,  will not be a bulletproof solution.

litghost:
Agreed, a more targetted fuzzer makes sense given how correlated these bits are.

mithro:
This is still happening when running both Artix 7 and Zynq 7 fuzzers;
```patch
-INT_L.BYP_ALT1.GFAN0 20_15 !22_15 !23_15 24_15 !25_15
+INT_L.BYP_ALT1.GFAN0 !00_10 !00_11 !01_09 !01_10 01_14 20_15 !22_15 !23_15 24_15 !25_15
```

tmichalak:
Commit https://github.com/SymbiFlow/prjxray/commit/3c764a65e14d34aca0821a75d64ef401fe5e8f71  reverts the MAKETODO_FLAGS so that the 054-pip-fan-alt fuzzer only solves the FAN_ALT.GFANs. Before the revert it solved BYP_ALT.GFANs as well. Maybe this is a regression that is causing conflicts @mithro spotted ?
I am testing it in #728

mithro:
What was the solution here?

tmichalak:
[...]

PR #620, opened 2019-02-07
WIP: adding DSP for FAN_ALTs and BYPs to 050-pip-seed 

acomodi:
I am still running the build but it seems that all the FAN_ALT?.GFAN? are being correctly solved.

This is to solve https://github.com/SymbiFlow/prjxray/issues/611 and https://github.com/SymbiFlow/prjxray/issues/567

### DIFF
--- a/fuzzers/050-pip-seed/Makefile
+++ b/fuzzers/050-pip-seed/Makefile
@@ -84,7 +84,7 @@ run:
 	# Because 056 cannot solve IMUX and LOGIC_OUTS, force zero entries for the
 	# LOGIC_OUTS or IMUX entries.
 	XRAY_DIR=${XRAY_DIR} MAKE="$(MAKE)" QUICK=$(QUICK) $(XRAY_DIR)/fuzzers/int_loop.sh \
-		--check-args "--max-iters 12 --min-progress 5 --zero-entries --zero-entries-filter (IMUX.*GFAN)|(FAN_ALT.\.GFAN)"
+		--check-args "--max-iters 12 --min-progress 50 --zero-entries --zero-entries-filter (IMUX.*GFAN)|(FAN_ALT.*GFAN)"
 	touch run.ok
 
 clean:
--- a/fuzzers/050-pip-seed/Makefile
+++ b/fuzzers/050-pip-seed/Makefile
@@ -84,7 +84,7 @@ run:
 	# Because 056 cannot solve IMUX and LOGIC_OUTS, force zero entries for the
 	# LOGIC_OUTS or IMUX entries.
 	XRAY_DIR=${XRAY_DIR} MAKE="$(MAKE)" QUICK=$(QUICK) $(XRAY_DIR)/fuzzers/int_loop.sh \
-		--check-args "--max-iters 12 --min-progress 50  --zero-entries --zero-entries-filter (IMUX.*GFAN)"
+		--check-args "--max-iters 12 --min-progress 5 --zero-entries --zero-entries-filter (IMUX.*GFAN)|(FAN_ALT.\.GFAN)"
 	touch run.ok
 
 clean:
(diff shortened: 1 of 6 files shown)


acomodi:
The fuzzer correctly solves some of the FAN_ALT?.GFAN? pips, but not all of them. I am trying to do another build with a `min_progress` set to a lower value, because the build finished after iteration `3`. Probably I will need to add some other features to the `DSP` to get all the remaining FAN_ALT pips

litghost:
> The fuzzer correctly solves some of the FAN_ALT?.GFAN? pips, but not all of them. I am trying to do another build with a `min_progress` set to a lower value, because the build finished after iteration `3`. Probably I will need to add some other features to the `DSP` to get all the remaining FAN_ALT pips

Update https://github.com/SymbiFlow/prjxray/blob/master/fuzzers/050-pip-seed/Makefile#L87 to include "FAN_ALT?.GFAN?" in the required solution set.

So something like "--zero-entries-filter (IMUX.*GFAN)|(FAN_ALT.\.GFAN)"

litghost:
Looks like you are close, FAN_ALT1.GFAN1 is missing stimulus thought:

```
grep "FAN_ALT.*GFAN" 050-pip-seed/build/1/segbits_int_l.db 
INT.FAN_ALT0.GFAN0 <m1 6> 21_00 25_00
INT.FAN_ALT1.GFAN1 <m1 0> <const0>
INT.FAN_ALT2.GFAN0 <m1 2> 21_16 25_16
INT.FAN_ALT3.GFAN1 <8 candidates>
INT.FAN_ALT4.GFAN0 <m1 7> 21_08 25_08
INT.FAN_ALT5.GFAN1 <m1 1> 00_20 01_43 21_40 25_40
INT.FAN_ALT6.GFAN0 <m1 8> 21_24 25_24
INT.FAN_ALT7.GFAN1 21_32 25_32
```

acomodi:
Yeah, but it should get solved in the other iterations

litghost:
Looks like it wont, at least on artix7.

litghost:
> Looks like it wont, at least on artix7.

The reason for this is that all DSP tile within the ROI are DSP_R, so the INT_R solution converges with this PR, but not the DSP_L solution.  If I fix #566, then all INT tiles will have base addresses, and then including some DSP_L should make this work plus adding the change to the zero entries regexp.

I'm working on #566, half way done.

acomodi:
Yeah, unfortunately there is only one DSP column in the actual ROI. Anyway maybe I should produce more stimulus to be sure to get all those pips

litghost:
I'm running a test removing the ROI on a7, I expect it to solve all of the FAN_ALT?.GFAN? pips.  If so, that is the simplest answer.

litghost:
> I'm running a test removing the ROI on a7, I expect it to solve all of the FAN_ALT?.GFAN? pips. If so, that is the simplest answer.

Experiment failed.  I think what is happening is that the DSP is not providing the expected stimulus because it is using the internal DSP TIEOFF block, rather than the INT tile TIEOFF block.  DSP was maybe not a good choice :/

acomodi:
With the latest build I have produced almost all the FAN_ALT.GFANs (FAN_ALT1/3/5 are still missing from `segbits_int_l`. My settings were:  `--min-progress` argument to `10` instead of `50` and without the regexp. 
I don't think it was a bad choice after all. Maybe by tweaking some parameters we could force the routing to the INT TIEOFF. I will continue checking in this direction and see what I can get.

acomodi:
Unfortunately I cannot get all the ALT_FAN?.GFAN?. I think it should be best to do a dedicated fuzzer

litghost:
> Unfortunately I cannot get all the ALT_FAN?.GFAN?. I think it should be best to do a dedicated fuzzer

Sounds good.  Can you take that?

acomodi:
@litghost sure, no problem. I will make a 054 fuzzer replacing the old one that has been deleted a few weeks ago

acomodi:
This can be closed due to this other PR https://github.com/SymbiFlow/prjxray/pull/639

Issue #338, opened 2018-12-18
segmatch silently ignores missing files


mcmasterg:
I accidentally executed:

```
$ /home/mcmaster/prjxray/build/tools/segmatch -o build/segbits_ioblx.db build/specimen_001/segdata_ioblm_[lr].txt
Reading build/specimen_001/segdata_ioblm_[lr].txt.
#of segments: 0
#of bits: 0
#of tags: 0
#of const0 tags: 0
#of const1 tags: 0
```

Which isn't a real file. However, it silently ignored it not existing. We should at least warn, if not error on this

PR #630, opened 2019-02-08
segmatch: added warning for missing files

acomodi:
This PR is to solve https://github.com/SymbiFlow/prjxray/issues/338

### DIFF
--- a/tools/segmatch.cc
+++ b/tools/segmatch.cc
@@ -235,6 +235,12 @@ int main(int argc, char** argv) {
 			printf("Reading %s.\n", argv[optind]);
 			std::ifstream f;
 			f.open(argv[optind]);
+
+			// Check if input file exists.
+			if (!f.good()) {
+				printf("WARNING: Input file does not exist!\n");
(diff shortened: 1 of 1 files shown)


tools/segmatch.cc
```diff
@@ -235,6 +235,12 @@ int main(int argc, char** argv) {
 			printf("Reading %s.\n", argv[optind]);
 			std::ifstream f;
 			f.open(argv[optind]);
+
+			// Check if input file exists.
+			if (!f.good()) {
+				printf("WARNING: Input file does not exist!\n");
```
litghost: Add a return -1?

mithro:
Is there any reason this shouldn't just he a fatal error?

acomodi:
Most probably there is no reason. If the file is not there it means that there was a previous error which hasn't been caught. Another possibility is that the `Makefiles` could reference to a wrong file destination, in which case the tool would still need to produce a fatal error. I think you are right and it is better to return a fatal error here

Issue #583, opened 2019-01-31
`python3 utils/checkdb.py --db-root database/artix7` broken?


mithro:
With the latest database it seems like `utils/checkdb.py` is broken (or now takes a *very* long time).

litghost:
Is this still an issue?

acomodi:
I am looking at this issue. It seems that it takes infinite time because it iterates over all the tiles (`~18000` entries). I have checked it and it should never end up in an infinite loop. I am trying to see if it can be somehow optimized

acomodi:
This can be closed thanks to https://github.com/SymbiFlow/prjxray/pull/631

Issue #575, opened 2019-01-29
Check out of bound bits


acomodi:
There are some bits which happen to be out-of-bound, specifically for what regards the zynq7 db. There should be an additional control that checks whether there are some out of bound bits.

For example:
`INT_L.FAN_ALT0.LOGIC_OUTS_L22 !22_00 !23_00 !25_00 21_00 24_00 30_45` 

INT tiles have no more than 28 frames, and the last bit is clearly out of bound.

litghost:
#574 should prevent this from happening, but check db should also be augmented to detect it.

PR #631, opened 2019-02-08
checkdb: optimized script and added out-of-bound check

acomodi:
This PR is to solve two issues:

- https://github.com/SymbiFlow/prjxray/issues/583: I have noticed that all the tiles are being checked. I have also seen that all the tiles with the same `tile_type` share the same bits, making it highly time consuming if there is the need to check every single one of them. This way the script is much faster, but I am not sure it is the best solution, because maybe some tiles of the same type have differences.
- https://github.com/SymbiFlow/prjxray/issues/575: during the extraction of the bits there is an assertion that checks whether the bit is consistent with the `frames` number of the tile.

### DIFF
--- a/prjxray/util.py
+++ b/prjxray/util.py
@@ -1,7 +1,6 @@
 import os
 import re
 from .roi import Roi
-from prjxray.grid import BlockType
 
 
 def get_db_root():
@@ -198,8 +197,8 @@ def gen_tile_bits(tile_segbits, tile, strict=False, verbose=False):
             for bit in tile_segbits[block_type][tag]:
                 # 31_06
                 word_column, word_bit, isset = bit
-                assert word_column <= frames, "ERROR: bit out of bound --> word_column = %s; frames = %s" % (
-                    word_column, frames)
+                assert word_column <= frames, "ERROR: bit out of bound --> tag: %s; word_column = %s; frames = %s" % (
+                    tag, word_column, frames)
                 yield word_column + baseaddr, word_bit + bitbase, tag
 
 
--- a/prjxray/util.py
+++ b/prjxray/util.py
@@ -3,6 +3,7 @@ import re
 from .roi import Roi
 from prjxray.grid import BlockType
 
+
 def get_db_root():
     # Used during tilegrid db bootstrap
     ret = os.getenv("XRAY_DATABASE_ROOT", None)
@@ -184,7 +185,8 @@ def gen_tile_bits(tile_segbits, tile, strict=False, verbose=False):
     '''
 
     for block_type in tile_segbits:
-        assert block_type.value in tile["bits"], "block type %s is not present in current tile" % block_type.value
+        assert block_type.value in tile[
+            "bits"], "block type %s is not present in current tile" % block_type.value
 
         block = tile["bits"][block_type.value]
 
@@ -196,7 +198,8 @@ def gen_tile_bits(tile_segbits, tile, strict=False, verbose=False):
             for bit in tile_segbits[block_type][tag]:
                 # 31_06
                 word_column, word_bit, isset = bit
-                assert word_column <= frames, "ERROR: bit out of bound --> word_column = %s; frames = %s" % (word_column, frames)
+                assert word_column <= frames, "ERROR: bit out of bound --> word_column = %s; frames = %s" % (
+                    word_column, frames)
                 yield word_column + baseaddr, word_bit + bitbase, tag
 
 
--- a/prjxray/util.py
+++ b/prjxray/util.py
@@ -172,42 +172,24 @@ def addr2btype(base_addr):
     return block_type_i2s[block_type_i]
 
 
-def gen_tile_bits(db_root, tilej, strict=False, verbose=False):
+def gen_tile_bits(db_file, tilej, strict=False, verbose=False):
     '''
-    For given tile yield
+    For given tile and corresponding db_file structure yield
     (absolute address, absolute FDRI bit offset, tag)
 
-    For each address space
-    Find applicable files
-    For each tag bit in those files, calculate absolute address and bit offsets
+    db_file:
+    {'<block_type_1>': [<bit lines>], '<block_type_2>': [<bit lines>], ...}
 
-    Sample file names:
-    segbits_clbll_l.db
-    segbits_int_l.db
-    segbits_bram_l.block_ram.db
+    For each tag bit in the corresponding block_type entry, calculate absolute address and bit offsets
     '''
     for block_type, blockj in tilej["bits"].items():
+        assert block_type in db_file, "Block type is not in current tile"
+
         baseaddr = int(blockj["baseaddr"], 0)
         bitbase = 32 * blockj["offset"]
         frames = tilej["bits"][block_type]["frames"]
 
-        if block_type == "CLB_IO_CLK":
-            fn = "%s/segbits_%s.db" % (db_root, tilej["type"].lower())
-        else:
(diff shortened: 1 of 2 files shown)


utils/checkdb.py
```diff
@@ -64,6 +64,7 @@ def check_tile_overlap(db, db_root, strict=False, verbose=False):
     Throw an exception if two tiles share an address
     '''
     mall = dict()
+    tiles_type_done = set()
```
litghost: This will not work because it doesn't account for tile type mixes.  Instead consider another algorithm.  Examples that are immediately obvious:

- Why does make_tile_mask take a db_root directory?  This implies that it is not using the same prjxray.db.Database or prjxray.grid.Grid object, implying that json and db files are loaded repeatedly
- Can make_tile_mask be split into two parts, one that is tile type specific (e.g. do it once) and then is specialized for differennt tile base addresses and word offsets.?

acomodi: Yeah, that was my concern. Anyway okay, I will investigate and see if I can better optimize the script.

utils/checkdb.py
```diff
@@ -64,35 +91,47 @@ def check_tile_overlap(db, db_root, strict=False, verbose=False):
...
-        # for tile_name, tilej in subtiles(["CLBLL_L_X14Y112", "INT_L_X14Y112"]):
+        tile_type = tilej["type"]
+
+        if tile_type not in tiles_type_done:
+            db_files = parsedb_file(db_root, db_files, tilej)
+            tiles_type_done.add(tile_type)
+
+        if tile_type not in mall:
+            verbose and print("Adding tile type: %s" % tile_type)
+            mall[tile_type] = {}
```
litghost: Why not just use prjxray.db.Database and prjxray.grid.Grid?  Why rewrite the parsing code?

acomodi: Done, now the script uses the libraries.

acomodi:
I discovered that the comparison between the already checked `tiles` and the new ones was the bottleneck of the script. By modifying the comparison mechanism it should take much less now

prjxray/util.py
```diff
@@ -172,43 +172,34 @@ def addr2btype(base_addr):
...
-        elif not os.path.exists(fn):
-            continue
-
-        for line, (tag, bits, mode) in parse_db_lines(fn):
-            assert mode is None
-            for bitstr in bits:
+
+    for block_type in tile_segbits:
+        assert block_type.value in tile[
+            "bits"], "block type %s is not present in current tile" % block_type.value
```
litghost: Get rid of .value for block_type, not required.

prjxray/util.py
```diff
@@ -172,43 +172,34 @@ def addr2btype(base_addr):
...
+        baseaddr = int(block["baseaddr"], 0)
+        bitbase = 32 * block["offset"]
+        frames = block["frames"]
+
+        for tag in tile_segbits[block_type]:
+            for bit in tile_segbits[block_type][tag]:
                 # 31_06
-                _bit_inv, (bit_addroff, bit_bitoff) = parse_tagbit(bitstr)
-                yield (baseaddr + bit_addroff, bitbase + bit_bitoff, tag)
+                word_column, word_bit, isset = bit
```
litghost: Please used the named fields rather than expanding the tuple

prjxray/util.py
```diff
@@ -172,43 +172,34 @@ def addr2btype(base_addr):
     return block_type_i2s[block_type_i]
 
 
-def gen_tile_bits(db_root, tilej, strict=False, verbose=False):
+def gen_tile_bits(tile_segbits, tile, strict=False, verbose=False):
```
litghost: Move gen_tile_bits out of prjxray/util.py, it is only used by checkdb.py

prjxray/util.py
acomodi: Actually I think it is necessary, the block type in tile_segbits comes from an `enum`. If i do not get its value, this error will be produced.
 
```
  File "utils/checkdb.py", line 159, in <module>
    main()
  File "utils/checkdb.py", line 155, in main
    run(args.db_root, strict=args.strict, verbose=args.verbose)
  File "utils/checkdb.py", line 141, in run
    check_tile_overlap(db, db_root, strict=strict, verbose=verbose)
  File "utils/checkdb.py", line 96, in check_tile_overlap
    verbose=verbose)
  File "utils/checkdb.py", line 34, in make_tile_mask
    tile_segbits, tilej, strict=strict, verbose=verbose):
  File "/home/alessandro/projects/symbiflow/tmp/prjxray/env/lib/python3.6/site-packages/prjxray/util.py", line 188, in gen_tile_bits
    "bits"], "block type %s is not present in current tile" % block_type
AssertionError: block type BlockType.CLB_IO_CLK is not present in current tile
```

prjxray/util.py
acomodi: Ok, makes sense

utils/checkdb.py
```diff
@@ -64,35 +65,57 @@ def check_tile_overlap(db, db_root, strict=False, verbose=False):
...
+    tiles_type_done = dict()
+    tile_segbits = dict()
 
     tiles_checked = 0
 
-    def subtiles(tile_names):
-        for tile_name in tile_names:
-            yield tile_name, db.tilegrid[tile_name]
-
     for tile_name, tilej in db.tilegrid.items():
```
litghost: Use db.grid() rather than directly use the underlying data.

prjxray/util.py
litghost: Ya, that's because you are using the wrong data structure here.  Rather than using the tile, get it from Grid.gridinfo_at_tilename()

utils/checkdb.py
acomodi: Done

prjxray/util.py
acomodi: Done

prjxray/util.py
acomodi: Done

utils/checkdb.py
```diff
@@ -17,7 +17,35 @@
 import glob
 
 
-def make_tile_mask(db_root, tile_name, tilej, strict=False, verbose=False):
+def gen_tile_bits(tile_segbits, tile_bits, strict=False, verbose=False):
+    '''
```
litghost: These arguments are not used?

utils/checkdb.py
```diff
@@ -17,7 +17,35 @@
...
+                word_column = bit.word_column
+                word_bit = bit.word_bit
+                assert word_column <= frames, "ERROR: bit out of bound --> tag: %s; word_column = %s; frames = %s" % (
+                    tag, word_column, frames)
+                yield word_column + baseaddr, word_bit + bitbase, tag
+
+
+def make_tile_mask(
+        tile_segbits, tile_name, tile_bits, strict=False, verbose=False):
     '''
```
litghost: These arguments are not used (once gen_tile_bits doesn't take them)?

acomodi: you mean `strict` and `verbose`? If yes, yhey can actually be removed

litghost: Ya, sorry.  I keep expecting highlighting to show in the comment

utils/checkdb.py
```diff
@@ -64,35 +92,59 @@ def check_tile_overlap(db, db_root, strict=False, verbose=False):
     Throw an exception if two tiles share an address
```
litghost: check_tile_overlap no longer needs the db_root argument

PR #632, opened 2019-02-08
WIP: Dup jobserver fd's for retry.

litghost:

PR #634, opened 2019-02-08
Update CI configuration:

litghost:
- Store output of logs rom run_fuzzer.py
 - Decrease number of processes running for 072 and 074 to lower memory load.

This PR should be merged after #632

### DIFF
--- a/.github/kokoro/db-full.sh
+++ b/.github/kokoro/db-full.sh
@@ -46,9 +46,9 @@ echo "----------------------------------------"
 	echo "----------------------------------------"
 
 	# Run the fuzzers
-	export MAX_VIVADO_PROCESS=$CORES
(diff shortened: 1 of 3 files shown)


.github/kokoro/db-full.sh
```diff
@@ -46,9 +46,9 @@ echo "----------------------------------------"
 	echo "----------------------------------------"
 
 	# Run the fuzzers
-	export MAX_VIVADO_PROCESS=$CORES
+	export MAX_VIVADO_PROCESS=$((CORES/2 < 20 ? CORES/2 : 20))
```
elmsfu: Why are we limiting to 20? Please add a comment to explain why it is using a ceiling and the choice of 20.

litghost: Done.

PR #636, opened 2019-02-08
Add RAM usage to output.

litghost:


fuzzers/run_fuzzer.py
```diff
@@ -474,9 +478,10 @@ def log(msg, *a, **k):
             if retcode is not None:
                 break
             log(
-                "Still running (1m:{:0.2f}%, 5m:{:0.2f}%, 15m:{:0.2f}%).\n{}",
+                "Still running (1m:{:0.2f}%, 5m:{:0.2f}%, 15m:{:0.2f}%).\n{}\n{}",
                 *get_load(),
                 PsTree.get(p.pid),
+                get_memory(),
             )
```
mithro: Can we keep the memory usage in the same line as the CPU info?
```suggestion
            log(
                "Still running (1m:{:0.2f}%, 5m:{:0.2f}%, 15m:{:0.2f}% {}).\n{}",
                *get_load(),
                get_memory(),
                PsTree.get(p.pid),
            )
```

PR #638, opened 2019-02-11
Add CLK HROW bits

litghost:


mithro:
Did some point in the past we already have this?

litghost:
> Did some point in the past we already have this?

Not to my knowledge.  To be clear, this is the HROW driver bits, not the HCLK bits.  Not the same thing.

fuzzers/040-clk-hrow-config/top.py
```diff
@@ -0,0 +1,79 @@
...
+                    random.choice(('SYNC', 'ASYNC')))
+
+                print(
+                    '''
+    (* KEEP, DONT_TOUCH, LOC = "{site}" *)
+    BUFHCE #(
+        .INIT_OUT({INIT_OUT}),
+        .CE_TYPE({CE_TYPE})
+        ) buf_{site} ();
+                    '''.format(**params))
```
elmsfu: I prefer templates like this not be defined inline, but recognize this is keeping with the pattern of existing fuzzers. Is there a reason to keep them inline?

fuzzers/040-clk-hrow-config/Makefile
```diff
@@ -0,0 +1,22 @@
+N ?= 15
+
+include ../fuzzer.mk
+
+database: build/segbits_clk_hrow.db
+
+build/segbits_clk_hrow.rdb: $(SPECIMENS_OK)
+	${XRAY_SEGMATCH} -o build/segbits_clk_hrow.rdb $(addsuffix /segdata_clk_hrow_top_r.txt,$(SPECIMENS)) $(addsuffix /segdata_clk_hrow_bot_r.txt,$(SPECIMENS))
```
elmsfu: Maybe define a variable for `$(addsuffix /segdata_clk_hrow_top_r.txt,$(SPECIMENS)) $(addsuffix /segdata_clk_hrow_bot_r.txt,$(SPECIMENS))` as it is used twice. If not wrap long lines. In general I think defining more variables helps makefile clarity.

fuzzers/041-clk-hrow-pips/clk_table.py
```diff
@@ -0,0 +1,30 @@
+HCLKS = 24
```
elmsfu: Can we wrap this in a function? Is there a reason this needs to be done on import? If so, please add a comment.

fuzzers/041-clk-hrow-pips/top.py
```diff
@@ -0,0 +1,293 @@
...
+
+def read_site_to_cmt():
+    with open(os.path.join(os.getenv('FUZDIR'), 'build',
+                           'cmt_regions.csv')) as f:
+        for l in f:
+            site, cmt = l.strip().split(',')
+            yield (site, cmt)
+
+
+CMT_RE = re.compile('X([0-9]+)Y([0-9]+)')
```
elmsfu: move to top with other REs

fuzzers/041-clk-hrow-pips/top.py
```diff
@@ -0,0 +1,293 @@
+import os
+import re
+import random
+random.seed(int(os.getenv("SEED"), 16))
+from prjxray import util
+from prjxray.db import Database
+
+XY_RE = re.compile('^BUFHCE_X([0-9]+)Y([0-9]+)$')
```
elmsfu: Is there a good place for this to be moved up as it is duplicated in fuzzers/040-clk-hrow-config/top.py

fuzzers/005-tilegrid/Makefile
```diff
@@ -96,6 +97,9 @@ cfg_int/build/segbits_tilegrid.tdb: build/basicdb/tilegrid.json
 orphan_int_column/build/segbits_tilegrid.tdb: build/basicdb/tilegrid.json
 	cd orphan_int_column && $(MAKE)
 
+clk_hrow/build/segbits_tilegrid.tdb: build/basicdb/tilegrid.json
```
elmsfu: I feel like all of these rules could be generalized with a pattern. I know it's not related to the change, is there an issue for makefile clean up or replacing make? I'm happy to open one.

fuzzers/041-clk-hrow-pips/clk_table.py
litghost: I assume you are referring to CLK_TABLE, and not `HCLKS = 24`?

fuzzers/005-tilegrid/Makefile
litghost: Ya, go ahead and file an issue.  My make foo isn't that strong, and it's "only" 4 lines per.

fuzzers/041-clk-hrow-pips/top.py
litghost: This is one of those fine lines between DRY and clarity.  A general site name to X/Y function is probably useful. I'll think about it.

fuzzers/040-clk-hrow-config/top.py
litghost: What is the advantage of moving them out of line?  That would appear to hurt readability?

elmsfu: I guess it's preference. The multi-line template really breaks readability of functions for me with python's whitespace syntax. Having the templates together would let me see the parts of the eventual output. But I can see those that prefer have the template where it is used.

PR #639, opened 2019-02-12
FAN_ALT.*GFAN pips fuzzer

acomodi:
This PR is to solve https://github.com/SymbiFlow/prjxray/issues/611 and https://github.com/SymbiFlow/prjxray/issues/567. I have run several builds to check that it produces correct results and it seems like it does.


fuzzers/054-pip-fan-alt/Makefile
```diff
@@ -0,0 +1,6 @@
+MAKETODO_FLAGS=--re "^INT_[LR].FAN_ALT.*GFAN"
+GENERATE_FLAGS=--todo ../todo.txt
+N = 48
+SEGMATCH_FLAGS=-m 30 -M 45 -c 5
```
litghost: Why -c 5?  -c is only set bits I believe?

acomodi: Yep, my mistake, it was a leftover for an ulterior check

PR #641, opened 2019-02-12
Add more CLK bits

litghost:
Adds bits and base addresses for:
 - Site BUFGCTRL
 - Tile CLK_BUFG_REBUF


fuzzers/005-tilegrid/clk_bufg/generate.tcl
```diff
@@ -0,0 +1,21 @@
+source "$::env(XRAY_DIR)/utils/utils.tcl"
```
acomodi: Maybe it is worth moving this tcl script to a `common` directory as many different fuzzers in 005-tilegrid share it?

litghost: Done

Issue #642, opened 2019-02-13
054-pip-fan-alt times out on zync7


litghost:
CI for artix7 passed with new 054-pip-fan-alt, but zync7 timed out after 50 iterations.

PR #645, opened 2019-02-13
Filter out FAN_ALT[0-9].GFAN[0-9] pips from 050.

litghost:


litghost:
Fixes #642

Issue #647, opened 2019-02-14
005-tilegrid sometimes failing


litghost:
On the CI, I've seen 005-tilegrid fail.  I've identified the problem as some non-determinism in tile order.  Fix is being working on.

PR #648, opened 2019-02-14
Sort tiles and revamp gen_fuzz_states to be more efficient.

litghost:
- gen_fuzz_states now generates all states in log2(N) + 2 samples
- Sorting tile names ensures deterministic matching between gen_fuzz_states and sites, which is required for deterministic solutions
- Lowers sample count to log2(Nmax) + 2 samples.  Nmax is set to max count across all 7-series parts
- Modified IOB and PLL fuzzer to be deterministic


litghost:
Please review at your leisure, committing this to get CI green.

litghost:
Fixes #647

Issue #646, opened 2019-02-13
Fuzzer 041 sometimes generates error


litghost:
ERROR: [Common 17-55] 'set_property' expects at least one object.
Resolution: If [get_<value>] was used to populate the object, check to make sure this command returns at least one valid object.

This is likely an edge case that was missed during testing.

litghost:
Replicates 100% of the time on zync

PR #649, opened 2019-02-14
Limit target pips on zync to avoid unsolvable bits.

litghost:
Zync part selected is missing some side clock connections, so solving those bits are impossible.  The selected artix and kintex parts do not have this issue.


litghost:
It is expected that CI will go green with this PR, waiting for that.

litghost:
Fixes #646

PR #650, opened 2019-02-14
Add required FASM features to harness

litghost:
Fixes issues where fasm2frames was not clearing some INT pips within the ROI, which uncovered a second issue where CLK bits were getting cleared.

After this PR is merged, harnesses will need to be regenerated.  Errors may be throw if the ROI specified overlaps with unknown bits, in which case the ROI must be shrunk, or the unknown bits documented.

In the long run, once all bits within the harness are documented, the required FASM features can be used to replace the design.bit, and a bitstream can be made soley from the output frames file.


utils/fasm2frames.py
```diff
@@ -51,9 +54,26 @@ def dump_frm(f, frames):
...
+    db = Database(db_root)
+    assembler = fasm_assembler.FasmAssembler(db)
+
+    if roi:
+        with open(roi) as f:
+            roi_j = json.load(f)
+        x1 = roi_j['info']['GRID_X_MIN']
+        x2 = roi_j['info']['GRID_X_MAX']
+        y1 = roi_j['info']['GRID_Y_MIN']
+        y2 = roi_j['info']['GRID_Y_MAX']
```
acomodi: Is the indentation correct here?

acomodi:
We should check whether the new `roi_harnesses` are compatible with this changes before merging though, I think?

litghost:
> We should check whether the new `roi_harnesses` are compatible with this changes before merging though, I think?

I've tested that the basys3 harness is compatible.

PR #656, opened 2019-02-14
segmatch.cpp: if f_in does not exits returns -1

acomodi:
This PR is to get en error if segmatch does not find the input file as suggested in https://github.com/SymbiFlow/prjxray/pull/630. This solves https://github.com/SymbiFlow/prjxray/issues/338

### DIFF
--- a/fuzzers/058-pip-hclk/Makefile
+++ b/fuzzers/058-pip-hclk/Makefile
@@ -4,10 +4,10 @@ SPECIMENS := $(addprefix build/specimen_,$(shell seq -f '%03.0f' $(N)))
 SPECIMENS_OK := $(addsuffix /OK,$(SPECIMENS))
 
 database: $(SPECIMENS_OK)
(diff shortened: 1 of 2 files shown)


litghost:
Please rebase this PR on master and push to let presubmit database build with this change against latest fuzzers.

litghost:
```
Reading build/specimen_002/segdata_hclk_l_design_*.txt.
ERROR: Input file does not exist!
```

This PR breaks 058, or exposes an issue in how it was written.

acomodi:
Yes, most probably there is an issue in how the fuzzer is written. I guess it is required that a file is present as input for segmatch

litghost:
> Yes, most probably there is an issue in how the fuzzer is written. I guess it is required that a file is present as input for segmatch

The fuzzer was written such that specimen_001 is only HCLK_L and specimen_002 is HCLK_R, hence why one has segdata.  A find invocation should resolve that?

acomodi:
Yes, that'll solve the issue

acomodi:
Fixes https://github.com/SymbiFlow/prjxray/issues/338.

Issue #658, opened 2019-02-14
Unstable output in INT fuzzers


litghost:
ERROR: 1 collisions
  00001518_058_16: had INT_L_X18Y78.INT_L.FAN_ALT1.SW2END2, got BRAM_L_X18Y75.BRAM_L.RAMB18_Y0.FIFO_MODE

acomodi:
What if we expand all the FAN_ALT* and not only the GFAN* in fuzzer 054-pip-fan-alt?

litghost:
Still determining if this is a bug with the INT or BRAM fuzzer.

litghost:
Looks like an instability in the BRAM FIFO config fuzzer:

BRAM_L.RAMB18_Y0.FIFO_MODE 24_240 27_150

acomodi:
Ok, probably more stimuli or stricter segmatch arguments are needed to 029

PR #662, opened 2019-02-15
029-bram-fifo: more specimens and added c arg to segmatch

acomodi:
this PR fixes https://github.com/SymbiFlow/prjxray/issues/658. I have run the fuzzer indefinitely for 1 hour waiting for a collision but it didn't fail. 

### DIFF
--- a/fuzzers/029-bram-fifo-config/Makefile
+++ b/fuzzers/029-bram-fifo-config/Makefile
@@ -1,11 +1,11 @@
-N ?= 5
+N ?= 20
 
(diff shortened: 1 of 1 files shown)


litghost:
-c doesn't work here, there are bits that have 2 bits.

acomodi:
Right, I didn't see the `IN_USE` ones. I will run it again with only more specimens for a couple of hours. If it fails again we need to find other stimuli.

acomodi:
After more than 50 iterations proved to be stable with 20 specimens

PR #667, opened 2019-02-19
Isolate IMUX<x>.GFAN[01] pips into seperate fuzzer.

litghost:
I've seen bad IMUX<x>.GFAN[01] solutions from 050, this is a simple fuzzer to isolate IMUX<x>.GFAN solving from 050.

This PR also removes the dependency of 050 on other INT fuzzers.  The dedicated INT fuzzers should not be working on PIPs that 050 is working on and vise-versa.


fuzzers/056-pip-rem/Makefile
```diff
@@ -1,6 +1,6 @@
 MAKETODO_FLAGS=--re ".*" --not-endswith ".VCC_WIRE"
 GENERATE_FLAGS=--todo ../todo.txt
-N = 48
+N = 24
```
acomodi: Is `24` enough for each iteration to get enough information per each iteration? (For kintex7 and zynq7 as well)

litghost: I was testing on kintex7.  24 should be more than enough actually.

litghost:
By not depending on 050, the INT fuzzers all race to write the piplist, which causes an explosion.  I know the fix.

acomodi:
Perfect, as I can see now the piplist is generated beforehand so that there is no `race condition` between the various fuzzers.
I would say that this `PIP` fuzzers are occupying a lot of "fuzzer numbers". I do not know if we will need more specific `int` fuzzers in the future, but it is possible we run out of places where to fit all of them. Would it be worth to move all of the `int` related fuzzers to a `int` fuzzer directory?

Issue #661, opened 2019-02-15
INT fuzzers randomly failing on CI


acomodi:
This happens with both kintex7 and zynq7 and is unstable from one build to another.

litghost:
I've also seen 052 fail, and maybe 053 fail.  It does seem to affect kintex7 and zynq7 more so than artix7.

litghost:
Latest from last run:

zync:
2019-02-20T08:22:23.782547 - 053-pip-ctrlin           - 1h25m: WARNING: Fuzzer failed!
kintex:
2019-02-20T07:00:51.396308 - 052-pip-clkin            - 31m35s: WARNING: Fuzzer failed!
2019-02-20T10:39:33.129469 - 056-pip-rem              - 19m17s: WARNING: Fuzzer failed!
2019-02-20T16:00:09.852344 - 057-pip-bi               - 1m40s: WARNING: Fuzzer failed!
2019-02-20T16:15:12.263067 - 057-pip-bi               - 16m42s: WARNING: Fuzzer failed!
2019-02-20T16:25:30.745754 - 057-pip-bi               - 27m01s: WARNING: Fuzzer failed!
artix:
<no failures>

acomodi:
One possibility is to increase the number of tries, but it will still be non-deterministic and could end up in failure

litghost:
I think it's worth cataloging where the faliures are occuring (e.g. which PIPs).

kgugala:
it happened here https://github.com/SymbiFlow/prjxray/pull/691 for Kintex on fuzzer 057:

```
2019-03-03T20:35:01.646102 - 057-pip-bi               - 2m48s: + make -f /tmpfs/src/github/symbiflow-prjxray-presubmit-db-kintex7/utils/top_generate.mk^M
2019-03-03T20:35:01.646102 - 057-pip-bi               - 2m48s: make[3]: warning: jobserver unavailable: using -j1.  Add '+' to parent make rule.^M
2019-03-03T20:35:01.646102 - 057-pip-bi               - 2m48s: make[3]: *** [vivado.ok] Error 1^M
2019-03-03T20:35:01.646102 - 057-pip-bi               - 2m48s: make[2]: *** [build/1/specimen_005/OK] Error 2^M
2019-03-03T20:35:01.646102 - 057-pip-bi               - 2m48s: make[2]: *** Waiting for unfinished jobs....^M
[...]

PR #672, opened 2019-02-20
Save fuzzer logs, even on success.

litghost:
Previous CI logic deleted logs files along with intermediate outputs
from fuzzers.  This makes debugging issues like #661 harder.

### DIFF
--- a/fuzzers/Makefile
+++ b/fuzzers/Makefile
@@ -11,6 +11,7 @@ FUZZONLY=N
 BITONLY=N
 
 all:
+clean: clean_fuzzers clean_logs
 
 define fuzzer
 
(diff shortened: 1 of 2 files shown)


fuzzers/Makefile
```diff
@@ -21,8 +22,10 @@ define fuzzer
 all: $(1)/run.ok
 
 # Make the clean target run `make clean` in the fuzzer's directory.
-clean::
+clean_fuzzers::
 	$$(MAKE) -C $(1) clean
+
+clean_logs::
 	rm -rf $(1)/logs
```
mithro: Can you make a `make clean` which calls both `clean_fuzzers` and `clean_logs`.

mithro: Opps! Looks like this is already done on line 14....

mithro:
Looks like the issue on CI was around keeping non-log files.

Just FYI - The artifact collection regexs are at https://github.com/SymbiFlow/prjxray/blob/52fbb756656dc0a7907263f882d53f1de59a9222/.github/kokoro/kokoro-cfg.py#L10-L25

PR #686, opened 2019-03-01
Use seperate build directory for each piplist.

litghost:
Vivado expects to be executed in a unique directory in batch mode.
Previous Makefile had all piplist's running the same directory,
resulting in a race condition.


litghost:
This helps with a flake that has been occurring with the 060 fuzzer.

mithro:
Needs a `make format`

litghost:
> Needs a `make format`

Done

mithro:
LGTM - Looks like 056-pip-rem is failing on CI?

PR #693, opened 2019-03-04
fuzzers: int_loop: exit on empty todo

### LINKED ISSUE #692: Fuzzers fail on generating DB when empty todo.txt was generated
When all the INTs were already solved the generated todo.txt is empty. Fuzzers still attempt to push the db ending in mergedb error.

kgugala:
fixes #692

### DIFF
--- a/fuzzers/int_loop.sh
+++ b/fuzzers/int_loop.sh
@@ -43,6 +43,10 @@ i=1
 while true; do
     ${MAKE} ITER=$i cleaniter
     ${MAKE} ITER=$i build/todo.txt
+    if [ ! -s build/todo.txt -a $i -eq 1 ]; then
+        echo "Empty TODO file, assuming all the ints were already solved!"
+        exit 0
+    fi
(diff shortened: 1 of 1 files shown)


fuzzers/int_loop.sh
```diff
@@ -43,6 +43,10 @@ i=1
 while true; do
     ${MAKE} ITER=$i cleaniter
     ${MAKE} ITER=$i build/todo.txt
+    if [ ! -s build/todo.txt ]; then
```
litghost: I don't think this helps, because the problem is that pushdb fails because there is nothing to push.

kgugala: Actually it helps in a case where empty todo was generated in the first iteration. Currently if such a situation happens the code breaks the loop and goes directly to pushdb and fails on empty segbits.

I'll extend the if empty todo with first iteration check.

kgugala: done, now the check is performed only for the first loop iteration

PR #699, opened 2019-03-06
fuzzers: 007: fix linprog callback to fit scipy 1.2.1

### LINKED ISSUE #698: fuzzer 007 fails with keyerror
```
fuzzers/007-timing/solve_linprog.py", line 103, in callback
    iters[0] = kwargs['nit']
KeyError: 'nit'
```

kgugala:
fixes #698

### DIFF
--- a/fuzzers/007-timing/solve_linprog.py
+++ b/fuzzers/007-timing/solve_linprog.py
@@ -100,9 +100,9 @@ def run_corner(
     printn = [0]
 
     def callback(xk, **kwargs):
-        iters[0] = kwargs['nit']
+        iters[0] = xk['nit']
         if time.time() - tlast[0] > 1.0:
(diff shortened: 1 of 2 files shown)


kgugala:
requirements.txt file already includes the scipy and downloads the 1.2.1 version, which does not work with current 007

litghost:
> requirements.txt file already includes the scipy and downloads the 1.2.1 version, which does not work with current 007

Sure, but scipy 1.2.1 was only released last month and scipy 1.2 (which introduced the break) was released Dec 17, 2018.  So to prevent future breakages, lets lock scipy to a version, but scipy has been demonstrated to introduce breaking changes in minor releases.

kgugala:
make sense. I added scipy version to the requirements file

Issue #694, opened 2019-03-04
Implement xc7frames2bit


litghost:
Currently creating new bitstreams is limited to using xc7patch to replace frames in an existing bitstream.  Once enough bits in the bitstream are defined, xc7patch can be skipped, and a bitstream can be defined directly from the frames file.  This is required to break out of the ROI.

PR #705, opened 2019-03-08
Implement the xc7frames2bit tool

tmichalak:

The tool is a WIP, because there is some additional testing needed as the tool produces a bitstream that when converted to fasm contains a lot of non-converted bits. This is still under investigation.
Currently I used the following script and compared the fasm files.
<pre>
${XRAY_BIT2FASM} design.bit  > design.fasm
${XRAY_FASM2FRAMES} design.fasm design.frames
${XRAY_DIR}/build/tools/xc7frames2bit --part_name $XRAY_PART --part_file $XRAY_PART_YAML --frm_file design.frames --output_file design_frames2bit.bit
${XRAY_BIT2FASM} design_frames2bit.bit > design_frames2bit.fasm
</pre>

### DIFF
--- a/lib/include/prjxray/xilinx/xc7series/frames.h
+++ b/lib/include/prjxray/xilinx/xc7series/frames.h
@@ -12,18 +12,29 @@
 namespace prjxray {
 namespace xilinx {
 namespace xc7series {
+
+// Contains frame information which is used for the generation
+// of the configuration package that is used in bitstream generation.
 class Frames {
        public:
 	typedef std::vector<uint32_t> FrameData;
 	typedef std::map<FrameAddress, FrameData> Frames2Data;
 
+	// Reads the contents of the frames file and populates
+	// the Frames container.
 	int readFrames(const std::string& frm_file_str);
+
+	// Adds empty frames that are present in the tilegrid of a specific part
+	// but are missing in the current frames container.
 	void addMissingFrames(const absl::optional<Part>& part);
+
+	// Returns the map with frame addresses and corresponding data
 	Frames2Data& getFrames();
 
        private:
 	Frames2Data frames_data_;
 
+	// Updates the ECC information in the frame.
 	void updateECC(FrameData& data);
 	uint32_t calculateECC(const FrameData& data);
 };
--- a/lib/include/prjxray/xilinx/xc7series/frames.h
+++ b/lib/include/prjxray/xilinx/xc7series/frames.h
@@ -1,0 +1,35 @@
+#ifndef PRJXRAY_LIB_XILINX_XC7SERIES_FRAMES_H
+#define PRJXRAY_LIB_XILINX_XC7SERIES_FRAMES_H
+
+#include <string>
+#include <vector>
+
+#include <absl/strings/str_split.h>
+#include <prjxray/xilinx/xc7series/configuration.h>
+#include <prjxray/xilinx/xc7series/frame_address.h>
+#include <prjxray/xilinx/xc7series/part.h>
+
+namespace prjxray {
+namespace xilinx {
+namespace xc7series {
+class Frames {
+       public:
+	typedef std::vector<uint32_t> FrameData;
+	typedef std::map<FrameAddress, FrameData> Frames2Data;
+
+	int readFrames(const std::string& frm_file_str);
+	void addMissingFrames(const absl::optional<Part>& part);
+	Frames2Data& getFrames();
+
+       private:
+	Frames2Data frames_data_;
+
+	void updateECC(FrameData& data);
+	uint32_t calculateECC(const FrameData& data);
+};
+
+}  // namespace xc7series
+}  // namespace xilinx
+}  // namespace prjxray
+
+#endif  // PRJXRAY_LIB_XILINX_XC7SERIES_FRAMES_H
--- a/lib/include/prjxray/xilinx/xc7series/utils.h
+++ b/lib/include/prjxray/xilinx/xc7series/utils.h
@@ -1,0 +1,34 @@
+
+#ifndef PRJXRAY_LIB_XILINX_XC7SERIES_UTILS_H_
+#define PRJXRAY_LIB_XILINX_XC7SERIES_UTILS_H_
+
+#include <prjxray/xilinx/xc7series/configuration.h>
+#include <prjxray/xilinx/xc7series/configuration_packet.h>
+#include <prjxray/xilinx/xc7series/frames.h>
+#include <prjxray/xilinx/xc7series/part.h>
+
+namespace prjxray {
+namespace xilinx {
+namespace xc7series {
+using PacketData = std::vector<uint32_t>;
+using BitstreamHeader = std::vector<uint8_t>;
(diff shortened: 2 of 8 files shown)


tmichalak:
This PR is related to issue #694

tools/xc7frames2bit.cc
```diff
@@ -0,0 +1,398 @@
...
+namespace xc7series = prjxray::xilinx::xc7series;
+
+using PacketData = std::vector<uint32_t>;
+using BitHeader = std::vector<uint8_t>;
+using ConfigurationPackets =
+    std::vector<std::unique_ptr<xc7series::ConfigurationPacket>>;
+using FrameAddress2Frames =
+    std::map<xc7series::FrameAddress, std::vector<uint32_t>>;
+
+int readFrames(const std::string& frm_file_str, FrameAddress2Frames& frames) {
```
litghost: Should readFrames be a function shared by xc7patch and xc7frames2bit?

tools/xc7frames2bit.cc
```diff
@@ -0,0 +1,398 @@
...
+	bit_header.push_back(0x0);
+
+	bit_header.insert(bit_header.end(), {'e', 0x0, 0x0, 0x0, 0x0});
+
+	return bit_header;
+}
+
+int writeBitstream(const ConfigurationPackets& packets,
+                   const std::string& output_file) {
+	std::ofstream out_file(output_file);
```
litghost: Needs to be binary mode.

tools/xc7frames2bit.cc
```diff
@@ -0,0 +1,398 @@
...
+			std::cerr << "Failed to read frames" << std::endl
+			          << "Exitting" << std::endl;
+			return 1;
+		}
+	}
+
+	// Put together a configuration package
+	ConfigurationPackets out_packets;
+
+	// Initialization sequence
```
litghost: Can we link to where this sequence came from?

tools/xc7frames2bit.cc
tmichalak: Yes, it could. I would just have to refactor the xc7patch to use it.

tools/xc7frames2bit.cc
tmichalak: Done

tools/xc7frames2bit.cc
tmichalak: The sequence came from https://www.kc8apf.net/2018/05/unpacking-xilinx-7-series-bitstreams-part-2/. It is exactly how it's done in xc7patch.

litghost: If it is identical to how it is done in xc7patch, please make function, and in the function comment link to that page.

tools/xc7frames2bit.cc
```diff
@@ -165,7 +165,7 @@ BitHeader createBitHeader() {
 
 int writeBitstream(const ConfigurationPackets& packets,
                    const std::string& output_file) {
-	std::ofstream out_file(output_file);
+	std::ofstream out_file(output_file, std::ofstream::out);
```
litghost: This is still wrong? See https://en.cppreference.com/w/cpp/io/ios_base/openmode

tmichalak: Oh, yes. You are right, it should be ``std::ofstream::binary``.
It doesn't change anything in the output file though.

tmichalak:
The tool is more or less ready. The fasm files generated by bit2fasm from the xc7frames2bit produced bitstream are identical with the fasm files generated from a vivado bitstream.

litghost:
> The tool is more or less ready. The fasm files generated by bit2fasm from the xc7frames2bit produced bitstream are identical with the fasm files generated from a vivado bitstream.

I'm still awaiting for the common code between xc7patch and xc7frames2bit to be moved to a library.

lib/include/prjxray/xilinx/xc7series/frames.h
```diff
@@ -0,0 +1,35 @@
...
+namespace xilinx {
+namespace xc7series {
+class Frames {
+       public:
+	typedef std::vector<uint32_t> FrameData;
+	typedef std::map<FrameAddress, FrameData> Frames2Data;
+
+	int readFrames(const std::string& frm_file_str);
+	void addMissingFrames(const absl::optional<Part>& part);
+	Frames2Data& operator()();
```
litghost: Please just make this just a named method instead of an operator.

tmichalak: Ok, this is done.

litghost:
Looking good, please add comments to all functions, class and methods describing them.

lib/include/prjxray/xilinx/xc7series/frames.h
```diff
@@ -0,0 +1,38 @@
...
+namespace xc7series {
+
+// Contains frame information which is used for the generation
+// of the configuration package that is used in bitstream generation.
+class Frames {
+       public:
+	typedef std::vector<uint32_t> FrameData;
+	typedef std::map<FrameAddress, FrameData> Frames2Data;
+
+	int readFrames(const std::string& frm_file_str);
```
litghost: Please add method comments.  What does "addMissingFrames" do, and why would I call it?  Etc.

lib/include/prjxray/xilinx/xc7series/utils.h
```diff
@@ -0,0 +1,34 @@
...
+#include <prjxray/xilinx/xc7series/part.h>
+
+namespace prjxray {
+namespace xilinx {
+namespace xc7series {
+using PacketData = std::vector<uint32_t>;
+using BitstreamHeader = std::vector<uint8_t>;
+using ConfigurationPackage = std::vector<std::unique_ptr<ConfigurationPacket>>;
+
+PacketData createType2ConfigurationPacketData(const Frames::Frames2Data& frames,
```
litghost: Comments?

lib/xilinx/xc7series/frames.cc
```diff
@@ -0,0 +1,106 @@
...
+namespace xilinx {
+namespace xc7series {
+
+Frames::Frames2Data& Frames::getFrames() {
+	return frames_data_;
+}
+
+// Reads the contents of the frames file and populates
+// the Frames container.
+int Frames::readFrames(const std::string& frm_file_str) {
```
litghost: Method level comments should be at the interface (e.g. header file) rather than implementation.  If a comment is required to understand the function/method, they should be in the body of the function/method.

lib/include/prjxray/xilinx/xc7series/frames.h
tmichalak: I actually added the comments to the cc files

lib/include/prjxray/xilinx/xc7series/utils.h
tmichalak: the same here, the comments are in the .cc file

lib/xilinx/xc7series/frames.cc
tmichalak: Ok, I will move all the comments to the header files.

PR #707, opened 2019-03-11
infra: Fix the diff output.

mithro:


litghost:
Please also modify the continuous configs to grab the diff.patch.

PR #724, opened 2019-03-14
roi_harness: Add ARTY-A7-UART configuration

acomodi:
This PR is to add `arty` support with its ROI harness. Currently UART pins, one LED and a button are correctly enabled in the harness.

Signed-off-by: Lukasz Dalek <ldalek@antmicro.com>


litghost:
Why not 4 LEDs, 4 buttons, and the UART?  This would make it equiv to the basys3 harness?

litghost:
> > Why not 4 LEDs, 4 buttons, and the UART? This would make it equiv to the basys3 harness?
> 
> Unfortunately there are issues if using all the LEDS and Buttons. This happens also for the zynq (some buttons, even if routed using the `roi_harness` minitest, do not properly work on HW)

Sounds like a problem in the harness definition?  You should adjust which routing stubs are used and see if you can resolve the issue.

acomodi:
Yes indeed, the issue is for sure in the `harness` definition. I believe that the `roi_harness` minitest has to be different for each board that is used and manual routing has to be adopted to get the required harnesses. I think that, given that the roi_harnesses will not be needed when we get IOBs and CLK, even if we do not have all the `LEDs` and `buttons` working, this PR could be considered valid right now just to have an example for the arty board.

PR #727, opened 2019-03-16
Solve remaining bits in the ROI

litghost:
This PR add fuzzers for the BUFG and HCLK_CMT interconnect, and completes the interconnect fuzzing of the CLK_HROW switch box.

With these fuzzers, all remaining bits in the basys3 ROI are solved, fixes #684.  This PR superceeds #717, but the comments from #717 apply here.

PR #736, opened 2019-03-21
add bits outside ROI to required features

tmichalak:


litghost:
I assume if you run this with the current DB, the basys3 harness fails?  If so, we need to wait for https://github.com/SymbiFlow/prjxray/pull/727 to be stabilized and merged, and for a new prjxray-db.

tmichalak:
Yes, the harness fails with the current DB. Waiting for #727.

PR #737, opened 2019-03-21
Re-enable the DSP fuzzer.

mithro:


mithro:
Reenabling the fuzzer seems to work fine.

PR #738, opened 2019-03-21
Add another IOI variant to the ppips

litghost:


acomodi:
Maybe it is worth to add also the BRKH_INT at this point, as I suppose those are ppips as well.

litghost:
> Maybe it is worth to add also the BRKH_INT at this point, as I suppose those are ppips as well.

Agreed.

acomodi:
Make format

PR #740, opened 2019-03-22
Update info file on CI

mithro:
Makes the patch produce by the CI system less confusing.


litghost:
This broke the CI:
```
Checking artix7 database
============================
python3: can't open file 'utils/db-check.py': [Errno 2] No such file or directory
Makefile:103: recipe for target 'db-check-artix7' failed
make: *** [db-check-artix7] Error 2
```

mithro:
@litghost Gah, that doesn't appear in the run-results top level output, I saw that just the pip runs were flaky, so merged it.

PR #706, opened 2019-03-08
fuzzers: Adding BEL timing fuzzer

kgugala:
This PR adds bel timing fuzzer. For now only Carry, LUT6 and FDRE BELs are handled. More to come

PR #755, opened 2019-04-04
utils: add sdfparse tool

### LINKED ISSUE #753: Implement python sdf parser
We need this to parse timing files and include delay info in arch.xml

kgugala:
Initial verision of the sdf parser. It can parse sdfs produced by bel timing fuzzer from #706

example result:
sdf input: https://pastebin.com/mtUN2syw
json output: https://pastebin.com/N04NRq85

this PR resolves #753

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -13,3 +13,4 @@ simplejson
 sympy
 textx
 yapf==0.24.0
+ply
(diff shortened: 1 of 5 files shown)


kgugala:
sure, this way it will be easier to use this code

mithro:
I imported this pull request into the repo at https://github.com/SymbiFlow/python-sdf-timing

I chose the name sdf-timing because there also seems to be a Scientific Data Format (an implementation of the HDF5 standard) which has existing Python parsers.

PR #764, opened 2019-04-06
IOB frame count is 40 not 42.

litghost:
Because IOB's live at the top of CLB_IO_CLK frame, the maximum on artix7
should be a base address of 0x004015A7, which from a base address of
0x00401580 is 40 frames.


litghost:
Looks like this PR is wrong, there are definitely 42 frames https://github.com/SymbiFlow/prjxray-db/blob/master/artix7/xc7a35tcpg236-1.yaml

Issue #768, opened 2019-04-06
BRAM patching has an offset of 2


litghost:
Using xc7patch to write to frame address 0x00820000 to will result in data getting written to 0x00820002.  I haven't pinned down if this is a regression, or it has always been like this.  @tmichalak you were in here last, any ideas?

litghost:
This error appears in 2aa3c309762a1ac2e3c5989e6419f4825a24ac0e, before the https://github.com/SymbiFlow/prjxray/pull/705 .  This might be a long standing bug.

litghost:
Replication steps:
1. ``build/tools/xc7patch --part_file prjxray-db/artix7/xc7a35tcpg236-1.yaml  --bitstream_file prjxray-db/artix7/harness/basys3/swbut/design.bit --output_file test.bit --frm_file top.frames``

Where top.frames is just:
```
0x00820000 0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x11190000,0x07DF8AB3,0x04351098,0x09EE160E,0x09BE05A5,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000,0x00000000
```

[...]

PR #769, opened 2019-04-08
xc7frames2bit: add missing frames starting with 0

tmichalak:
Issue #768 showed that there was a bug when filling in the configuration with empty frames. Instead of starting with frame 0 it started with the first frame in the frames file. If the frames file already contained frame with address 0x00000000 (which was usually the case) everything worked fine, but with a custom frames file that was missing frame 0 we ended up with incorrect frame addresses when comparing the results of bitread.
Note however, that this is not related to the bug in #768.


litghost:
Please add some tests

PR #770, opened 2019-04-08
xc7series: fix condition detecting 2 empty frames between rows

tmichalak:
This PR is to solve #768 
xc7 bitstreams appear to have 2 frames of padding between rows.
This is taken into account in the Type2 configuration package creation: https://github.com/SymbiFlow/prjxray/blob/master/lib/xilinx/xc7series/utils.cc#L35

However, the condition in ``configuration.h`` that is used by bitread doesn't check the block_type nad top/bottom half conditions:
https://github.com/SymbiFlow/prjxray/blob/master/lib/include/prjxray/xilinx/xc7series/configuration.h#L152

This PR makes sure the conditions in both places are aligned.

### DIFF
--- a/lib/CMakeLists.txt
+++ b/lib/CMakeLists.txt
@@ -16,6 +16,7 @@ add_library(libprjxray
 	xilinx/xc7series/row.cc
 	xilinx/xc7series/frames.cc
 	xilinx/xc7series/utils.cc
+	xilinx/xc7series/ecc.cc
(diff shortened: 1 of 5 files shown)


litghost:
Please add some tests.

litghost:
Output of artix database looks the same, so this doesn't cause a regression on the database.   Add some tests and LGTM.

PR #771, opened 2019-04-08
Add fuzzer for documenting pin to pad relationship for part.

litghost:

### DIFF
--- a/fuzzers/075-pins/Makefile
+++ b/fuzzers/075-pins/Makefile
@@ -1,0 +1,23 @@
+N := 1
+SPECIMENS := $(addprefix build/specimen_,$(shell seq -f '%03.0f' $(N)))
+SPECIMENS_OK := $(addsuffix /OK,$(SPECIMENS))
+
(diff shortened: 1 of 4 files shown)


mithro:
What does the output look like?

litghost:
```
pin,site,tile
M9,IPAD_X0Y1,MONITOR_BOT_X46Y79
L10,IPAD_X0Y0,MONITOR_BOT_X46Y79
P20,IOB_X0Y49,LIOB33_SING_X0Y49
P22,IOB_X0Y48,LIOB33_X0Y47
```

mithro:
The only other CSV file we have in the database is the `elements_count.csv`. Should we be using JSON here instead?

litghost:
> The only other CSV file we have in the database is the `elements_count.csv`. Should we be using JSON here instead?

JSON seems unnecessary here, and complicates the fuzzer output.  There isn't structure, so a simple flat CSV seems fine.

PR #821, opened 2019-05-06
Add objects and docstrings for timing model.

litghost:
This is the beginning of the prjxray representation of the 7-series timing model.  This will form the basis for the 7-series route timing import.  Another PR will add a fuzzer to add the input values to the model.

I'm putting this PR up for review of the object model.

### DIFF
--- a/prjxray/tile.py
+++ b/prjxray/tile.py
@@ -8,7 +8,7 @@ TileDbs = namedtuple(
     'TileDbs', 'segbits block_ram_segbits ppips mask tile_type')
 
 
-class OutpinTiming(namedtuple('OutpinTiming', 'delays drive_resistance')):
+class OutPinTiming(namedtuple('OutPinTiming', 'delays drive_resistance')):
     """ Timing for site output pins.
 
     Attributes
@@ -22,7 +22,7 @@ class OutpinTiming(namedtuple('OutpinTiming', 'delays drive_resistance')):
     pass
 
 
-class InpinTiming(namedtuple('InpinTiming', 'delays capacitance')):
+class InPinTiming(namedtuple('InPinTiming', 'delays capacitance')):
     """ Timing for site input pins.
 
     Attributes
@@ -125,7 +125,7 @@ class SitePin(namedtuple('SitePin', 'name wire timing')):
         all sites of the same type.
     wire : str
         Wire name within the tile.  This name is site instance specific.
-    timing : Either InpinTiming or OutpinTiming
+    timing : Either InPinTiming or OutPinTiming
         Timing of site pin. May be None if database lacks timing information.
 
     """
@@ -177,13 +177,13 @@ def get_pip_timing(pip_timing_json):
 
 
 def get_site_pin_timing(site_pin_info):
-    """ Convert site_pin_info JSON into InpinTiming or OutpinTiming object.
+    """ Convert site_pin_info JSON into InPinTiming or OutPinTiming object.
 
     Returns
     -------
     If timing information is not present for this site pin, returns None.
-    If this is an output pin, returns OutpinTiming.
-    If this is an input pin, returns InpinTiming.
+    If this is an output pin, returns OutPinTiming.
+    If this is an input pin, returns InPinTiming.
 
     """
     if isinstance(site_pin_info, str):
@@ -198,13 +198,13 @@ def get_site_pin_timing(site_pin_info):
 
     if 'cap' in site_pin_info:
         assert 'res' not in site_pin_info
-        return wire, InpinTiming(
+        return wire, InPinTiming(
             delays=delays,
             capacitance=float(site_pin_info['cap']) / CAPACITANCE_FACTOR,
         )
     else:
         assert 'res' in site_pin_info
-        return wire, OutpinTiming(
+        return wire, OutPinTiming(
             delays=delays,
             drive_resistance=float(site_pin_info['res']) / RESISTANCE_FACTOR,
         )
--- a/prjxray/tile.py
+++ b/prjxray/tile.py
@@ -56,7 +56,7 @@ class PipTiming(namedtuple('PipTiming',
 class Pip(namedtuple(
         'Pip',
     ('name', 'net_to', 'net_from', 'can_invert', 'is_directional', 'is_pseudo',
-     'is_pass_transitor', 'timing', 'backward_timing'))):
+     'is_pass_transistor', 'timing', 'backward_timing'))):
     """ Pip information.
 
     Attributes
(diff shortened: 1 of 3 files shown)


prjxray/timing.py
```diff
@@ -0,0 +1,104 @@
...
+
+Inpins have instrinsic delays and a sink capacitance.
+
+The elmore delay is the RC tree formed by these 5 components.  Out pins and
+buffer switches are the roots of the elmore tree. Buffer switches
+and inpins are leafs of the elmore tree. Wires and pass-through switches are
+nodes in the tree, and share their capacitance upstream and downstream using a
+pi-model.
+
+"""
```
mithro: It might be worth using [`type hints`](https://docs.python.org/3/library/typing.html)? See example
https://github.com/SymbiFlow/symbiflow-arch-defs/blob/0c856ca4c89645b7d1866cdf77ce36a9366eddde/utils/lib/pb_type.py#L4-L32

prjxray/tile.py
```diff
@@ -1,38 +1,251 @@
...
-    'Pip', 'name net_to net_from can_invert is_directional is_pseudo')
-""" Site - Represents an instance of a site within a tile.
 
-name - Name of site within tile, instance specific.
-prefix - Prefix of site naming in Xilinx parlance.
-type - What type of slice this instance presents.
-pins - Instaces of site pins within this site and tile.  This is an tuple of
-       SitePin tuples, and is specific to this instance of the site within
-       the tile.
+class OutpinTiming(namedtuple('OutpinTiming', 'delays drive_resistance')):
```
mithro: Should these be `OutPinTiming` / `InPinTiming` ?

prjxray/timing.py
```diff
@@ -0,0 +1,615 @@
...
+
+The elmore delay is the RC tree formed by these 5 components.  Out pins and
+buffer switches are the roots of the elmore tree. Buffer switches
+and inpins are leafs of the elmore tree. Wires and pass-transistor switches are
+nodes in the tree. Wires share their capacitance upstream and downstream using
+a pi-model.
+
+Example timing tree:
+
++------+
```
mithro: Please indent the timing tree.

prjxray/timing.py
```diff
@@ -0,0 +1,615 @@
...
+        ),
+    }
+
+
+# Math models are used to represent abstract operations for the timing models.
+# This is useful for creating excel workbooks that can update dynamically, or
+# generating a model with symbolic constants to be backsolved.
+
+
+class ExcelMathModel(object):
```
mithro: I feel like this might be better in it's own file?

prjxray/timing.py
```diff
@@ -0,0 +1,615 @@
...
+        Must call propigated_delays on all children of this node, and add this
+        node to elements.
+
+        Arguments
+        ---------
+        elements : list of TimingNode's
+            List of delay nodes between root of this tree and this node.
+        math : MathModel
+            Math model to use to compute delays
+        """
```
mithro:
```suggestion
        """
        pass
```

prjxray/timing.py
```diff
@@ -0,0 +1,615 @@
...
+class Inpin(DownstreamNode):
+    """ Represents a site input pin.
+
+    Represents leaf of timing model.  Once model is connected and delays
+    are propigate (by calling Outpin,propigated_delays), get_delays will
+    correctly return the list of delay elements from the root to this leaf.
+
+    Arguments
+    ---------
+
```
mithro: Types?

litghost: The types here are intentionally omitted.  The types vary based on the math model in used.

mithro:
Maybe consider using [Python 3 Dataclasses](https://docs.python.org/3/library/dataclasses.html)? (Not 100% sure it makes sense here?) I only found out about them recently and they reduce a lot of boilerplate stuff...

litghost:
> Maybe consider using [Python 3 Dataclasses](https://docs.python.org/3/library/dataclasses.html)? (Not 100% sure it makes sense here?) I only found out about them recently and they reduce a lot of boilerplate stuff...

I'm having a hard time seeing how Dataclasses apply here...

mithro:
> > Maybe consider using [Python 3 Dataclasses](https://docs.python.org/3/library/dataclasses.html)? (Not 100% sure it makes sense here?) I only found out about them recently and they reduce a lot of boilerplate stuff...
> 
> I'm having a hard time seeing how Dataclasses apply here...

They are generally useful in place of a `namedtuple` -- especially if you don't want them to be immutable. 



```python

class OutpinTiming(namedtuple('OutpinTiming', 'delays drive_resistance')):
    """ Timing for site output pins.
    Attributes
    ----------
    delays : dicts of PvtCorner to IntristicDelay
        Intristic delay of output pin.
    drive_resistance : float
        Resistance of drive output pin (milliOhms).
    """
    pass
```

verses

```python
from typing import Dict
from dataclasses import dataclass, field

@dataclass
class OutpinTiming:
    delays: Dict[PvtCorner, IntristicDelay] = field(default_factory=dict)
    drive_resistance: float
```

([The attrs library](http://www.attrs.org/en/stable/) is another option
compared to dataclasses).

Frustratingly I can't see a way to have the docstring as part of the attributes
on the dataclass.

litghost:
> > > Maybe consider using [Python 3 Dataclasses](https://docs.python.org/3/library/dataclasses.html)? (Not 100% sure it makes sense here?) I only found out about them recently and they reduce a lot of boilerplate stuff...
> > 
> > 
> > I'm having a hard time seeing how Dataclasses apply here...
> 
> They are generally useful in place of a `namedtuple` -- especially if you don't want them to be immutable.
> 
> ```python
> class OutpinTiming(namedtuple('OutpinTiming', 'delays drive_resistance')):
>     """ Timing for site output pins.
>     Attributes
>     ----------
>     delays : dicts of PvtCorner to IntristicDelay
>         Intristic delay of output pin.
>     drive_resistance : float
>         Resistance of drive output pin (milliOhms).
>     """
>     pass
> ```
> 
> verses
> 
> ```python
> from typing import Dict
> from dataclasses import dataclass, field
> 
> @dataclass
> class OutpinTiming:
>     delays: Dict[PvtCorner, IntristicDelay] = field(default_factory=dict)
>     drive_resistance: float
> ```
> 
> ([The attrs library](http://www.attrs.org/en/stable/) is another option
> compared to dataclasses).
> 
> Frustratingly I can't see a way to have the docstring as part of the attributes
> on the dataclass.

But namedtuples are exactly what we want, simple immutable data stores.

prjxray/tile.py
litghost: Fixed

prjxray/timing.py
litghost: Done

prjxray/timing.py
litghost: Okay

Issue #820, opened 2019-05-06
030-iob fuzzer doesn't solve several bits for zynq7


tmichalak:
Issue #746 is keeping https://github.com/SymbiFlow/symbiflow-arch-defs/pull/500 from being merged.
The analysis done in #746 showed that the missing bits are present in the database for Artix7.
Some of the bits are calculated in the 030-iob fuzzer. It has to be fixed in order to solve the following missing bits:
<pre>
RIOB33.IOB_Y0.IN_ONLY 38_118 39_119 39_125
RIOB33.IOB_Y0.LVCMOS25_LVCMOS33_LVTTL.IN 38_86 39_85 39_87 39_113
</pre>

tmichalak:
The problem can be related to the fact that we don't have IOBs on the left side. Nevertheless for the Y0 site (type IOB33M) I have always 0 candidates for IN_OUT_COMMON for all iostandards and drive strengths.
This leads to no IN_ONLY feature.

Issue #746, opened 2019-03-27
unknown bit in zynq7 harness


mithro:
Using database https://github.com/mithro/prjxray-db/commit/88d9a13ebdf431ec9bd94377817daf4d46f7a938
```
Writing bitstream ./design.bit...
INFO: [Vivado 12-1842] Bitgen Completed Successfully.
INFO: [Project 1-118] WebTalk data collection is enabled (User setting is ON. Install Setting is ON.).
INFO: [Common 17-186] '/usr/local/google/home/tansell/github/SymbiFlow/prjxray/minitests/roi_harness/build/design/usage_statistics_webtalk.xml' has been successfully sent to Xilinx on Tue Mar 26 18:26:29 2019. For additional details about this file, please refer to the WebTalk help file at /opt/Xilinx/Vivado/2017.2/doc/webtalk_introduction.html.
9 Infos, 5 Warnings, 0 Critical Warnings and 0 Errors encountered.
write_bitstream completed successfully
write_bitstream: Time (s): cpu = 00:00:07 ; elapsed = 00:00:06 . Memory (MB): peak = 2350.082 ; gain = 332.453 ; free physical = 17416 ; free virtual = 39424
INFO: [Common 17-206] Exiting Vivado at Tue Mar 26 18:26:29 2019...
++ fgrep CRITICAL vivado.log
+ test -z ''
+ /usr/local/google/home/tansell/github/SymbiFlow/prjxray/build/tools/bitread --part_file /usr/local/google/home/tansell/github/SymbiFlow/prjxray/database/zynq7/xc7z010clg400-1.yaml -F 0x00000000:0xffffffff -o design.bits -z -y design.bit
Bitstream size: 2083839 bytes
Config size: 520922 words
Number of configuration frames: 5144
DONE
+ python3 /usr/local/google/home/tansell/github/SymbiFlow/prjxray/utils/bit2fasm.py --verbose design.bit
+ python3 /usr/local/google/home/tansell/github/SymbiFlow/prjxray/utils/fasm2frames.py design.fasm design.frm
[...]

PR #823, opened 2019-05-07
Resolve missing IOB bits for Zynq

tmichalak:
This PR solves issue #820.
In #746 we have several bits missing, some of them belong to the IOB tiles.
Those bits were resolved correctly for Artix, hence it had to be analyzed why this doesn't happen for Zynq.
Turned out that alike Artix there seems to be a broken tile that has the bits which are being solved always on.
After this fix the IOB related bits are no longer unknown in the roi harness for zynq.
Also in this PR a small change in the pushdb target was introduced to avoid creating a segbits_liob.db in the database since there are no IOB on the left hand side in Zynq7.

### DIFF
--- a/fuzzers/030-iob/Makefile
+++ b/fuzzers/030-iob/Makefile
@@ -12,9 +12,11 @@ build/segbits_xiob33.db: build/segbits_xiob33.rdb process_rdb.py bits.dbf
 	${XRAY_MASKMERGE} build/mask_xiob33.db $$(find -name segdata_liob33.txt) $$(find -name segdata_riob33.txt)
 
 pushdb:
+ifneq (${XRAY_DATABASE}, zynq7)
 	${XRAY_MERGEDB} liob33 build/segbits_xiob33.db
-	${XRAY_MERGEDB} riob33 build/segbits_xiob33.db
 	${XRAY_MERGEDB} mask_liob33 build/mask_xiob33.db
+endif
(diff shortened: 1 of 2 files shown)


fuzzers/030-iob/Makefile
```diff
@@ -12,9 +12,11 @@ build/segbits_xiob33.db: build/segbits_xiob33.rdb process_rdb.py bits.dbf
 	${XRAY_MASKMERGE} build/mask_xiob33.db $$(find -name segdata_liob33.txt) $$(find -name segdata_riob33.txt)
 
 pushdb:
+ifneq (${XRAY_DATABASE}, zynq7)
```
litghost: Just remove the ifneq? You added them as part of making things green, but with the change to generate.py, things should be okay with them?

tmichalak: Because for Zynq we only have RIOBs? There is the PS on the left instead of LIOBs.

kgugala: > Because for Zynq we only have RIOBs? There is the PS on the left instead of LIOBs.
This is true only for smaller Zynq's. Bigger ones (kintex based) have RIOB's, but we don't support those now.

tmichalak: Agreed. However currently the fuzzer assumes that the bits for liob are identical with the ones for riob which will probably not be true for Zynqs. This will have to be adressed once we start supporting other Zynqs but currently it is not true so why add this to the database.

PR #831, opened 2019-05-14
057-pip-bi: Increase number of tries to find a suitable PIP INT tile

tmichalak:
This PR is aimed at solving the stability issue of the 057-pip-bi fuzzer reported in issue #661.
Currently the number of tries was set to 3. In this case to find the right tile needed to successfully route through the specified BI PIPs the number of attempts needs to be higher (sometimes 4-5).
Added 10 to be on the safe side. This doesn't affect run time much anyway.


tmichalak:
The CI failed on 053-pip-ctrlin which is addressed in #849

tmichalak:
I know that CI is red, but this is because of the instability of 053 which is subject to #849.
I tested this fuzzer locally for more than 50 iterations and it worked fine after the fix.
Before it failed after 2-3 iterations.
If we merged this it would leave us with only 41 being disabled.

PR #835, opened 2019-05-14
BEL timings fixes

kgugala:
fixes #832 
fixes #833 
fixes #834

PR #836, opened 2019-05-14
utils: makesdf: sort the output

kgugala:
Signed-off-by: Karol Gugala <kgugala@antmicro.com>

### DIFF
--- a/utils/makesdf.py
+++ b/utils/makesdf.py
@@ -24,8 +24,8 @@ def produce_sdf(timings, outdir):
     (SDFVERSION \"3.0\")
     (TIMESCALE 1ps)
 """
-        for site in timings[slice]:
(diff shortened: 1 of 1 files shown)


mithro:
Poke me when this finishes and I'll see about getting the change into the prjxray-db output.

litghost:
> Poke me when this finishes and I'll see about getting the change into the prjxray-db output.

Poke

mithro:
Should be live now - https://github.com/SymbiFlow/prjxray-db/commit/34ea6eb08a63d21ec16264ad37a0a7b142ff6031

mithro:
I rebased https://github.com/SymbiFlow/prjxray/pull/835 on top of this CL and pushed it so that we can get the updated output.

Issue #784, opened 2019-04-16
Timeout on 041-clk-hrow-pips fuzzer


tmichalak:
The 041-clk-hrow-pips tends to timeout most of the times.
The objective is to get to the bottom of the timeout, but it would be beneficial in general for the run time of the entire database generation to analyze the root cause and shorten the run time of this fuzzer.

tmichalak:
The fuzzer times out because after few iterations it ends up with a todo list that it's unable to solve no matter how many further iterations will be performed:
One example of such list:
<pre>
CLK_HROW_BOT_R.CLK_HROW_BOT_R_CK_BUFG_CASCO21.CLK_HROW_CK_IN_R8
CLK_HROW_BOT_R.CLK_HROW_BOT_R_CK_BUFG_CASCO21.CLK_HROW_CK_IN_R9
CLK_HROW_BOT_R.CLK_HROW_BOT_R_CK_BUFG_CASCO27.CLK_HROW_CK_IN_L9
CLK_HROW_BOT_R.CLK_HROW_BOT_R_CK_BUFG_CASCO2.CLK_HROW_CK_IN_L8
CLK_HROW_BOT_R.CLK_HROW_BOT_R_CK_BUFG_CASCO2.CLK_HROW_CK_IN_L9
CLK_HROW_BOT_R.CLK_HROW_BOT_R_CK_BUFG_CASCO3.CLK_HROW_CK_IN_R9
</pre>

litghost:
Ya, the R -> CASCO were very annoying to get right.  I'd try tuning the TODO routing portion and see if it is missing diversity.  I believe you mentioned that these bits were getting dropped due to bit conflicts?

tmichalak:
The bits we had bit conflicts with were: 
<pre>
CLK_HROW_TOP_R.CLK_HROW_TOP_R.CLK_HROW_CK_INT_0_1_ACTIVE
</pre>
Not sure if that is related.

Could you be more specific about what kind of diversity is a must have?

litghost:
Didn't we say that ``CLK_HROW_TOP_R.CLK_HROW_TOP_R.CLK_HROW_CK_INT_0_1_ACTIVE`` were fake bits?  So how is there a conflict?

tmichalak:
Yes, once we assumed they are fake we excluded them. They were conflicting before we had done this.
I was simply referring to what you wrote earlier:
>  I believe you mentioned that these bits were getting dropped due to bit conflicts?

[...]

PR #837, opened 2019-05-15
Add target to cleanup all piplists

tmichalak:
It is possible for a PIP fuzzer to fail after reaching the defined iteration timeout due an incompatible piplist.
This happens when running all the fuzzers for one architecture and then running them for another architecture after just calling ``make clean`` in the ``fuzzers`` directory. The clean target doesn't clean the piplists which can cause problems during the execution of some pip fuzzers.
Namely, the ``todo`` file used in the pip fuzzer, which is generated based on the specific piplist generated for one architecture may not be compatible for another architecture which might not have the same pips.
The symptom is usually that the fuzzer which is set to finish within max 50 iterations, advances to a certain iteration, but later on performs all remaining iterations until 50 without being able to solve any of the remaining PIPs on the todo list, simply because the new architecture doesn't have those PIPs.

This PR adds an explicit clean_piplists target to the Makefile in the ``fuzzers`` directory and also uses that in kokoro's run script.
A separate target was used instead of adding the cleanup to the clean_fuzzers target to avoid forcing the piplist generation for each iteration that is used in pip_loop.mk.

### DIFF
--- a/fuzzers/050-pip-seed/Makefile
+++ b/fuzzers/050-pip-seed/Makefile
@@ -3,7 +3,7 @@ ifeq ($(QUICK),Y)
 N ?= 1
 SEGMATCH_FLAGS=
 else
-N ?= 48
+N ?= 60
 # Do relatively large batch to keep parallelism high
 # LCM between 12 (CPUs on my system) and 16, a common CPU count
 SEGMATCH_FLAGS=-m 15 -M 45
--- a/.github/kokoro/db-full.sh
+++ b/.github/kokoro/db-full.sh
@@ -176,5 +176,6 @@ echo "----------------------------------------"
 	echo
 	echo "Cleaning up so CI doesn't save all the excess data."
 	make clean_fuzzers
+	make clean_piplists
 )
 echo "----------------------------------------"
--- a/fuzzers/Makefile
+++ b/fuzzers/Makefile
@@ -11,10 +11,13 @@ FUZZONLY=N
 BITONLY=N
 
 all:
-clean: clean_fuzzers clean_logs clean_locks
+clean: clean_fuzzers clean_piplists clean_logs clean_locks
 
 clean_locks:
 	rm -rf /tmp/segbits_*.db.lock
+
+clean_piplists:
+	rm -rf $(XRAY_FUZZERS_DIR)/piplist/build
 
 define fuzzer
 


kgugala:
LGTM

mithro:
While I think this is a good CL I don't think it can be a cause of the failures? Each Kokoro builder is generating a fresh database for exactly one device type?

tmichalak:
@mithro: you are right, this PR doesn't fix the instability.
It simply adds a CL which is very important when running the pip fuzzers locally and switching between architectures. I described above the consequences of not cleaning the piplists when running the 041 fuzzer for another architecture.

mithro:
Please update the description and then I can merge.

tmichalak:
@mithro The description has been updated.

mithro:
@tmichalak I don't see it updated? It still mention that is solving 041-clk-hrow-pips fuzzer reported in #784?

tmichalak:
@mithro: I guess now the description is fine.

tmichalak:
Changed to WIP as it's constantly red.
Despite he fact that it's RED due to the the prjxray stability issues, we can't merge it until it's green.

fuzzers/050-pip-seed/Makefile
```diff
@@ -3,6 +3,9 @@ ifeq ($(QUICK),Y)
 N ?= 1
 SEGMATCH_FLAGS=
 else
+ifeq (${XRAY_DATABASE}, zynq7)
+N=60
```
litghost: Should be "N ?= 60" to accommodate "QUICK".

tmichalak: Sure, I will fix that right away.

tmichalak:
> Same change to N variable, and LGTM

Which N variable do you have in mind here?

litghost:
Why is this PR WIP?

fuzzers/050-pip-seed/Makefile
litghost: Still not fixed?

tmichalak:
@litghost I removed the WIP and pushed (forgot to push the last time) the change related to one of you review comments. 
There is however one more comment of yours:

>Same change to N variable, and LGTM

Which N variable do you have in mind here?

litghost:
> > Same change to N variable, and LGTM
> 
> Which N variable do you have in mind here?

https://github.com/SymbiFlow/prjxray/pull/837#discussion_r286535301

tmichalak:
@litghost Hmm, okay, so you only meant one change. I fixed that particular line just few minutes ago. Not sure why kokoro is stuck?

litghost:
> @litghost Hmm, okay, so you only meant one change. I fixed that particular line just few minutes ago. Not sure why kokoro is stuck?

Likely a problem on the github side, leave it for a bit.

litghost:
Ya https://www.githubstatus.com/

PR #838, opened 2019-05-16
041-clk-hrow-pips: Fix timeout and bit collision problems

tmichalak:
The original problem which led to timeouts is that starting with a certain iteration the fuzzer (segmatch) was unable to find a clear solution for the PIPs from the todo list.
In this particular case it was caused by the fact that the  PIPs with the ``CLK_HROW_CK_IN_`` source wire, e.g. ``CLK_HROW_BOT_R_CK_BUFG_CASCO7.CLK_HROW_CK_IN_L8`` had 3 candidate bits while the SEGMATCH_ARGS had the limiation set to 2 ( ``-c 2``). This is because in fact only two bits are for the ``CLK_HROW_BOT_R_CK_BUFG_CASCO7.CLK_HROW_CK_IN_L8`` PIP and the third bit is for ``CLK_HROW_CK_IN_L8_ACTIVE``.
In order for segmatch to solve the bits correctly we need another PIP with exactly the same clock source wire on the todo list to disambiguate the _ACTIVE from the PIP bits.

This PR uses the mechanism added in PR #869 
Adjusted the MAKETODO_ARGS in fuzzer 041 to create a balance todo list 
<pre>
--balance-wire-re "^CLK_HROW_" --balance-wire-cnt "2"
</pre>

Another issue this PR addresses is an occasional db-check failure, e.g.
<pre>
Checking kintex7 database
============================
Checking individual .db...
segbits_*.db: 23 okay
mask_*.db: 21 okay
Checking aggregate dir...
ERROR: 2 collisions
  00000c81_048_25: had INT_R_X25Y124.INT_R.CLK1.SR1END1, got CLK_HROW_TOP_R_X67Y130.CLK_HROW_TOP_R.CLK_HROW_CK_INT_0_1_ACTIVE
  00000c81_051_24: had INT_R_X25Y125.INT_R.CLK0.ER1END1, got CLK_HROW_TOP_R_X67Y130.CLK_HROW_TOP_R.CLK_HROW_CK_INT_1_0_ACTIVE

Checking aggregate dir...
ERROR: 1 collisions
  00020c81_048_25: had CLK_HROW_TOP_R_X67Y182.CLK_HROW_TOP_R.CLK_HROW_CK_INT_0_1_ACTIVE, got INT_R_X25Y174.INT_R.CLK1.ER1END1
</pre>

``CLK_HROW_TOP_R.CLK_HROW_CK_INT_1_0_ACTIVE`` which is causing the problem is most probably not real. 
This PR fixes this problem by simply ignoring such PIPs during segdata generation.


fuzzers/041-clk-hrow-pips/generate.py
```diff
@@ -2,9 +2,16 @@
 
 import os
 import os.path
+import re
 from prjxray.segmaker import Segmaker
 
 
+def is_fake_feature(src):
```
litghost: "is_fake_feature" is a poor name.  Maybe "src_has_active_bit" and invert the boolean output

tmichalak: This does sound like a better name. Done.

tmichalak:
List of problems after a specific run:
Artix7:
041-clk-hrow-pips - timeout
056-pip-rem failing with ValueError: INT.FAN_ALT5.BYP_BOUNCE1: inconsistent bit index^ - bit calculated by 050-pip-seed
Kintex7:
053-pip-ctrlin - timeout
Zynq7:
041-clk-hrow-pips - timeout

fuzzers/041-clk-hrow-pips/generate.tcl
```diff
@@ -23,8 +23,33 @@ proc write_pip_txtdata {filename} {
     close $fp
 }
 
+# This function returns the path to the current todo file
```
litghost: I'd like the why of this function in the comment.  The what is less important, but documentation inputs/outputs and the "why" of this function are what I'm interested in.

tmichalak:
CI is red, because the solution didn't work.
I am still looking into it and will re-request review once I am done.

tmichalak:
@litghost I re-requested a review from you. but mostly for the todo list balancing part.
This PR transformed into a DNM branch with all remaining stabilization PRs merged (FAN_ALT.BYP_BOUNCE fix #868 and failing test case packing #856).
Once I can prove that the mechanism works for stabilizing 041 I could create a separate PR for the mechanism and then this PR would only make use of it, i.e. define the extra MAKETODO_FLAGS.

tmichalak:
Issues for fails in Artix and Zynq have been created: #874 and #875.
The fixes are being tested on the DNM PR: #848

mithro:
@tmichalak What is the status of this pull request?

tmichalak:
@mithro: waiting for green CI

tmichalak:
@mithro @litghost CI is green, so is the CI for the experimental stabilization PR #848.
I am not sure if we want to force the CI to rerun for a couple of times before we actually merge it.
However, reenabling 041 is quite important from the harness point of view.
Note, that this PR relies on the todo balancing (#869) which is also green BTW.

tmichalak:
@mithro sure thing, let's wait then.

PR #849, opened 2019-05-24
053-pip-ctrlin: Fall back to todos bigger than specified number of lines

### LINKED ISSUE #843: Timeout on 053-pip-ctrlin
Fuzzer 053 just like 041 (#784) fails with timeout from time to time for each architecture. 

Example log: https://source.cloud.google.com/results/invocations/e8050c23-234b-45c2-a31f-a8b5c91c8023/log

tmichalak:

This PR fixes #843
The origin of the problem is the same as in #838

There are two solution variants:

1. Move the solution of the CTRL*.GFAN01 bits from the 053-pip-ctrlin fuzzer to a new dedicated fuzzer (059-pip-ctrlin-gfan) (https://github.com/antmicro/prjxray/tree/prjxray_stabilization_053_pip_ctrlin_separation). Validated the solution by running both fuzzers for around 30 times without any timeout.
2. Implement the same solution as for #838. This is the preferred solution as it doesn't required adding a new fuzzer to the list.

### DIFF
--- a/fuzzers/053-pip-ctrlin/generate.tcl
+++ b/fuzzers/053-pip-ctrlin/generate.tcl
@@ -21,8 +21,37 @@ route_design
 
 # write_checkpoint -force design.dcp
 
+# This function checks the input todo file to ensure there is enough stimulus.
+# By default if the input todo file is bigger than the specified number of lines then
+# it is used in the subsequent stages of the script.
+# However, when the input todo is smaller then a todo file from a previous iteration,
+# which meets the minimum lines requirement, gets chosen.
+# This helps with increasing the stimuli without increasing the
+# global number of specimen used in every iteration.
+proc get_todo {{min_lines 0}} {
+    if {[info exists ::env(ITER)]} {
+        set current_iter $::env(ITER)
+    } else {
+        set current_iter 1
+    }
+    set todo_file "../../todo/${current_iter}_all.txt"
+    lassign [exec wc -l $todo_file] line_count file_name
+    if {$min_lines == 0} {
+        return $file_name
+    }
+    if {$current_iter == 1 && $line_count < $min_lines} {
+        error "ERROR: Initial TODO is too small"
+    }
+    while {$line_count < $min_lines} {
+        incr current_iter -1
+        set todo_file "../../todo/${current_iter}_all.txt"
+        lassign [exec wc -l $todo_file] line_count file_name
+    }
+    return $file_name
+}
 
-set fp [open "../../todo.txt" r]
+
+set fp [open [get_todo 10] r]
 set todo_lines {}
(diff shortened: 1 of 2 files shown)


mithro:
@tmichalak Why is this preferred over adding a new fuzzer?

tmichalak:
@mithro Basically to avoid having several copies of the same fuzzer differing only in the todo regexp. If it required a change in the segmatch args then this would call for a new fuzzer.

fuzzers/053-pip-ctrlin/generate.tcl
```diff
@@ -20,9 +20,27 @@ place_design
 route_design
 
 # write_checkpoint -force design.dcp
+proc get_todo {} {
```
litghost: Write a function comment please

tmichalak:
Yes, but this only takes the TODOs from the todo list generated for a given iteration which gets smaller from iteration to iteration as we solve new features. What I want to achieve here is override the todo generated for the current iteration and take the todos from a previous iteration.

fuzzers/053-pip-ctrlin/generate.tcl
```diff
@@ -21,8 +21,35 @@ route_design
 
 # write_checkpoint -force design.dcp
 
+# This function returns the path to the current todo file
+# that meets the specified minimum lines criteria, i.e.
+# if the todo file generated in the current iteration has less
+# than the minimum number of lines then the todo file
+# from the previous iteration is used
```
litghost: Why is this important?  I know see what you are doing, but I don't understand how it helps.

tmichalak: What I observed was similar to what we had in 041, it timed out because it reached a todo with PIPs  the fuzzer could not find a solution for. Since the generate.tcl bases on the content of the todo list the more lines it has the bigger the stimuli and more segdata for a given iteration.
The segmatch step in the todo based fuzzers takes into account all existing segdata.
So basically this is an alternative and somehow more targeted way of increasing the stimuli.

litghost: > So basically this is an alternative and somehow more targeted way of increasing the stimuli.

So say this? Your comment is mostly a text description of the function itself.  Try to shorten what it is doing, and focus on the why.  Something akin to:
```
# Checks TODO and ensures there is enough stimulus
#
# By default when input todo is long <explain>
# When input todo is short does <X> to help with <Y>
```

tmichalak:
CI failed because 050 failed to solve the FAN_ALT.BYP_BOUNCE and 056 didn't solve it properly (inconsistent bit index).
Hence it's not directly related to this change but something that has to revisited. I increased the number of specimen in 055 already. Seems that's not enough.

PR #851, opened 2019-05-24
roi_harness: Ignore several unknown bits for Zynq7

tmichalak:

Once the prjxray database generation is stable we will be able to re-enable the test harnesses.
For Artix restoring the harness should just work. For Zynq however, there is a small change required  in the create_design_json script due to some extra bits I am writing about here: https://github.com/SymbiFlow/prjxray/issues/746#issuecomment-490499100 
This PR implements a white listing mechanism per architecture in order to successfully pass the generation of design.json despite the fact that the white listed bits are not in the database

### DIFF
--- a/minitests/roi_harness/create_design_json.py
+++ b/minitests/roi_harness/create_design_json.py
@@ -3,6 +3,7 @@ import csv
 import argparse
 import sys
 import fasm
+import os
 from prjxray.db import Database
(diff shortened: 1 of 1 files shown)


tmichalak:
Will need to create a fuzzer for the PERFRAMECRC and document the remaining bits.

PR #856, opened 2019-05-28
kokoro: Pack failing test cases instead of vivado logs

tmichalak:

When analyzing the root cause of a failing fuzzer we just need its vivado logs.
Therefore there is no need to collect all of them.
What is missing though is often the contents of the build directory.
Collecting it for all fuzzers would kill the CI.
This PR implements the creation of an tgz archive with only the failing test cases.

### DIFF
--- a/.github/kokoro/db-full.sh
+++ b/.github/kokoro/db-full.sh
@@ -54,18 +54,19 @@ echo "----------------------------------------"
 	# which is well before the 05x INT fuzzers complete.
 	export MAX_VIVADO_PROCESS=$((CORES/2 < 20 ? CORES/2 : 20))
 	set -x +e
-	script --return --flush --command "make -j $CORES MAX_VIVADO_PROCESS=$MAX_VIVADO_PROCESS" -
+	tmp=`mktemp`
+	script --return --flush --command "make -j $CORES MAX_VIVADO_PROCESS=$MAX_VIVADO_PROCESS" $tmp
 	DATABASE_RET=$?
 	set +x -e
-	echo "----------------------------------------"
-
-	# Collect the Vivado logs into one tgz archive
-	echo "Collecting Vivado logs"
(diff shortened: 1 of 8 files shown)


.github/kokoro/db-full.sh
```diff
@@ -54,18 +54,18 @@ echo "----------------------------------------"
 	# which is well before the 05x INT fuzzers complete.
 	export MAX_VIVADO_PROCESS=$((CORES/2 < 20 ? CORES/2 : 20))
 	set -x +e
-	script --return --flush --command "make -j $CORES MAX_VIVADO_PROCESS=$MAX_VIVADO_PROCESS" -
+	tmp=`mktemp`
+	script --return --flush --command "make -j $CORES MAX_VIVADO_PROCESS=$MAX_VIVADO_PROCESS" $tmp
```
litghost: As far as I can tell, this printed to STDOUT before.  Should you be using "tee" here?

tmichalak: It still is printed to STDOUT.
From script manual:
> (..)script  makes a typescript of everything displayed on your terminal. (...) If the argument file is given, script saves the dialogue in this file.  If no filename is given, the dialogue is saved in the file typescript.

I just added the temp file to easily delete it after is it used.

tmichalak: This doesn't change the fact that there is a bug in where I delete the temp file. It fails because it is in a different scope. I will fix that right away.

tmichalak:
@litghost 
I added the force-run label in order to get the CI green, but this case is exactly as in #831 - it failed on fuzzer 53. In this case however we can wait until 53 is stable and merged and then rebase this PR.

tmichalak:
@litghot: Artix failed because of the issue that PR #868 fixes. We can wait and rebase this PR on master once #868 is merged or merge this already.

PR #857, opened 2019-05-28
Fixed fuzzer 007 to make it extract missing timings for FFs

mkurc-ant:
This fix allows the 007 fuzzer to correctly extract timings SR -> QL and SR -> QH for flip-flops and latches.

I added support for explicit pin name aliases of BELs. There is also an explicit BEL suffix map. It stores timing name suffixes that has to be preserved. Both maps are stored in JSON files.


fuzzers/007-timing/bel/bel_suffix_map.json
```diff
@@ -0,0 +1,4 @@
+{
+    "ff_init": ["ql", "qh"],
```
litghost: This seems duplicate-ish with the pin_alias.  Should these be 1 file?

mkurc-ant: It is not a duplicate. This map is intended to provide explicit suffixes of a bel name which cannot be detected using other manners.

For example, for the FF_INIT we have different SR to Q timings depending on whether we are setting Q to 0 or 1 asynchronously. This corresponds do either FDCE (`*_ql`) or FDPE (`*_qh`) primitive. So we need to have separate FF_INIT_QL and FF_INIT_QH in SDF as their counterparts. The same is for the REG_INIT_FF bel.

Now, the timings for FF_INIT are reported by Vivado as following:

```
ff_init_ce_clk
ff_init_ce_clk
ff_init_din_clk
ff_init_din_clk
ff_init_init_clk
ff_init_sr_clk_qh
ff_init_sr_clk_ql
ff_init_sr_clk_qh
ff_init_sr_clk_ql
ff_init_init_clk
ff_init_clk_q
ff_init_sr_qh
ff_init_sr_ql
```

Note that we have eg. `ff_init_sr_clk_qh` and `ff_init_sr_qh`. 

The former is a recovery/remove time for SR w.r.t. CLK for the FDPE. The latter corresponds to an async iopath timing between SR and Q. Those names refer both to the output pin (Q) and bel configuration variant (`ql` and `qh`). 

I added the alias map for pin names so we can correctly detect that both QL and QH refer to the same pin Q. But we also need to know that those names are not just pin aliases. They describe the configuration variants.

I am open to suggestions if you have a different solution to the problem in mind.

@kgugala We also had a discussion about that. You may want to look into the explanation above.

kgugala: I agree with @litghost that this is duplicating info from pin_mappings. My idea here is to use the custom pin mappings as the last resort when we were unable to find the pins. If a pin is found using this mapping we emit a timing entry using the pin name, for cell instance named with the original property.
@mkurc-ant does that make sense?

mkurc-ant:
@kgugala I changed the implementation according to you suggestion. It works as expected.

fuzzers/007-timing/bel/tim2json.py
```diff
@@ -166,47 +214,54 @@ def read_raw_timings(fin, properties, pins, site_pins):
                                 elif pins[slice][site_name][delay_btype_orig][
                                         orig_pin]['direction'] == 'OUT':
                                     bel_output = pin
-                                speed_model_clean = remove_pin_from_model(
-                                    pin.lower(), speed_model_clean)
+                                if not is_alias:
```
kgugala: you check this in every case. Maybe it is worth moving this check to `remove_pin_from_model` to avoid code duplication?

otherwise LGTM

mkurc-ant: Yes but effectively it would be calling a function "remove" with an argument which says "do not remove".

kgugala: alternatively you may change the remove function to not fail if the provided pin was not found in the bel string and simply return the original string.

I really don't like the `if not is_alias` repeated in so many places

mkurc-ant: Ok, fixed that.

kgugala:
I'm hijacking this PR - I need a similar functionality to handle BRAM timings, so instead of creating a new functionality I'll refactor this to be more generic

kgugala:
I have rebased this PR and added all the features needed to extract the timings for BRAM's.

@litghost can you take a look on this?

fuzzers/007-timing/bel/fixup_timings_txt.py
```diff
@@ -0,0 +1,77 @@
+#!/usr/bin/env python3
+
+import argparse
+
+
+def fix_line(line, site, filetype):
```
litghost: This needs some explanation of what is going on

fuzzers/007-timing/bel/pin_alias_map.json
```diff
@@ -0,0 +1,60 @@
...
+    "reg_init": {
+        "Q": {
+             "names" : ["QL", "QH"],
+             "property" : true
+         }
+    },
+    "rambfifo36e1" : {
+         "RSTRAMARSTRAMLRST" : {
+             "names" : ["rstramarstl"],
+             "property" : false
```
litghost: "property"?  Not sure I understand what this field is for or how to use it?

fuzzers/007-timing/bel/tim2json.py
```diff
@@ -63,26 +93,59 @@ def pin_in_model(pin, model, direction=None):
             return True, pin
         elif extended_pin_name in model.split('_'):
             return True, extended_pin_name
+        elif (pin_aliases is not None) and (pin.upper() in pin_aliases):
+            for alias in pin_aliases[pin.upper()]['names']:
```
litghost: Factor this loop into a function and add some tests

fuzzers/007-timing/bel/tim2json.py
```diff
@@ -63,26 +93,59 @@ def pin_in_model(pin, model, direction=None):
...
     else:
         # pin name is multi word, search for a string
-        return (pin in model), pin
+        if pin in model:
+            return True, pin
+        elif pin_aliases is not None:
+            for alias in pin_aliases.get(pin.upper(), ()):
+                pin_alias = alias.lower()
+                if pin_alias in model:
+                    return True, pin
```
litghost: This loop can fall off without a return?  Missing assert?

fuzzers/007-timing/bel/pin_alias_map.json
kgugala: probably not the best name.

There are two types of aliases we need to handle. One is a pin name combined with a property e.g. `ql` or `qh` in `dff`. Here the actual pin name is `q` while `l` and `h` determines the `S` or `R` configuration of the `dff`. In this case we should emit a different celltype timing for the considered timing path (e.g. `DFF_S` or `DFF_R`)

The second type is when the actual BEL pin name is different than the one included in the timing string e.g. BRAM's `rstramau` from timing string refers to the `RSTRAMARSTRAMU` BEL pin. In this case we simply translate the pin name.

Maybe this field should be renamed to something like `is_property_related`?

fuzzers/007-timing/bel/fixup_timings_txt.py
kgugala: done

fuzzers/007-timing/bel/tim2json.py
kgugala: done

fuzzers/007-timing/bel/tim2json.py
kgugala: the assert is not necessary after refactoring

fuzzers/007-timing/bel/fixup_timings_txt.py
```diff
@@ -0,0 +1,98 @@
...
+    """
+    Squashes the entries for multilpe sites into one.
```
litghost: Spelling?

fuzzers/007-timing/bel/fixup_timings_txt.py
```diff
@@ -0,0 +1,98 @@
...
+    This is required when entries are defined for a different
+    site than they are reported.
+    Such situation happend e.g.
+    for BRAM_[LR]. All the entries are defined for RAMBFIFO36E1,
+    while they are reported for RAMB18E1 or FIFO18E1
+
+    Args:
+    line: raw dump file line
```
litghost: Use numpy docstring style?

fuzzers/007-timing/bel/fixup_timings_txt.py
```diff
@@ -0,0 +1,98 @@
...
+    assert filetype in [
+        "timings", "pins", "properties"
+    ], "Unsupported filetype"
+
+    line = line.split()
+    newline = list()
+    sites_count = int(line[1])
+    newline.append(line[0])
+    # we'll emit only one site
+    newline.append("1")
```
litghost: I think a short description of the input file format is required to make sense of this code.  Maybe add a description in a README?  Or something?  I find this code hard to review because it assumes a file structure.

Does it make any sense to first make a more structured form, and then manipulate the structured form?

fuzzers/007-timing/bel/tim2json.py
```diff
@@ -45,11 +45,87 @@ def clean_bname(bname):
     return bname
 
 
-def pin_in_model(pin, model, direction=None):
+def find_aliased_pin(pin, model, pin_aliases):
+    """
+    Searches for aliased pins
+
+    Args:
+        pin: pin name to look for
```
litghost: Use numpy docstring style?

mithro:
@kgugala What is the status of this pull request?

fuzzers/007-timing/bel/fixup_timings_txt.py
kgugala: I added README to the BEL fuzzer describing how it wokrs and files format

fuzzers/007-timing/bel/fixup_timings_txt.py
kgugala: fixed

fuzzers/007-timing/bel/fixup_timings_txt.py
kgugala: done

fuzzers/007-timing/bel/tim2json.py
kgugala: done

kgugala:
> @kgugala What is the status of this pull request?

I have addressed all the comments and rebased it on top of current master. It is ready for review

fuzzers/007-timing/bel/pin_alias_map.json
```diff
@@ -0,0 +1,60 @@
+{
+    "ff_init": {
+        "Q": {
+             "names" : ["QL", "QH"],
+             "is_property_related" : true
```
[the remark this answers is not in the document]
kgugala: I must have missed that. I have updated the docstrings so they discuss this

kgugala:
I have rebased this PR on top of current master, as there were some stabilization fixes merged. Maybe this will make the CI green

Issue #861, opened 2019-05-31
Stabilization issues related to FAN_ALT.BYP_BOUNCE 


tmichalak:
We tend to get red CI due to some problems with solving the FAN_ALT.BYP_BOUNCE bits.

Error 1:
<pre>
2019-05-25T23:36:18.124223 - 056-pip-rem              - 4m02s:   File "/tmpfs/src/github/symbiflow-prjxray-presubmit-db-artix7/utils/dbfixup.py", line 153, in add_zero_bits^M
2019-05-25T23:36:18.124223 - 056-pip-rem              - 4m02s:     zero_range(tag, bits, 22, 25)^M 
2019-05-25T23:36:18.124223 - 056-pip-rem              - 4m02s:   File "/tmpfs/src/github/symbiflow-prjxray-presubmit-db-artix7/utils/dbfixup.py", line 36, in zero_range^M
2019-05-25T23:36:18.124223 - 056-pip-rem              - 4m02s:     raise ValueError("%s: inconsistent bit index" % tag)^M
2019-05-25T23:36:18.124223 - 056-pip-rem              - 4m02s: ValueError: INT.FAN_ALT5.BYP_BOUNCE1: inconsistent bit index^M
</pre>

This is caused by the fact that the fuzzer that solves this PIP gets a solution with an incorrect bit index e.g.
<pre>
Old bit index: 46, new: 31
bits: {'20_46', '25_46', '25_31'}
2019-05-12T16:11:55.799065 - 056-pip-rem              - 1h14m:     zero_range(tag, bits, 22, 25) 
2019-05-12T16:11:55.799065 - 056-pip-rem              - 1h14m:   File "/home/build/tmichalak/stabilization/prjxray/utils/dbfixup.py", line 36, in zero_range
2019-05-12T16:11:55.799065 - 056-pip-rem              - 1h14m:     raise ValueError("%s: inconsistent bit index" % tag)
ValueError: INT.FAN_ALT5.BYP_BOUNCE1: inconsistent bit index
</pre>

Error 2:
<pre>
ERROR: 1 collisions00401280_076_04: had INT_R_X13Y37.INT_R.FAN_ALT6.BYP_BOUNCE1, got CLBLL_R_X13Y37.CLBLL_R.SLICEL_X1.CEUSEDMUX
</pre>
This is because 050-pip-seed fuzzer calculates the following bits: 
<pre>
[...]

PR #868, opened 2019-06-05
Fix instability issues with the FAN_ALT.BYP_BOUNCE PIPs

tmichalak:
This PR fixes issue #861 
Background:
The FAN_ALT.BYP_BOUNCE bits were either solved by the 050-pip-seed fuzzer or eventually by the 056-pip-rem. The problems described in the issue were caused by insufficient stimuli which was very random due to the very random nature of 050 and 056.
Solution:
Add a dedicated fuzzer solely for the FAN_ALT.BYP_BOUNCE bits and remove them from the list of PIPs to be solved by the 050-pip-seed fuzzer.


fuzzers/Makefile
```diff
@@ -104,6 +104,7 @@ $(eval $(call fuzzer,056-pip-rem,049-int-imux-gfan 050-pip-seed 051-pip-imuxlout
 $(eval $(call fuzzer,057-pip-bi,056-pip-rem))
 ifneq ($(QUICK),Y)
 $(eval $(call fuzzer,058-pip-hclk,005-tilegrid))
+$(eval $(call fuzzer,059-pip-byp-bounce,048-int-piplist))
```
litghost: pip-byp-bounce seems like it should be done before 056?

tmichalak: Yes, you are correct. Fixed.

tmichalak:
@litghost it's green now

PR #869, opened 2019-06-06
Implement todo lists balancing mechanism

tmichalak:
This PR adds a mechanism for the creation of balanced todo lists.

Background
This mechanism is needed for fixing instability problems in some fuzzers which led to timeouts. 
The root cause of the timeout is that starting with a certain iteration the fuzzer (segmatch) was unable to find a clear solution for the PIPs from the todo list.
In a particular case (#784) the timeout was caused by the fact that the  PIPs with the ``CLK_HROW_CK_IN_`` source wire, e.g. ``CLK_HROW_BOT_R_CK_BUFG_CASCO7.CLK_HROW_CK_IN_L8`` had 3 candidate bits while the SEGMATCH_ARGS had the limiation set to 2 ( ``-c 2``). This is because in fact only two bits are for the ``CLK_HROW_BOT_R_CK_BUFG_CASCO7.CLK_HROW_CK_IN_L8`` PIP and the third bit is for ``CLK_HROW_CK_IN_L8_ACTIVE``.
In order for segmatch to solve the bits correctly we need another PIP with exactly the same clock source wire on the todo list to disambiguate the _ACTIVE from the PIP bits.

Implementation
The mechanism was added to the int_maketodo.py script responsible for the creation of todo lists.
It was designed to be generic in order to create more balanced todo list based on few parameters specified with the use of additional switches (--balance-wire-re , --balance-wire-direction and --balance-wire-cnt)
Example todo list:
<pre>
CLK_HROW_TOP_R.CLK_HROW_TOP_R_CK_BUFG_CASCO0.CLK_HROW_CK_IN_R8
CLK_HROW_TOP_R.CLK_HROW_TOP_R_CK_BUFG_CASCO21.CLK_HROW_CK_IN_R9
</pre>

Command to create a balanced todo file:
<pre>
python3 ../int_maketodo.py --sides "bot_r,top_r" --pip-type clk_hrow --seg-type clk_hrow --re ".*" --balance-wire-re "^CLK_HROW_CK_IN_" --balance-wire-direction "src" --balance-wire-cnt "2"
</pre>

Final, balanced todo list:
<pre>
CLK_HROW_TOP_R.CLK_HROW_TOP_R_CK_BUFG_CASCO27.CLK_HROW_CK_IN_R8
CLK_HROW_TOP_R.CLK_HROW_TOP_R_CK_BUFG_CASCO21.CLK_HROW_CK_IN_R9
CLK_HROW_TOP_R.CLK_HROW_TOP_R_CK_BUFG_CASCO0.CLK_HROW_CK_IN_R8
CLK_HROW_TOP_R.CLK_HROW_TOP_R_CK_BUFG_CASCO7.CLK_HROW_CK_IN_R9
</pre>

Another example:
Initial todo:
<pre>
CLK_HROW_TOP_R.CLK_HROW_TOP_R_CK_BUFG_CASCO0.CLK_HROW_CK_IN_R8
CLK_HROW_TOP_R.CLK_HROW_TOP_R_CK_BUFG_CASCO0.CLK_HROW_CK_IN_R9
</pre>
Command:
<pre>
python3 ../int_maketodo.py --sides "bot_r,top_r" --pip-type clk_hrow --seg-type clk_hrow --re ".*" --balance-wire-re "^CLK_HROW_TOP_R_CK_BUFG_CASCO" --balance-wire-direction "dst" --balance-wire-cnt "4"
</pre>
Balanced todo:
<pre>
CLK_HROW_TOP_R.CLK_HROW_TOP_R_CK_BUFG_CASCO0.CLK_HROW_CK_IN_R3
CLK_HROW_TOP_R.CLK_HROW_TOP_R_CK_BUFG_CASCO0.CLK_HROW_CK_IN_R9
CLK_HROW_TOP_R.CLK_HROW_TOP_R_CK_BUFG_CASCO0.CLK_HROW_CK_IN_R8
CLK_HROW_TOP_R.CLK_HROW_TOP_R_CK_BUFG_CASCO0.CLK_HROW_CK_IN_L0
</pre>

### DIFF
--- a/fuzzers/int_maketodo.py
+++ b/fuzzers/int_maketodo.py
@@ -33,8 +33,78 @@ def load_pipfile(pipfile, verbose=False):
     return todos, tile_type
 
 
+def balance_todo_list(
+        todos,
+        balance_wire_re,
+        balance_wire_direction,
+        balance_wire_cnt,
+        verbose=False):
+    """Balance the contents of the todo list
+
+    Todo list balancing allows to specify the name, direction and minimal number of
+    occurrences of a PIP wire in the final todo list.
+    The mechanism should be used in cases where a fuzzer times out because of an
+    unsolvable todo list, i.e. the netlist and resulting segdata generated from an
+    iteration keep segmatch from properly disambiguating the bits for some features.
+
+    When the balance wire name regexp is specified it's guaranteed that all PIPs
+    with matching wire name (whether we want to match a src or dst wire has to be
+    specified with the --balance-wire-direction switch) will have at least the number
+    of entries specified with the --balance-wire-cnt switch in the final todo list.
+    """
+    if balance_wire_re is not None:
+        todo_wires = {}
+        verbose and print("Start balancing the TODO list")
+        for todo in todos:
+            tile_type, dst, src = todo.split(".")
+            wire = src
+            other_wire = dst
+            if balance_wire_direction not in "src":
+                wire = dst
+                other_wire = src
+            balance_wire_match = re.match(balance_wire_re, wire)
+            if balance_wire_match is None:
+                continue
+            if wire not in todo_wires:
+                todo_wires[wire] = set()
+            todo_wires[wire].add(other_wire)
+        for wire, other_wires in todo_wires.items():
+            if len(other_wires) >= balance_wire_cnt:
+                continue
+            else:
+                for todo in orig_todos:
+                    tile_type, dst, src = todo.split(".")
+                    if balance_wire_direction in "src":
+                        if wire in src and dst not in todo_wires[wire]:
+                            todo_wires[wire].add(dst)
+                    else:
+                        if wire in dst and src not in todo_wires[wire]:
+                            todo_wires[wire].add(src)
+                    if len(todo_wires[wire]) == balance_wire_cnt:
+                        break
+        for wire, other_wires in todo_wires.items():
+            assert len(other_wires) >= balance_wire_cnt, "Len is " + str(
+                len(other_wires))
+            for other_wire in other_wires:
+                line = tile_type + "."
+                if balance_wire_direction in "src":
+                    line += other_wire + "." + wire
+                else:
+                    line += wire + "." + other_wire
+                verbose and print("Adding {}".format(line))
+                todos.add(line)
+        verbose and print("Finished balancing the TODO list")
+
+
 def maketodo(
-        pipfile, dbfile, intre, exclude_re=None, not_endswith=None,
+        pipfile,
+        dbfile,
+        intre,
+        exclude_re=None,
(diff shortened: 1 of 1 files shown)


tmichalak:
CI is green, i.e. it doesn't break anything.

fuzzers/int_maketodo.py
```diff
@@ -86,12 +96,58 @@ def maketodo(
...
 
         if include:
-            print(line)
+            filtered_todos.add(line)
         else:
             drops += 1
         lines += 1
     verbose and print('Print %u entries w/ %u drops' % (lines, drops))
 
+    if balance_wire_re is not None:
```
mithro: Please add a comment which describes what balancing is, why it is needed and the theory of how it works.

tmichalak:
@mithro Just added some comment. Could you take a look if it's clear enough?

fuzzers/int_maketodo.py
mithro: You wrapping of these comments is very strange. Can you rewrap it?

tmichalak: Should be better now?

fuzzers/int_maketodo.py
```diff
@@ -86,11 +96,70 @@ def maketodo(
             include = re.match(exclude_re, line) is None
 
         if include:
-            print(line)
+            filtered_todos.add(line)
         else:
             drops += 1
         lines += 1
     verbose and print('Print %u entries w/ %u drops' % (lines, drops))
+    '''
```
mithro: This should still be a comment rather than a string.

tmichalak: I wanted to be consistent with the rest of the comments in the int_maketodo.py file. These other comments also tend to use multi line strings for commenting.
Should I change it to '#' anyway?

mithro: They are docstrings - see https://www.python.org/dev/peps/pep-0257/

tmichalak: Changed it to multi-line docstrings then.

mithro: A docstring goes on a function, not in the middle of a function.
```python
def abc(xxx):
    ''''This is a doc string''''
    pass

def abc(xxx):
    """This is a doc string"""
    pass

def abc(xxx):
    a = 1

    """This is a *not* doc string"""

   y = 1
```

tmichalak: This actually made me think that it would be cleaner to extract the balancing code into a function and then use a doc string instead of a long comment.

mithro: That is a good idea.

tmichalak: Done

PR #870, opened 2019-06-06
An utility script that helps to find missing segbits

mkurc-ant:
This script allows to list PIPs which are missing segbits. It does not check any "non-pip" features.

The script operates on the prjxray files. It looks for all non-pseudo PIPs in each tile and then tries to find segbits for them. If a PIP is mared as non-pseudo in the `tile_type_*.json` file and is defined in `ppips_*.db` as "always" or "hint" then it is considered as pseudo.

A name pattern can be specified for tile types of interest and tile types of non-interest.

### DIFF
--- a/utils/find_missing_segbits.py
+++ b/utils/find_missing_segbits.py
@@ -8,6 +8,7 @@ and checks if those pips are listed there.
 
 Missing segbits for pips are reported as well as missing segbit files.
 """
+import sys
 import logging
 import json
 import argparse
@@ -74,7 +75,7 @@ def read_segbits(segbits_file):
 # =============================================================================
 
 
-def main():
+def main(argv):
     """
     The main
     """
@@ -100,7 +101,7 @@ def main():
         default=[],
         help="Tile type name regex list for tile types to include")
 
-    args = parser.parse_args()
+    args = parser.parse_args(argv[1:])
 
     logging.basicConfig(level=50 - args.verbose * 10, format="%(message)s")
 
@@ -214,8 +215,10 @@ def main():
         else:
             logging.warning(" OK: no pips")
 
+    return 0
+
 
 # =============================================================================
 
 if __name__ == "__main__":
-    main()
+    sys.exit(main(sys.argv))
--- a/utils/find_missing_segbits.py
+++ b/utils/find_missing_segbits.py
@@ -80,6 +80,8 @@ def main(argv):
     The main
     """
 
+    exitcode = 0
+
     # Parse arguments
     parser = argparse.ArgumentParser(
         description=__doc__,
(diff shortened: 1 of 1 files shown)


mkurc-ant:
@litghost Could you please verify if I didn't miss anything in the algorithm ?

utils/find_missing_segbits.py
```diff
@@ -0,0 +1,221 @@
...
+
+        # No pips
+        else:
+            logging.warning(" OK: no pips")
+
+
+# =============================================================================
+
+if __name__ == "__main__":
+    main()
```
mithro:
```suggestion
    import sys
    sys.exit(main(sys.argv))
```

utils/find_missing_segbits.py
```diff
@@ -0,0 +1,221 @@
...
+            if missing_bits > 0:
+                logging.critical(
+                    " MISSING: no bits for {}/{} pips!".format(
+                        missing_bits, missing_bits + known_bits))
+            else:
+                logging.critical(" OK: no missing bits")
+
+        # No pips
+        else:
+            logging.warning(" OK: no pips")
```
mithro:
```suggestion
            logging.warning(" OK: no pips")
            return 0
```

utils/find_missing_segbits.py
```diff
@@ -0,0 +1,221 @@
...
+
+                    # The pip has segbits
+                    else:
+                        known_bits += 1
+
+            # Report missing bit count
+            if missing_bits > 0:
+                logging.critical(
+                    " MISSING: no bits for {}/{} pips!".format(
+                        missing_bits, missing_bits + known_bits))
```
mithro:
```suggestion
                logging.critical(
                    " MISSING: no bits for {}/{} pips!".format(
                        missing_bits, missing_bits + known_bits))
                return 1
```

utils/find_missing_segbits.py
mkurc-ant: Fixed that.

utils/find_missing_segbits.py
mkurc-ant: What is the difference ?

utils/find_missing_segbits.py
mkurc-ant: Again, what's the difference ?

utils/find_missing_segbits.py
plaes: Function returns with `0` and which is then propagated to `sys.exit(main(sys.argv))` below...

mkurc-ant: Yep, that one I got. My question is about the other two issues.

utils/find_missing_segbits.py
mithro: You want a non-zero exit code when missing bits are found.

mkurc-ant: Ok, fixed that. Now the script returns -1 if there are any missing bits found.

tmichalak:
@mkurc-ant Please rebase the branch on the current master which contains a kokoro runtime fix (#879)

mkurc-ant:
@tmichalak Done

PR #881, opened 2019-06-13
Litex SoC minitest

mkurc-ant:
This test generates LiteX SoC design which utilizes VexRiscV CPU, DDR memory and Ethernet. The design is implemented using both Yosys+Vivado and Vivado only flow. Output bitstream(s) is then processed using bit2fasm utility which allows to identify missing degbits.


mkurc-ant:
@mithro I agree. I decline this PR and will issue a new one.

Issue #875, opened 2019-06-07
Duplicate tag in 044-clk-bufg-pips


tmichalak:
**Symptom**
Duplicate tags.
(https://source.cloud.google.com/results/invocations/3a54b56c-a5c7-432d-954b-89b40c0cf2ed/targets )

<pre>
XRAY_DATABASE_DIR=/tmpfs/src/github/symbiflow-prjxray-presubmit-db-zynq7/fuzzers/044-clk-bufg-pips/build/database bash /tmpfs/src/github/symbiflow-prjxray-presubmit-db-zynq7/utils/mergedb.sh clk_bufg_bot_r build/segbits_clk_bufg_bot_r.db
WARNING: got duplicate tag CLK_BUFG_BOT_R.CLK_BUFG_BUFGCTRL10_I0.CLK_BUFG_BOT_R_CK_MUXED20
  Orig line: CLK_BUFG_BOT_R.CLK_BUFG_BUFGCTRL10_I0.CLK_BUFG_BOT_R_CK_MUXED20 !26_167 !26_23 !26_24 !27_166 26_168 27_22
  New line : CLK_BUFG_BOT_R.CLK_BUFG_BUFGCTRL10_I0.CLK_BUFG_BOT_R_CK_MUXED20 !26_167 !27_166 26_168 
</pre>
or 
<pre>
XRAY_DATABASE_DIR=/tmpfs/src/github/symbiflow-prjxray-presubmit-db-kintex7/fuzzers/044-clk-bufg-pips/build/database bash /tmpfs/src/github/symbiflow-prjxray-presubmit-db-kintex7/utils/mergedb.sh clk_bufg_top_r build/segbits_clk_bufg_top_r.db
WARNING: got duplicate tag CLK_BUFG_TOP_R.CLK_BUFG_BUFGCTRL10_I1.CLK_BUFG_IMUX26_2
  Orig line: CLK_BUFG_TOP_R.CLK_BUFG_BUFGCTRL10_I1.CLK_BUFG_IMUX26_2 !26_164 !26_244 !27_165 !27_245 26_165 26_245
  New line : CLK_BUFG_TOP_R.CLK_BUFG_BUFGCTRL10_I1.CLK_BUFG_IMUX26_2 !26_164 !27_165 26_165
</pre>
**Analysis**
[...]

PR #848, opened 2019-05-24
[DNM] WIP: Prjxray stabilization test

tmichalak:
This PR is to test all stabilization related fixes together on the GCS runner.
There are separate PRs for each of the commits.

PR #889, opened 2019-06-17
Fix duplicate tag in 044-clk-bufg-pips

tmichalak:
This PR solves issue #875

PIPs with CLK_BUFG_BOT_R_CK_MUXED src wire were solved incorrectly. If the segmatch constraint was -c 1 then it would have been solved correctly, however in order to solve some other 2 bit features in this fuzzer the number of candidates had to be set to 2 (-c 2)

In this PR a separate fuzzer is created for the clk-bufg-muxed PIPs and their calculation is moved from fuzzer 044 to the newly created.

The fix has been tested on the DNM branch (#848)


fuzzers/046-clk-bufg-muxed-pips/generate.py
```diff
@@ -0,0 +1,99 @@
+#!/usr/bin/env python3
+
+from prjxray.segmaker import Segmaker
+import os
+import os.path
+import re
+
+
+def bitfilter(frame, word):
+    if frame < 26:
```
plaes: `return frame >= 26`

tmichalak: This is of course a valid suggestion, but in my opinion the explicit returns make it easier to understand what is filtered out by the bitfilter.

tmichalak:
Red CI on Zynq, will investigate if this is related to this change or another sign of the instability.

tmichalak:
The red CI on Zynq is related to #895.
I am re-running the build, I think the chances for #895 to appear are pretty thin.

tmichalak:
@litghost Do you mean the dbf? I removed the contents on purpose since the PIPs in bits.dbf from fuzzer 044 are not being calculated in this fuzzer, nor are they in the segdata files and therefore appear in the rdb as <const0>. 
BTW when pasting a link could you make sure it's through ``console.cloud.google.com`` rather than ``cnsviewer2.corp.google.com`` - I don't have access to the latter.

litghost:
> @litghost Do you mean the dbf?

Ya, dbf.  If you look at the segbits diff, the new output is missing some of the negative bits defined by the dbf, so it was required and you are no longer using it.

litghost:
Please rebase and lets re-run.

PR #896, opened 2019-06-18
Fix problem with falsely ignored PIPs

### LINKED ISSUE #895: PIPs falsely ignored by the int_generate.py script.
One of the CI runs (https://source.cloud.google.com/results/invocations/b5621a37-f0c2-43fb-a4c1-f8a8d50ef71c/targets) revealed an issue in the int_generate.py script causing the 056-pip-rem fuzzer to fail occasionally with the following message:
<pre>
Compiling segment data.
  File "/home/tmichalak/Development/X-Ray/prjxray2/fuzzers/int_generate.py", line 94, in <module>
    segmk.write()
  File "/home/tmichalak/Development/X-Ray/prjxray2/prjxray/segmaker.py", line 399, in write
    ]) != 0, "Didn't  generate any segments"
AssertionError: Didn't  generate any segments
</pre>

Problem description:
In the failing test case we had the following PIP in the todo list:
<pre>
INT_R.LH0.SS6END3
</pre>

whereas the design, read by the int_generate.py script from the generated design.txt had among others the same PIP but on an INT_L site:
<pre>
INT_R.LH0.SS6END3
INT_L.LH0.SS6END3
</pre>

The script implements an ignoring mechanism for PIPs that are not in the todo list, however it only adds the src and dst wires to the ignore pip list, without the tile information.
This leads to falsely adding a PIP present in the todo list to the list of ignored PIPs.

Solution:
Obviously, we also need to keep the information whether the PIP is on the INT_L or INT_R tile up to the point where the ignoring mechanism comes in.

tmichalak:

This PR fixes #895 
A detailed explanation of the problem can be found in the issue.
This PR modifies the script to keep the information whether the PIP is on the INT_L or INT_R tile up to the point where the ignoring mechanism comes in.

### DIFF
--- a/fuzzers/int_generate.py
+++ b/fuzzers/int_generate.py
@@ -54,10 +54,11 @@ with open(args.design, "r") as f:
         if tile not in tiledata:
             tiledata[tile] = {"pips": set(), "srcs": set(), "dsts": set()}
 
-        if pip in pipdata:
-            assert pipdata[pip] == (src, dst)
+        tile_lr = ("_".join(tile.split("_")[0:2]))
+        if (tile_lr, pip) in pipdata:
+            assert pipdata[(tile_lr, pip)] == (src, dst)
         else:
-            pipdata[pip] = (src, dst)
+            pipdata[(tile_lr, pip)] = (src, dst)
 
         tiledata[tile]["pips"].add(pip)
         tiledata[tile]["srcs"].add(src)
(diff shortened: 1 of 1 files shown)


fuzzers/int_generate.py
```diff
@@ -54,10 +54,11 @@
         if tile not in tiledata:
             tiledata[tile] = {"pips": set(), "srcs": set(), "dsts": set()}
 
-        if pip in pipdata:
-            assert pipdata[pip] == (src, dst)
+        tile_lr = ("_".join(tile.split("_")[0:2]))
```
litghost: Not all interconnect fuzzers have _LR suffixes.  This may break in those cases.

tmichalak: @litghost The CI results and diffs look fine to me? Could you take a look and tell me if you see something plain wrong there?

litghost: I'd like to hold off on this PR until 044 is enabled.

tmichalak:
@litghost the CI results look good to me as far as the problem which was addressed here is concerned.
However while browsing the results something else came up: #914.

litghost:
> @litghost the CI results look good to me as far as the problem which was addressed here is concerned.
> However while browsing the results something else came up: #914.

Should this PR be merged? Or will this PR be replaced by something that addresses #914.

tmichalak:
This should be merged. It addresses a different and quite rare but still, stability issue.

Issue #817, opened 2019-05-03
Add feature for each input of a mux


HackerFoo:
Example:

```
CLBLM_L.SLICEM_X0.ALUT.DI1MUX.AI 00_00
CLBLM_L.SLICEM_X0.ALUT.DI1MUX.BDI1 !00_00
CLBLM_L.SLICEM_X0.BLUT.DI1MUX.BI 00_20
CLBLM_L.SLICEM_X0.BLUT.DI1MUX.DI !00_20
```

This makes it easier to read, and will allow detecting conflicts.

mithro:
A grep inside https://github.com/SymbiFlow/prjxray-db/blob/master/artix7/segbits_clblm_l.db gives;
```
CLBLM_L.SLICEM_X0.ALUT.DI1MUX.AI 00_00
CLBLM_L.SLICEM_X0.BLUT.DI1MUX.BI 00_20
CLBLM_L.SLICEM_X0.CLUT.DI1MUX.CI 01_43
```

Should this really be;
```
CLBLM_L.SLICEM_X0.DI1MUX.AI 00_00
CLBLM_L.SLICEM_X0.DI1MUX.BI 00_20
CLBLM_L.SLICEM_X0.DI1MUX.CI 01_43
```

mithro:
This seems to be done by the https://github.com/SymbiFlow/prjxray/tree/master/fuzzers/019-clb-ndi1mux fuzzer...

mithro:
https://github.com/SymbiFlow/prjxray/tree/master/fuzzers/019-clb-ndi1mux/minitest has some more comments?

mithro:
BTW If it is the missing zero bits -- then bits.dbf file is what should be used I think?

See https://github.com/SymbiFlow/prjxray/blob/36e6290ec6638f88962d23582a3c5329e6d665d3/fuzzers/036-iob-ologic/bits.dbf#L1-L3 and https://github.com/SymbiFlow/prjxray/blob/36e6290ec6638f88962d23582a3c5329e6d665d3/fuzzers/036-iob-ologic/Makefile#L9-L11

litghost:
> A grep inside https://github.com/SymbiFlow/prjxray-db/blob/master/artix7/segbits_clblm_l.db gives;
> 
> ```
> CLBLM_L.SLICEM_X0.ALUT.DI1MUX.AI 00_00
> CLBLM_L.SLICEM_X0.BLUT.DI1MUX.BI 00_20
[...]

PR #900, opened 2019-06-19
Add support to zero db to support simple groups.

litghost:
Previously these kinds of zero groups would require encoding the
final bits, rather than tags.  This is extends the dbfixup to
construct groups via groups of tags, rather than groups of bits.

This is PR adds support to enable https://github.com/SymbiFlow/prjxray/issues/817 to be completed.


utils/dbfixup.py
```diff
@@ -51,57 +51,153 @@ def bits_str(bits):
...
-            zdb, allzero_tag = zdb.split(",")
-
-        if "|" in zdb:
-            a, b = zdb.split("|")
-            a = a.split()
-            b = b.split()
-        else:
-            a = zdb.split()
-            b = a
+        Zero db format examples:
```
mithro: So, I know you didn't come up with this, but I do find this syntax quite confusing....

utils/dbfixup.py
```diff
@@ -51,57 +51,153 @@ def bits_str(bits):
...
 
-        bitmatch = False
-        for bit in a:
-            if bit in bits:
-                bitmatch = True
+        Ex: 01_02 04_05
```
mithro: Doesn't this mean `01_02` and `04_05` part of a group even if neither of these bits are discovered to be connected?

utils/dbfixup.py
```diff
@@ -51,57 +51,153 @@ def bits_str(bits):
...
-            continue
+        Ex: 01_02 04_05|07_08 10_11
+        If any bits from the first group occur,
+        default bits in the second group to zero
+
+        Ex: 01_02 04_05,ALL_ZERO
+        ALL_ZERO is an enum that is part of the group but is all 0
+        It must have 0 candidates
+
+        Ex: CLB.SLICE_X0.CLKINV ^ CLB.SLICE_X0.NOCLKINV
```
mithro: I think this would be a more correct syntax?
```
Ex: ^CLB.SLICE_X0.CLKINV,NOCLKINV
```

litghost: No?

utils/dbfixup.py
litghost: Would you like to be more specific about what is confusing, it doesn't seem to bad.  This comment is not actionable, and I'm not planning on changing every dbf file over a vague comment.

utils/dbfixup.py
litghost: Your question doesn't make sense, rephrase?

mithro: If I understand things correctly, then when starting with the following output;
```
01_02 BLAH.A
04_05 BLAH.B
```

----

Then with `.dbf` file with the following contents;
```
01_02 04_05
```
The output will make sure that `BLAH.A` and `BLAH.B` are used together as a group and can never be set at the same time. IE The actual bit pattern should be,
```
 01_02 !04_05 BLAH.A
!01_02  04_05 BLAH.B
```

----

Then with a `.dbf` file with the following contents;
```
01_02 04_05,ALL_ZERO
```

You get the following bit patterns,
```
 01_02 !04_05 BLAH.A
!01_02  04_05 BLAH.B
!01_02 !04_05 BLAH.ALL_ZERO
```

----

Then with a `.dbf` file with the following contents;
```
01_02 04_05|07_08 10_11
```

You get the following bit patterns,
```
 01_02 !04_05 !07_08 !10_11 BLAH.A
!01_02  04_05 !07_08 !10_11 BLAH.B
!01_02 !04_05  07_08 !10_11 ????
!01_02 !04_05 !07_08  10_11 ????
```

----

Is my understanding incorrect?

litghost: > Then with a `.dbf` file with the following contents;
> 
> ```
> 01_02 04_05|07_08 10_11
> ```
>
>  01_02 !04_05 !07_08 !10_11 BLAH.A
> !01_02  04_05 !07_08 !10_11 BLAH.B
> !01_02 !04_05  07_08 !10_11 ????
> !01_02 !04_05 !07_08  10_11 ????

No, the result is:
```
01_02 BLAH.A
04_05 BLAH.B
```

litghost:
Red CI is due to 044 failing, re-running for now.

utils/dbfixup.py
mithro: What about the others?

litghost: You mean ``07_08`` and ``10_11``?  Nothing special?

HackerFoo:
Is there an explanation of the syntax change? Is it:
```
FOO.A FOO.B 01_02 03_04
```

to mean "if `FOO.A` and `FOO.B`, then `01_02` and `03_04`"?

If so, maybe we could add an arrow to match a common notation:

```
FOO.A FOO.B -> 01_02 03_04

# could also be written
FOO.A FOO.B -> 01_02
FOO.A FOO.B -> 03_04
```

If it's `01_02` or `03_04`, we might need a SAT solver.

PR #908, opened 2019-06-25
007-timing: added missing aliases for bram timing

acomodi:

This PR is to add missing aliases to the `pin_alias_map.json`. The reason is that some timing definitions were missing in the bram sdf files because some pins/ports were skipped as the `tim2json.py` was not able to find the timing relative to those pins/ports


acomodi:
Merging as CI went green.

PR #916, opened 2019-06-27
Minitests for SRLs

mkurc-ant:
This PR adds several minitests for SRL32 (for now). Helps to identify which FASM features and in which combination are required to make SRLs work.

Each design can be built using Yosys+Vivado flow. After a build `bit2fasm` and `segprint` scripts are invoked.


litghost:
Can you test if LUT's can be packed in with the SRL's?

mkurc-ant:
@litghost Yes, they can. I added tests which show that.

PR #933, opened 2019-07-08
Fuzzer for DOUTMUX.MC31 and DFFMUX.MC31

mkurc-ant:
Modified fuzzer `015-clb-nffmux` and `016-clb-noutmux` to fuzz bits for `DFFMUX.MC31` and `DOUTMUX.MC31`

### DIFF
--- a/fuzzers/015-clb-nffmux/Makefile
+++ b/fuzzers/015-clb-nffmux/Makefile
@@ -1,4 +1,8 @@
-N := 3
-CLB_DBFIXUP=Y
-include ../clb.mk
+# Specimen count
+N := 5
+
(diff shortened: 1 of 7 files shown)


litghost:
PR is closed?

mkurc-ant:
@litghost The problem with the PR is that 016 outputs feature tags both for SLICEL and SLICEM.  The MC31 stuff is only for SLICEM. I closed this one and started working on modifying 018 which deals with RAM (and SRL) stuff for SLICEM. Once I finish I'll open a new one.

mkurc-ant:
After some work I realized that it would be better to reuse existing fuzzers and modify their makefiles so that features for SLICEM do not end up in SLICEL db files.

Issue #929, opened 2019-07-05
IOI3_INTER pip bits unknown


acomodi:
While investigating on https://github.com/SymbiFlow/prjxray/issues/893, I discovered that there are unknown bits related to the `IOI3_INTER` switch box (belonging to the `IOI3` tile).

To get this information I have been instantiating simple OBUFs to connect a GND constant of an INT_L tie-off to the output PAD.
By inspecting the design I have noticed that the only sites/switch_boxes that were used were the IOB, OLOGIC (for which there already are bits documented), INT_L and IOI3_INTER.

I have then progressively instantiated the same design on other above locations and diffed the output of `bit2fasm`, obtaining the following differences:
```
$diff top-obuf1.fasm top-obuf2.fasm 
1,2c1,2
< INT_L_X0Y1.GFAN0.GND_WIRE
< INT_L_X0Y1.IMUX_L34.GFAN0
---
> INT_L_X0Y3.GFAN0.GND_WIRE
> INT_L_X0Y3.IMUX_L34.GFAN0
4,11c4,11
< LIOB33_X0Y1.IOB_Y0.ODDR.DDR_CLK_EDGE.OPPOSITE_EDGE
< LIOB33_X0Y1.IOB_Y0.PULLTYPE.PULLDOWN
< LIOB33_X0Y1.IOB_Y0.SLEW.FAST
< LIOB33_X0Y1.IOB_Y1.LVCMOS18.DRIVE.I12_I8
< LIOB33_X0Y1.IOB_Y1.ODDR.DDR_CLK_EDGE.OPPOSITE_EDGE
< LIOB33_X0Y1.IOB_Y1.OSERDESE.DATA_RATE_TQ.BUF
< LIOB33_X0Y1.IOB_Y1.PULLTYPE.NONE
< LIOB33_X0Y1.IOB_Y1.SLEW.SLOW
---
> LIOB33_X0Y3.IOB_Y0.ODDR.DDR_CLK_EDGE.OPPOSITE_EDGE
> LIOB33_X0Y3.IOB_Y0.PULLTYPE.PULLDOWN
> LIOB33_X0Y3.IOB_Y0.SLEW.FAST
> LIOB33_X0Y3.IOB_Y1.LVCMOS18.DRIVE.I12_I8
> LIOB33_X0Y3.IOB_Y1.ODDR.DDR_CLK_EDGE.OPPOSITE_EDGE
[...]

PR #934, opened 2019-07-08
IOI3_INTER pips fuzzer

acomodi:

This PR is to solve https://github.com/SymbiFlow/prjxray/issues/929.
This is an initial attempt. It creates designs which use the global clock network, necessary to enable and disable the unknown bits relative to the pips.

`top.py` instantiates 16 BUFGs which are randomly used by the I/OLOGIC, I/OSERDES and IDELAYE2 sites.

This still need work as some of the pips are `pseudo-pips` and need to be excluded by the computation. Another thing is that there are other pips that are not relative to the clocking network and need to be addressed in a different way.


acomodi:
@litghost I think that kokoro failed for other reasons as the fuzzer is not active. CI instead failed for formatting reasons.

Anyway regarding the clock sources, I still need to understand how those pips are activated or not and probably you are right, I need to instantiate different clock sources to get to use all of them. 
As next step I will proceed with making the fuzzer more robust and I will analyze as well the clock connections.

acomodi:
@litghost Sure, I'll solve the issue and push a new version.
Anyhow, to give a background on this fuzzer, I have bumped in some issues that led me to use the following strategies:

- Use only global clock sources (for now), as I have seen that litex uses only those pips. This because, using all of the clock sources resulted in the fuzzer not being able to find many of the desired bits.
- Re-routing through tcl using the pips in the `todo.txt` list. This to increase the chances of getting the right solutions.

Right now only the `OLOGICE3` clock sources pips are enabled in the fuzzer as, some of the pips belonging to the `ILOGICE3` result in having only one configuration bit, which generates a conflict in a subsequent iteration of the fuzzer.

There is also an issue for which I cannot get all the pips even `OLOGICE3`-related which I still don't understand the reason, as there is a great abundance, after a few iterations, of tags with both zeros and ones.

acomodi:
I have an issue for which, after some iterations, the fuzzer ends up finding the same bits over and over again, even though the specimens contain very diverse tags combinations of 1s and 0s.
What I get as a result is the following:
```
IOI3.IOI_OCLK_0.IOI_LEAF_GCLK2 29_94 30_94
IOI3.IOI_OCLK_0.IOI_LEAF_GCLK3 29_96 30_96
IOI3.IOI_OCLK_1.IOI_LEAF_GCLK2 28_33 31_33
IOI3.IOI_OCLK_1.IOI_LEAF_GCLK3 28_31 31_31
IOI3.IOI_OLOGIC0_CLKDIV.IOI_LEAF_GCLK2 28_123 31_123
IOI3.IOI_OLOGIC0_CLKDIV.IOI_LEAF_GCLK3 28_125 31_125
IOI3.IOI_OLOGIC1_CLKDIV.IOI_LEAF_GCLK2 29_04 30_04
IOI3.IOI_OLOGIC1_CLKDIV.IOI_LEAF_GCLK3 29_02 30_02
```

litghost:
> I have an issue for which, after some iterations, the fuzzer ends up finding the same bits over and over again, even though the specimens contain very diverse tags combinations of 1s and 0s.
> What I get as a result is the following:
> 
> ```
> IOI3.IOI_OCLK_0.IOI_LEAF_GCLK2 29_94 30_94
> IOI3.IOI_OCLK_0.IOI_LEAF_GCLK3 29_96 30_96
> IOI3.IOI_OCLK_1.IOI_LEAF_GCLK2 28_33 31_33
> IOI3.IOI_OCLK_1.IOI_LEAF_GCLK3 28_31 31_31
> IOI3.IOI_OLOGIC0_CLKDIV.IOI_LEAF_GCLK2 28_123 31_123
> IOI3.IOI_OLOGIC0_CLKDIV.IOI_LEAF_GCLK3 28_125 31_125
> IOI3.IOI_OLOGIC1_CLKDIV.IOI_LEAF_GCLK2 29_04 30_04
> IOI3.IOI_OLOGIC1_CLKDIV.IOI_LEAF_GCLK3 29_02 30_02
> ```

Sorry, what is the problem?  Those bits seem reasonable?

acomodi:
@litghost So, the problem is that, while solving the pips in the todo list, after some iterations, the fuzzer hangs and does not solve the remaining pips and all that it outputs are the bits that I have pasted.

I think that probably I should randomize also the I/OSERDESes parameters to get more data.

litghost:
> @litghost So, the problem is that, while solving the pips in the todo list, after some iterations, the fuzzer hangs and does not solve the remaining pips and all that it outputs are the bits that I have pasted previously.
> 
> I think that probably I should randomize also the I/OSERDESes parameters to get more data.

Oh, sure.  Are the pips left in the TODO required (or believed to be required) for LiteX?

litghost:
I've refactored 037, and it can capture all of the non-inverted clock signals. I suspect that this might have to do with the fact that the inverted (B) signals are unused for SDR modes?  Unclear at this time.  Take a look and see if you can figure out what is going with the inverted clock pips.

litghost:
I've pushed another commit to https://github.com/SymbiFlow/prjxray/compare/master...litghost:ioi3-pips?expand=1, I've been able to isolate the CLKB pips.

The only PIPs left to find are related to the OCLKM pips, which likely need similiar treatment at the CLKB pips.

acomodi:
@litghost Great, thanks. The fuzzer now looks cleaner. I have cherry-picked your changes and fixed/cleaned some minor things.

I will further modify it to get the OCLK-related pips (which shouldn't be used by litex).

One question about segmatch: what does the `-c -1` allow to do? Probably that was one of the reasons I couldn't get any bits when using multiple clock sources because the same bit was used in many different pips.

acomodi:
@litghost, there is an issue with `random.choices` which is not supported in Python versiones `< 3.6`.

Would it be better to change the CI environment to use 3.6 or the clock random choice mechanism?

mithro:
conda environment could be upgrade to 3.7, shall we just use that?

fuzzers/037-iob-pips/Makefile
```diff
@@ -0,0 +1,48 @@
...
+	cp ${XRAY_DATABASE_DIR}/${XRAY_DATABASE}/segbits*.db build/database/${XRAY_DATABASE}
+	XRAY_DATABASE_DIR=${FUZDIR}/build/database ${XRAY_MERGEDB} ioi3_l build/segbits_ioi3_x.db
+	XRAY_DATABASE_DIR=${FUZDIR}/build/database ${XRAY_MERGEDB} ioi3_r build/segbits_ioi3_x.db
+
+build/cmt_regions.csv: output_cmt.tcl
+	mkdir -p build
+	cd build/ && ${XRAY_VIVADO} -mode batch -source ${FUZDIR}/output_cmt.tcl
+
+pushdb: database
+	${XRAY_MERGEDB} ioi3_l build/segbits_ioi3_x.db
```
tmichalak: The segbits file needs to contain the tile_type in its name, in this case it's RIOI3 or LIOI3. Otherwise the database will not get populated correctly and the bits will not be taken into account by bit2fasm.

fuzzers/037-iob-pips/Makefile
```diff
@@ -0,0 +1,48 @@
...
+	${XRAY_MERGEDB} ioi3_r build/segbits_ioi3_x.db
```
tmichalak: Same as above

fuzzers/037-iob-pips/Makefile
acomodi: Ok, to do that I will need to modify `segmaker.py` to allow having a different convention from `<tile>_[lr]`

litghost:
> @litghost, there is an issue with `random.choices` which is not supported in Python versiones `< 3.6`.
> 
> Would it be better to change the CI environment to use 3.6 or the clock random choice mechanism?

Sorry for the confusion, ``random.sample`` is more than good enough, so lets leave the python version alone.

litghost:
> One question about segmatch: what does the `-c -1` allow to do? Probably that was one of the reasons I couldn't get any bits when using multiple clock sources because the same bit was used in many different pips.

It removes the "max bits" limit.  Given that all solutions appear to be 3 bit solutions, we should be able to replace that again with `-c 3`.

litghost:
FYI looks like ``IOI_OLOGIC[01]_CLKDIV`` might need more than 3 bits, most of the other solutions do not.

acomodi:
@litghost In this case I am thinking we could split up the fuzzer in two (e.g. `037a` and `037b`) to use different `-c` tuning

fuzzers/037-iob-pips/Makefile
acomodi: Done.

fuzzers/037-iob-pips/Makefile
acomodi: Done.

acomodi:
I could not yet get the bits relative to the `OCLKM` apart from the ones related to IMUXes as sources.

`OCLKM_[01]` are connected to the OCLKB wire of the ILOGICE3. They are currently not used in Litex, even though they should be solved.

I suppose that it would be better to focus on the currently unknown bits in that still need to be solved and postpone this for later.
I believe the issue is that even though OCLKM pips are stimulated, vivado does not use them as OCLK and OCLKB are active with specific configurations of the IDDR/ISERDES. I still couldn't find the correct configurations.

I have also excluded the following pips (which I have double checked and are not used in litex):

- `IOI3.IOI_OLOGIC[01]_CLKB.IOI_OCLK_[01]`: This is interesting, because it seems that CLKB for OLOGIC is never activated in any design (as there is no port when instantiating ODDR/OSERDES related to CLKB)
- All `PHASER` related pips
- `IOI_ILOGIC0_CLK.IOI_IMUX22_1`: this pip is used very rarely and gives unstable results. I will try to give the right stimuli, but for now is removed from the todo list.

fuzzers/int_maketodo.py
```diff
@@ -209,12 +209,17 @@ def run(
...
 
-        if side != "":
-            side = "_" + side
+        if side == "xl":
+            filename = "l{}".format(pip_type)
+        elif side == "xr":
+            filename = "r{}".format(pip_type)
+
+        if side == "l" or side == "r":
+            filename = "{}_{}".format(pip_type, side)
```
litghost: This change breaks fuzzer 047

fuzzers/037-iob-pips/generate.py
```diff
@@ -0,0 +1,104 @@
+#!/usr/bin/env python3
+
+from prjxray.segmaker import Segmaker
+import os, re
+import os.path
+
+
+def bitfilter(frame, word):
+    if frame < 28:
```
litghost: I believe this is too restrictive.  When looking into remaining bits, I believe this should be "if frame > 26".  Try it out.

fuzzers/int_maketodo.py
acomodi: Done

fuzzers/037-iob-pips/generate.py
acomodi: Ok, you mean `if frame < 26` I imagine, right?

litghost: Yes

fuzzers/int_maketodo.py
```diff
@@ -209,12 +209,20 @@ def run(
...
+
+        if side == "":
```
litghost: Can we make this an if/elif/else tree, with the else an assert so it fails at the end of the if/else tree?  The previous error was "filename var does not exist" type error.

fuzzers/000-init-db/Makefile
```diff
@@ -22,6 +22,8 @@ SEGBITS=\
       dsp \
       hclk \
       int \
+      ioi \
```
litghost: This is not longer correct, and lioi3 and rioi3 files do not get initialized to empty.  Need to update.

fuzzers/037-iob-pips/generate.py
acomodi: With that I get the following bits for CLKB:
```
IOI3.IOI_ILOGIC0_CLKB.IOI_IMUX20_1 26_71 27_70 27_98 30_80 31_67 31_77
IOI3.IOI_ILOGIC0_CLKB.IOI_IOCLK0 26_71 27_70 27_98 30_76 30_78 31_67
IOI3.IOI_ILOGIC0_CLKB.IOI_IOCLK1 26_71 27_70 27_98 30_76 30_80 31_67
IOI3.IOI_ILOGIC0_CLKB.IOI_IOCLK2 26_71 27_70 27_98 31_67 31_77 31_79
IOI3.IOI_ILOGIC0_CLKB.IOI_IOCLK3 26_71 27_70 27_98 31_67 31_77 31_81
IOI3.IOI_ILOGIC0_CLKB.IOI_LEAF_GCLK0 26_71 27_70 27_98 30_74 31_67 31_79
IOI3.IOI_ILOGIC0_CLKB.IOI_LEAF_GCLK1 26_71 27_70 27_98 30_74 31_67 31_81
IOI3.IOI_ILOGIC0_CLKB.IOI_LEAF_GCLK2 26_71 27_70 27_98 30_74 30_78 31_67
IOI3.IOI_ILOGIC0_CLKB.IOI_LEAF_GCLK3 26_71 27_70 27_98 30_74 30_80 31_67
IOI3.IOI_ILOGIC0_CLKB.IOI_LEAF_GCLK4 26_71 27_70 27_98 31_67 31_75 31_79
IOI3.IOI_ILOGIC0_CLKB.IOI_LEAF_GCLK5 26_71 27_70 27_98 31_67 31_75 31_81
IOI3.IOI_ILOGIC0_CLKB.IOI_RCLK_FORIO0 26_71 27_70 27_98 30_78 31_67 31_75
IOI3.IOI_ILOGIC0_CLKB.IOI_RCLK_FORIO1 26_71 27_70 27_98 30_80 31_67 31_75
IOI3.IOI_ILOGIC0_CLKB.IOI_RCLK_FORIO2 26_71 27_70 27_98 30_76 31_67 31_79
IOI3.IOI_ILOGIC0_CLKB.IOI_RCLK_FORIO3 26_71 27_70 27_98 30_76 31_67 31_81
```

Where all the 26 and 27 frame address bits are the same and do not change for all of them. I am unsure whether this is correct.

litghost: Ya, I'm wondering if that is correct.  I guess it could be ILOGIC bits that are correlated with the use of CLKB.  Maybe an improvement for the ILOGIC fuzzer?

acomodi: I will take a look into that, for now I restored the frame check to be `< 28`

fuzzers/000-init-db/Makefile
acomodi: Done

fuzzers/int_maketodo.py
acomodi: Done

fuzzers/037-iob-pips/generate.py
acomodi: So, if I allow ILOGIC fuzzer not to generate CLKB, I get a `ISERDESE2.USE` additional bit which is `27_70` for bottom ilogic and `26_57` for top ilogic.

`27_98` is instead related to `IOB33.IOB_Y0.IFF.DDR_CLK_EDGE.OPPOSITE_EDGE` and is produced in the ILOGIC fuzzer. I am quite confused on why they do show up only in the CLKB related pips though.

I think that, at least, the 26 and 27 frame address bits are not related to IOI3 pips.

acomodi:
As suspected some of the IOI3 pips do end up conflicting with `ISERDES.IN_USE`.
IMO, ISERDES needs to be updated before merging this PR.

```
Checking artix7 database
============================
Checking individual .db...
segbits_*.db: 33 okay
mask_*.db: 38 okay
Checking aggregate dir...
ERROR: 6 collisions
  0040001c_053_03: had LIOI3_X0Y25.LIOI3.IOI_ILOGIC0_CLK.IOI_RCLK_FORIO1, got LIOB33_X0Y25.LIOB33.IOB_Y0.IDDR_OR_ISERDES.IN_USE
  0040001c_053_19: had LIOI3_X0Y25.LIOI3.IOI_OCLK_0.IOI_IOCLK0, got LIOB33_X0Y25.LIOB33.IOB_Y0.ISERDES.IN_USE
  0040001d_052_12: had LIOI3_X0Y25.LIOI3.IOI_OCLK_1.IOI_LEAF_GCLK0, got LIOB33_X0Y25.LIOB33.IOB_Y1.ISERDES.IN_USE
  0040001d_052_28: had LIOI3_X0Y25.LIOI3.IOI_ILOGIC1_CLK.IOI_LEAF_GCLK5, got LIOB33_X0Y25.LIOB33.IOB_Y1.IDDR_OR_ISERDES.IN_USE
  0040001e_052_28: had LIOI3_X0Y25.LIOI3.IOI_ILOGIC1_CLKB.IOI_LEAF_GCLK5, got LIOB33_X0Y25.LIOB33.IOB_Y1.IDDR_OR_ISERDES.IN_USE
  0040001f_053_03: had LIOI3_X0Y25.LIOI3.IOI_ILOGIC0_CLKB.IOI_RCLK_FORIO0, got LIOB33_X0Y25.LIOB33.IOB_Y0.IDDR_OR_ISERDES.IN_USE
  File "utils/checkdb.py", line 180, in <module>
    main()
  File "utils/checkdb.py", line 176, in main
    run(args.db_root, verbose=args.verbose)
  File "utils/checkdb.py", line 163, in run
    check_tile_overlap(db, verbose=verbose)
  File "utils/checkdb.py", line 139, in check_tile_overlap
    raise ValueError("%s collisions" % len(collisions))
ValueError: 6 collisions
Makefile:109: recipe for target 'db-check-artix7' failed
make: *** [db-check-artix7] Error 1
```

litghost:
ISERDES update in #977

litghost:
With #977 merged, hopefully no more bit conflicts.  Re-running CI.

PR #935, opened 2019-07-08
Expand ILOGIC fuzzer to document additional ISERDES bits.

litghost:


fuzzers/035-iob-ilogic/generate.py
```diff
@@ -4,6 +4,16 @@
 from prjxray import verilog
 import json
 
+# Set to true to enable additional tags useful for tracing bit toggles.
+DEBUG_FUZZER = False
+
+
+def bitfilter(frame, word):
```
tmichalak: I always wonder where this information comes from other than that they keep the bits from being solved nicely.

fuzzers/035-iob-ilogic/generate.py
```diff
@@ -158,31 +178,25 @@ def main():
...
 
-    segmk.compile()
+            if DEBUG_FUZZER:
+                for k in d:
+                    segmk.add_site_tag(
+                        site, 'param_' + k + '_' + str(d[k]).replace(
+                            ' ', '').replace('\n', ''), 1)
+
+    segmk.compile(bitfilter=bitfilter)
     segmk.write(allow_empty=True)
```
tmichalak: Looking into the code `allow_empty` lifts the assertion when there are no segment bits for a tag. Why is it needed here?

litghost: It is possible that some specimens will not generate any segbits, but otherwise has no affect.  Solution should be returned.

fuzzers/035-iob-ilogic/generate.py
litghost: You comment is about right.  The filter in this case is removing bits that are potentially correlated, but not believed to be directly part of the solution.  Concretely, INT_[LR] bits are in the (0, 25), but there are other bits that need to be filtered.

PR #946, opened 2019-07-15
Fuzzer for IDELAY

mkurc-ant:
This PR adds a separate fuzzer for IDELAY blocks. Reduces number of unknown bits for the LiteX design by 45. It also adds a modification to the segmaker to support sites with differing Y location.

The solution is written into the database for [LR]IOI3.IDELAY sites instead of for IOB tiles.


mkurc-ant:
@tmichalak Ok, So I removed all things related to the my proposed solver. Also fixed the bug that showed in the CI build.

fuzzers/035a-iob-idelay/top.py
```diff
@@ -0,0 +1,229 @@
...
+import json
+
+from prjxray import util
+from prjxray.db import Database
+
+# =============================================================================
+
+
+def get_loc(name):
+    m = re.match("^\S+_X([0-9]+)Y([0-9]+)$", name)
```
litghost: I believe ``get_loc`` can be replaced by a call to [util.create_xy_fun](https://github.com/SymbiFlow/prjxray/blob/master/fuzzers/040-clk-hrow-config/top.py#L11)

fuzzers/035a-iob-idelay/top.py
```diff
@@ -0,0 +1,229 @@
...
+
+    tile_list = []
+    for tile_name in sorted(grid.tiles()):
+        if "IOB33" not in tile_name or "SING" in tile_name:
+            continue
+        tile_list.append(tile_name)
+
+    def key(name):
+        x, y = get_loc(name)
+        return y + x * 10000
```
litghost: Why not just return tuple (x, y)?

mkurc-ant: Fixed.

fuzzers/035a-iob-idelay/top.py
mkurc-ant: It can't be used here as it assumes constant prefix but I am searching for LIOB33 and RIOB33 which differ.

litghost: Wouldn't ``util.create_xy_fun('[LR]IOB33')`` work?

mkurc-ant: Right, that works :)

PR #970, opened 2019-07-23
Refactor IOB fuzzer.

litghost:
- Add SSTL135
 - Refactor process_rdb to handle varying SLEW by IOSTANDARD.
 - Remove bits that belong to OLOGIC (and are now documented in #971).

Fixes #965

PR #971, opened 2019-07-23
Add some additional OLOGIC bits.

litghost:
This adds a couple features, but most importantly it documents some that were mistakenly associate with the 030 IOB fuzzer, which was fixed in #970.


mkurc-ant:
Kokoro broken again? Can't view the log...

mithro:
Kokoro is broken because of https://github.com/SymbiFlow/prjxray/issues/972

PR #973, opened 2019-07-25
Add ROI harness for the X1Y0 clock region

HackerFoo:
This adds a simple harness in the X1Y0 clock region (bottom right) where two inputs and two outputs are connected to the closest corresponding LUT.

This is helpful to understand IO.

- tested with roi_inv.v on hardware

Screenshot from 2019-07-24 17-17-58

### DIFF
--- a/minitests/roi_harness/basys3-x1y0-fixed.xdc
+++ b/minitests/roi_harness/basys3-x1y0-fixed.xdc
@@ -1,0 +1,49 @@
+# assign pins
+set_property -dict {PACKAGE_PIN V2 IOSTANDARD LVCMOS33} [get_ports {din[0]}]
+set_property -dict {PACKAGE_PIN W2 IOSTANDARD LVCMOS33} [get_ports {din[1]}]
+set_property -dict {PACKAGE_PIN V3 IOSTANDARD LVCMOS33} [get_ports {dout[0]}]
+set_property -dict {PACKAGE_PIN W3 IOSTANDARD LVCMOS33} [get_ports {dout[1]}]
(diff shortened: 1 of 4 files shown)


minitests/roi_harness/basys3-x1y0-fixed.xdc
```diff
@@ -0,0 +1,44 @@
...
+set_property LOC SLICE_X64Y39 [get_cells {roi/ins[1].lut}]
+
+## outs
+set_property BEL A6LUT [get_cells {roi/outs[0].lut}]
+set_property LOC SLICE_X64Y38 [get_cells {roi/outs[0].lut}]
+set_property BEL A6LUT [get_cells {roi/outs[1].lut}]
+set_property LOC SLICE_X64Y37 [get_cells {roi/outs[1].lut}]
+
+# fix routes
+set_property FIXED_ROUTE { { IOB_IBUF0 RIOI_I0 RIOI_ILOGIC0_D IOI_ILOGIC0_O IOI_LOGIC_OUTS18_1 INT_INTERFACE_LOGIC_OUTS18 WL1BEG_N3 NW2BEG0 IMUX7 CLBLM_M_A1 }  } [get_nets {din_IBUF[0]}]
```
litghost: Why is this required at all?  No IO is required from the ROI at all?

litghost:
> This is helpful to understand IO.

So why make an ROI harness at all?  Why not just make a minitests?

minitests/roi_harness/basys3-x1y0-fixed.xdc
HackerFoo: So that I can test that I haven't broken anything by moving the ROI.

Issue #962, opened 2019-07-23
Bits for base address 0x41580, frame 33 


tmichalak:
Following 233 bits reported as missing for Litex:
<pre>
# In frame 0x004015a1 233 bits were not converted.
{ unknown_bit = "004015a1_80_1", unknown_segment = "0x00401580", unknown_segbit = "33_2561" }
{ unknown_bit = "004015a1_96_1", unknown_segment = "0x00401580", unknown_segbit = "33_3073" }
{ unknown_bit = "004015a1_47_23", unknown_segment = "0x00401580", unknown_segbit = "33_1527" }
{ unknown_bit = "004015a1_0_9", unknown_segment = "0x00401580", unknown_segbit = "33_09" }
{ unknown_bit = "004015a1_16_9", unknown_segment = "0x00401580", unknown_segbit = "33_521" }
{ unknown_bit = "004015a1_0_11", unknown_segment = "0x00401580", unknown_segbit = "33_11" }
{ unknown_bit = "004015a1_16_11", unknown_segment = "0x00401580", unknown_segbit = "33_523" }
{ unknown_bit = "004015a1_32_9", unknown_segment = "0x00401580", unknown_segbit = "33_1033" }
{ unknown_bit = "004015a1_32_11", unknown_segment = "0x00401580", unknown_segbit = "33_1035" }
{ unknown_bit = "004015a1_48_9", unknown_segment = "0x00401580", unknown_segbit = "33_1545" }
{ unknown_bit = "004015a1_48_11", unknown_segment = "0x00401580", unknown_segbit = "33_1547" }
{ unknown_bit = "004015a1_0_17", unknown_segment = "0x00401580", unknown_segbit = "33_17" }
{ unknown_bit = "004015a1_16_17", unknown_segment = "0x00401580", unknown_segbit = "33_529" }
{ unknown_bit = "004015a1_32_17", unknown_segment = "0x00401580", unknown_segbit = "33_1041" }
{ unknown_bit = "004015a1_48_17", unknown_segment = "0x00401580", unknown_segbit = "33_1553" }
{ unknown_bit = "004015a1_80_13", unknown_segment = "0x00401580", unknown_segbit = "33_2573" }
{ unknown_bit = "004015a1_80_22", unknown_segment = "0x00401580", unknown_segbit = "33_2582" }
{ unknown_bit = "004015a1_80_23", unknown_segment = "0x00401580", unknown_segbit = "33_2583" }
[...]

PR #975, opened 2019-07-25
036-ologic: add IN_USE oserdes

acomodi:

This PR adds the `OSERDES.IN_USE` tag.
More info in this issue https://github.com/SymbiFlow/prjxray/issues/962.

### DIFF
--- a/fuzzers/036-iob-ologic/generate.py
+++ b/fuzzers/036-iob-ologic/generate.py
@@ -37,6 +37,8 @@ def main():
 
             handle_data_width(segmk, d)
 
+            segmk.add_site_tag(site, 'OSERDESE.IN_USE', d['use_oserdese2'])
+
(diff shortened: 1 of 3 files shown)


acomodi:
@litghost, I think a similar approach should be used to get the `ISERDES.IN_USE` bits as well, as they most probably do include some `IOI3_INTER` pips given by the presence of clock signals every time an ISERDES is instantiated.

litghost:
> @litghost, I think a similar approach should be used to get the `ISERDES.IN_USE` bits as well, as they most probably do include some `IOI3_INTER` pips given by the presence of clock signals every time an ISERDES is instantiated.

If I remember correctly, the approach of disconnecting clocks does not work.  Even if the clock is disconnected, a GLOBAL_LOGIC[01] is routed instead.  Worth trying though.

PR #976, opened 2019-07-25
Add additional features for IDELAY.

litghost:


litghost:
> What additional features does it solve ?

It solves the IDELAY.IN_USE feature, which required decoupling the IOI3 pips.  #980 solves the same bit.

PR #977, opened 2019-07-25
ISERDES FASM feature improvements

litghost:
- Add non-decimal prefixes to feature parts
 - Remove clocks in some samples to decouple ioi pips.

### DIFF
--- a/fuzzers/035-iob-ilogic/top.py
+++ b/fuzzers/035-iob-ilogic/top.py
@@ -189,6 +189,13 @@ def use_iserdese2(p, luts, connects):
             'clk_BUFG2',
         ))
 
+    clkdiv = random.choice(('0', ))
+
+    if random.randint(0, 1):
+        clknet = '0'
+        clkbnet = '0'
+        oclknet = '0'
+
     print(
         '''
     (* KEEP, DONT_TOUCH, LOC = "{ilogic_loc}" *)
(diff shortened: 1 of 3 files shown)


fuzzers/035-iob-ilogic/top.py
```diff
@@ -189,6 +189,15 @@ def use_iserdese2(p, luts, connects):
             'clk_BUFG2',
         ))
 
+    clkdiv = random.choice(('0', ))
+
+    if random.randint(0, 1):
+        clknet = '0'
+        clkbnet = '0'
+        oclknet = '0'
+        #clkdiv = ''
```
acomodi: Leftover commented line?

acomodi: IMO `generate.py` needs to be modified accordingly not to add the `IS_<clk_type>_INVERTED` tag when the clock is not present. If the clocks are not added, segmatch will not be able to correctly solve the `INVERTED` tags.
E.g. there could show up situations for which the `INVERTED` feature is set to `1`, but the clock is not added.

litghost: This change does not disconnect the clocks, it just changes the pip uses to drive the clock (GCLK_LEAK -> IMUX), which decouples the analysis.  INVERTED bits are still being solved.

acomodi: Ok, got it. I didn't realize that the clocks were set to zero, therefore connected to HARD0 of the INT_L next to it.

PR #980, opened 2019-07-26
035c-idelay-use: fuzzer to get the IN_USE tag for idelays

acomodi:

This PR is needed to document the `IN_USE` bits of the `IDELAYs`.
Being those bits highly correlated to other IDELAYs features, they could not be solved in the idelay original fuzzer.

This small fuzzer instead filters out bits for relative to IDELAYs features to get only the ones relative to the `IN_USE` tag.


acomodi:
@litghost I wonder whether this PR is actually useful given that I have just learned that https://github.com/SymbiFlow/prjxray/pull/976 solves the same bits.

litghost:
> @litghost I wonder whether this PR is actually useful given that I have just learned that #976 solves the same bits.

I'm evaluating that now.  I think I'd rather just merge #976, as the change that decouples the IOI3 pips is probably a good change to have either way.

litghost:
Duplicate of #976

PR #983, opened 2019-07-26
Add initial DIFF_ support to IOB fuzzer.

litghost:
- Add IN_DIFF and OUT_DIFF features.
- Add IN_TERM features.

This appears to solve all remaining LiteX IOB bits in frames 38/39.

### DIFF
--- a/fuzzers/030-iob/top.py
+++ b/fuzzers/030-iob/top.py
@@ -62,6 +62,19 @@ def run():
         "SSTL135": ["DIFF_SSTL135"],
     }
 
+    IN_TERM_ALLOWED = [
+        'SSTL15',
+        'SSTL15_R',
+        'SSTL18',
+        'SSTL18_R',
+        'SSTL135',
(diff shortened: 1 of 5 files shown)


fuzzers/030-iob/top.py
```diff
@@ -93,70 +115,151 @@ def run():
...
+                p['pad_wire'] = 'do[{}]'.format(o_idx)
+                p['iwire'] = luts.get_next_output_net()
+                if drives is not None:
+                    p['DRIVE'] = random.choice(drives)
+                else:
+                    p['DRIVE'] = None
+                p['SLEW'] = verilog.quote(random.choice(slews))
+
+                o_idx += 1
+            elif p['type'] == 'OBUFDS':
```
acomodi: Shouldn't there be also `OBUFT`? Maybe it is not needed, as there already is `OBUFTDS` providing the tristate information.

PR #984, opened 2019-07-26
Create script for generating remaining LiteX bit report.

litghost:
This report is fairly fragile, but works well enough for the remaining
LiteX bits.


acomodi:
@litghost I believe this could be merged even if CI was red. The issue is with a slight instability of fuzzer `047-hclk-ioi` which generates duplicate tags. Probably stricter segmatch options can solve the issue.

PR #993, opened 2019-07-30
Add ISERDESE2.DATA_WIDTH's 2-8.

litghost:
10 and 14 require ISERDESE2 MASTER/SLAVE which are not currently in use.


fuzzers/035-iob-ilogic/generate.py
```diff
@@ -19,13 +19,43 @@ def handle_data_width(segmk, d):
     if 'DATA_WIDTH' not in d:
         return
 
-    if d['DATA_RATE'] == 'DDR':
-        return
-
-    for opt in [2, 3, 4, 5, 6, 7, 8, 10, 14]:
+    # It appears several widths end up with the same bitstream pattern.
+    # This groups those widths together for documentation.
```
kgugala: this looks like some kind of optimization in bits usage. Since 5 and 7 widths cannot be used in DDR mode, the SDR bit is most probably used to set the width configuration in the device

litghost: I like that theory.  I originally documented this feature as "W_ODD", which would totally 100% correlate with SDR modes.

PR #1004, opened 2019-07-31
Solve OCLKM pips in IOI3

litghost:
- Also relaxes pip list filtering to capture additional bits.


litghost:
> LGTM
> 
> I think the following PIPs are the last missing in Litex:
> 
> ```
> RIOI3.IOI_OLOGIC0_CLKB.IOI_OCLK_0
> RIOI3.IOI_OLOGIC1_CLKB.IOI_OCLK_1
> RIOI3.IOI_OLOGIC1_CLKB.IOI_OCLKM_1
> RIOI3.IOI_OLOGIC0_CLKB.IOI_OCLKM_0
> ```
> 
> Does this change solve them?

I think that is still the gap.

PR #1012, opened 2019-08-05
Segbits utils

mkurc-ant:
This PR holds two useful utility scripts to be used for segbits files:

`maskview.py`
Shows bits read from single/multiple .db/.rdb files on a 2D plot (frame vs. bit). Can also show conflicts (same bit defined in multiple files). See docstring for more details.

`segview.py`
Shows bit patterns corresponding to tags in .db/.rdb files. Allows to visually determine feature groups etc. See docstring for more details.

### DIFF
--- a/utils/segview.py
+++ b/utils/segview.py
@@ -25,14 +25,14 @@ def tagmap(tag):
     tag = tag.replace("CLBLM_L", "CLB")
     tag = tag.replace("CLBLM_M", "CLB")
 
-    tag = tag.replace("SLICEL",  "SLICE")
-    tag = tag.replace("SLICEM",  "SLICE")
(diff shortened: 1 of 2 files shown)


mkurc-ant:
Travis failed due to some internet connectivity issues...

utils/segview.py
```diff
@@ -0,0 +1,228 @@
...
+    with open(file_name, "r") as fp:
+        lines = fp.readlines()
+
+        # Parse lines
+        for line in lines:
+            line = line.strip()
+            fields = line.split()
+
+            if len(fields) < 2:
+                print("Malformed line: '%s'" % line)
```
litghost: Should this be a terminal error?
If not, should this be output on stderr so it doesn't get lost in the output?

PR #1025, opened 2019-08-16
WIP: Add bitstream tools for Spartan6

tmichalak:
This PR aims at adding Spartan6 support to existing prjxray bitstream tools (bitread, patch, frames2bit).
It has been tested on the Mimas V2 Spartan6 board (https://numato.com/product/mimas-v2-spartan-6-fpga-development-board-with-ddr-sdram) which features the Spartan XC6SLX9 CSG324 FPGA.

The main differences between series7 (according to UG380) that were taken into account were the word length, which is 16 bit instead of 32, and frame word count which is 65 instead of 101.
The configuration registers and their corresponding addresses have been adjusted.
The initialization, startup and finalization sequence has been derived from an original bitstream.

The code in the current state builds separate part specific tools, i.e. (bitread, xc7patch, xc7frames2bit and their Spartan6 counterparts).
The next stage is to merge them into part independent applications.

### DIFF
--- a/lib/include/prjxray/xilinx/spartan6/bitstream_reader.h
+++ b/lib/include/prjxray/xilinx/spartan6/bitstream_reader.h
@@ -1,0 +1,94 @@
+#ifndef PRJXRAY_LIB_XILINX_XC7SERIES_BITSTREAM_READER_H
+#define PRJXRAY_LIB_XILINX_XC7SERIES_BITSTREAM_READER_H
+
+#include <algorithm>
+#include <iostream>
+#include <memory>
+#include <vector>
+
+#include <absl/types/optional.h>
+#include <absl/types/span.h>
+
+#include <prjxray/big_endian_span.h>
+#include <prjxray/xilinx/spartan6/configuration_packet.h>
+
+namespace prjxray {
+namespace xilinx {
+namespace spartan6 {
+
+class BitstreamReader {
+       public:
+	using value_type = ConfigurationPacket;
+
+	class iterator
+	    : public std::iterator<std::input_iterator_tag, value_type> {
+	       public:
+		iterator& operator++();
+
+		bool operator==(const iterator& other) const;
+		bool operator!=(const iterator& other) const;
+
+		const value_type& operator*() const;
+		const value_type* operator->() const;
+
+	       protected:
+		explicit iterator(absl::Span<uint32_t> words);
+
+	       private:
+		friend BitstreamReader;
+
+		ConfigurationPacket::ParseResult parse_result_;
+		absl::Span<uint32_t> words_;
+	};
+
+	// Construct a reader from a collection of 16-bit, big-endian words.
+	// Assumes that any sync word has already been removed.
+	BitstreamReader(std::vector<uint32_t>&& words);
+
+	// Construct a `BitstreamReader` from a Container of bytes.
+	// Any bytes preceding an initial sync word are ignored.
+	template <typename T>
+	static absl::optional<BitstreamReader> InitWithBytes(T bitstream);
+
+	const std::vector<uint32_t>& words() { return words_; };
+
+	// Returns an iterator that yields `ConfigurationPackets`
+	// as read from the bitstream.
+	iterator begin();
+	iterator end();
+
+       private:
+	static std::array<uint8_t, 4> kSyncWord;
+
+	std::vector<uint32_t> words_;
+};
+
+template <typename T>
+absl::optional<BitstreamReader> BitstreamReader::InitWithBytes(T bitstream) {
+	// If this is really a Xilinx 7-Series bitstream, there will be a sync
+	// word somewhere toward the beginning.
+	auto sync_pos = std::search(bitstream.begin(), bitstream.end(),
+	                            kSyncWord.begin(), kSyncWord.end());
+	if (sync_pos == bitstream.end()) {
+		return absl::optional<BitstreamReader>();
+	}
+	sync_pos += kSyncWord.size();
+
+	// Wrap the provided container in a span that strips off the preamble.
+	absl::Span<typename T::value_type> bitstream_span(bitstream);
+	auto config_packets =
+	    bitstream_span.subspan(sync_pos - bitstream.begin());
+
+	// Convert the bytes into 16-bit, big-endian words.
+	auto big_endian_reader = make_big_endian_span<uint16_t>(config_packets);
+	std::vector<uint32_t> words{big_endian_reader.begin(),
+	                            big_endian_reader.end()};
+
+	return BitstreamReader(std::move(words));
+}
+
+}  // namespace spartan6
+}  // namespace xilinx
+}  // namespace prjxray
+
+#endif  // PRJXRAY_LIB_XILINX_XC7SERIES_BITSTREAM_READER_H
--- a/lib/CMakeLists.txt
+++ b/lib/CMakeLists.txt
@@ -21,6 +21,28 @@ add_library(libprjxray
 target_include_directories(libprjxray PUBLIC "include")
 target_link_libraries(libprjxray absl::optional absl::strings absl::span absl::time yaml-cpp)
 
+add_library(libsp6
+	database.cc
+	memory_mapped_file.cc
+	segbits_file_reader.cc
+	xilinx/spartan6/bitstream_reader.cc
+	xilinx/spartan6/bitstream_writer.cc
+	xilinx/spartan6/block_type.cc
+	xilinx/spartan6/configuration_bus.cc
+	xilinx/spartan6/configuration_column.cc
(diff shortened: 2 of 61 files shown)


koriakin:
Here are the differences between Spartan 6 and Series 7 bitstream formats:

1. the obvious one: words are 16-bit instead of 32-bit, there is a different register & command set, and all registers & commands have different numbers even if they are common between devices
2. frame addresses now have two parts (since the registers are 16-bit). The format is:

  - low register: minor frame number (index within column)
  - high register, bits 0-7: major frame number (ie. column)
  - high register, bits 8-11: clock region row index
  - high register, bits 12-15: frame kind

3. Frame kinds are:

  - 0: general config, always has 16×64 + 16 bits (64 bits per CLB row + 16 bits in the middle for HCLK row).  Columns have a variable number of frames depending on their type.
  - 1: BRAM data; I'm not completely sure how long they are (bitgen always emits a full set of them without gaps), but assuming they have the same size as general config results in a reasonable parse.  The geometry is different than general config: there are always 72 frames per column per clock row, 18 frames per blockram (a frame contains data of a single block RAM, as opposed to general config frames, which span the whole column of a clock region)
  - 2: IOB data; there is only one frame with this kind, and it's an extra-special one with size varying between devices.  It has 64 bits per each I/O pad (ie. per IOB* site) plus 384 bits in the middle of every device side (for REG_[LRTB] tiles).  The IOBs and REGs are arranged in clockwise order, starting from lower right corner of the device.  The FLR register has to be written with the size of this frame, in words.

4. For a normal bitstream all frames are uploaded together in a single FDRI write without gaps. A single final extra 16-bit pad word is written after the IOB frame.

5. There is errata in Spartan 6 devices (https://www.xilinx.com/support/answers/39999.html) which requires a special bitstream format when RAMB8BWER primitives are used: first, a normal full FDRI write is performed, covering all frames, but with the RAM kind bits set to 18kbit instead of 9kbit.  Afterwards, the frames containing RAM kind bits are written again, this time with 9kbit size (this is because the BRAM data config circuitry doesn't work correctly when the RAMs are set to 9kbit size). 

6. There are no ECC bits in Spartan 6.

7. There is, however, a post-configuration CRC.  I haven't figured out the exact algorithm yet.  The expected value for this CRC is written to the EXP_SIGN register, and it includes a subset of frame data.  Inclusion of some frames in the CRC can be disabled via special config bits in the HCLK area, which are set when a dynamically-reconfigurable primitive is instantiated in the area (SLICEM with writable LUTs, GTP, and IO primitives).

8. The bitstream CRC is different from Series 7, here's my implementation:

```
class CrcSpartan6:
    CRC_WIDTH = 22
    POLY = 0x409081

    def __init__(self):
        self.state = 0

    def reset(self):
        self.state = 0

    def feed(self, data):
        # data bits 0-15 are register data, bits 16-21 are register address.  Yes, the CRC register
        # is shifted only by 1 bit for every 22 bits of input.
        self.state <<= 1
        if self.state & 1 << self.CRC_WIDTH:
            self.state ^= self.POLY
        self.state ^= data
```

tmichalak:
@koriakin: Thanks a lot for sharing this information with me !
@mithro @litghost: the basic bitread functionality for Spartan6 is done, however, we will need real yaml descriptions for chosen parts. Do we know which Spartan6 parts we want to focus on first?
Does it make sense to focus on the bitstreamer writer part now or do we rather want to merge the spartan6 reader part into series7 bitread?

mithro:
@tmichalak - which do you think is better?

The MimasV2 would be a good first board to target. Followed by the Atlys and Opsis boards.

mithro:
FYI --
 - [ ] The [Numato MimasV2](https://numato.com/product/mimas-v2-spartan-6-fpga-development-board-with-ddr-sdram) is a Spartan XC6SLX9 in CSG324 package.
 - [ ] The [Numato Opsis](https://numato.com/product/numato-opsis-fpga-based-open-video-platform) is a Xilinx Spartan-6 S6LX45T in a FGG488 package.
 - [ ] The [Digilent Atlys](https://store.digilentinc.com/atlys-spartan-6-fpga-trainer-board-limited-time-see-nexys-video/) is a Xilinx Spartan-6 LX45 FPGA, 324-pin BGA package.

tmichalak:
@mithro Forking, i.e. working on the writer part in the separate binary should get us faster to the state where we can check whether we understand and can read/generate the bitstream for Spartan6 correctly, including tests on hardware. Merging the two versions, if we make such decision, is mostly software engineering/refactoring work.
As far as the Spartan6 part is concerned it's a matter of whether we want to go with size (the LX45T is of comparable size with Arty's xc7A35T - 43k logic cells vs 33k logic cells) or affordability (the LX45T boards with Spartan6 cost 4 times more than an Arty board). Not sure which is better for the community.

mithro:
Affordable is the better option.

plaes:
> Affordable is the better option.
There are also these "[repurposed](https://github.com/q3k/chubby75/blob/master/doc/getting_started/getting_started.md)" sub-20€ RV901T led controller boards with Spartan6 XC6LX16.

tmichalak:
@plaes Do you have the schematics of this board? We're you able to program it?

plaes:
> @plaes Do you have the schematics of this board? Were you able to program it?

I only received two of these boards few days ago and I only had time to scrape off the soldermask on the JTAG pins before leaving the office.

Regarding the schematics, there the board has been delayered and scanned, and most (if not all) connector pins have been [reverse-engineered](https://github.com/q3k/chubby75/blob/master/doc/hardware.md) and there's a [migen-based platform description](https://github.com/q3k/chubby75/blob/master/platform.py) with [rgmii](https://github.com/q3k/chubby75/blob/master/rgmii_test.py) / [sdram](https://github.com/q3k/chubby75/blob/master/sdram_test.py) test targets.

mithro:
@tmichalak - FYI @kgugala has a couple of these line driver boards sitting on his desk (I gave him a couple).

lib/include/prjxray/xilinx/spartan6/bitstream_reader.h
```diff
@@ -0,0 +1,94 @@
...
+
+#include <absl/types/optional.h>
+#include <absl/types/span.h>
+
+#include <prjxray/big_endian_span.h>
+#include <prjxray/xilinx/spartan6/configuration_packet.h>
+
+namespace prjxray {
+namespace xilinx {
+namespace xc7series {
```
litghost: Is the idea to leave the name space as xc7series until we merge everything back together?

tmichalak: Not really, I focused on getting bitread to correctly read a Spartan6 bitstream and produce a sensible bits file first. I will focus on refactoring once I am sure this works fine, .i.e get the sp6patch to write a valid bitstream with the data obtained from bitread (since the bitstream reader class is in a common library for both tools). Expect a PR with sp6patch soon.

litghost:
@tmichalak What's left for this PR?  I see it marked as WIP, but I'm not sure why.

tmichalak:
@litghost This PR was created before we made the decision to merge the bitstream tools and was supposed to add Spartan6 specific versions of the tool. I am pretty close to having all of the tools merged, i.e. all architectures will share the same library with some considerable amount of C++ templates. 
Hence I guess this PR should be closed without merging and I will open up a new one for the merge?

fred-freeman:
Sorry for poking this really old PR but this is the last place where I see some of the tooling I was looking for.

I am trying to track down what happened with all of this and getting a bit lost. What was the conclusion of this effort? One thing I see reference to here but is lost in future work is `export XRAY_ISE="${XRAY_UTILS_DIR}/planAhead.sh"`

I'm a bit confused how the fuzzers would work without bitstream generation but then master has tons of files from bitread to libs which include support for spartan6. Is there still work for the parser needed before even bothering with adding fuzzer support?

tmichalak:
@fred-freeman Spartan-6 FPGA Configuration User Guide UG380 defines the low-level bitstream format (BIT header, but-width detection, synchronization word). Packet structure and overall configuration sequence are similar in concept and design to Series 7, but Spartan 6 uses a 16-bit word and has 65 word frames. There are several other differences taken into account in the binary tools which basically let you read an decode the configuration data from the bitstream as well as  patch a bitstream with the configuration. However, the meaning of each bit/word/frame is yet to be determined with the fuzzers.

fred-freeman:
Gotcha I did not realize all of this tooling was just to parse the configuration headers and start to parse the frame data. I saw the references to ISE in code and incorrectly assumed work had also been done to interface ISE with the fuzzers. So https://github.com/f4pga/ideas/issues/10 is still active/needed.

tmichalak:
@fred-freeman Yes, the tooling to decode the bitstream and putting it back is there. However, the task to correlate the bits with the features has not been done yet.

Issue #996, opened 2019-07-30
Create ISERDES minitest


litghost:
A minitest is required for the ISERDES.  The purpose of this minitest is the following:
 - The design should work either a Basys3 or Arty
 - Should be verifiable that it works on hardware (unclear how to do this)

The purpose of this minitest is 3 fold:
 - It should demonstrate that all bits required to use the ISERDESE2 are documented
 - It should be useful in testing that fasm2v supports the ISERDESE2
 - The verilog should be usable by symbiflow-arch-defs to verify ISERDESE2 support

mkurc-ant:
> Should be verifiable that it works on hardware (unclear how to do this)

I have an idea for that: We can make a design with a transmitter part and a receiver part. The transmitter will use ordinary logic and output data on some pins. Then there will be a receiver with ISERDES on other pin(s). And we make a external short between those pins with a wire. Can be done on eg. Basys3

litghost:
I like it.  The PMod port seems like a good candidate here?

mkurc-ant:
Yes, definitely. And a similar thing can be done with OSERDES - transmitt via serdes and receive via logic.

mithro:
@mkurc-ant - It would be good to have a real design in the future too (like the big LiteX Linux design). Something based on [liteiclink](https://github.com/enjoy-digital/liteiclink) would be a great test here.

mkurc-ant:
https://github.com/SymbiFlow/prjxray/pull/1030

dlharmon:
A bidirectional pin using IOBUF can have both a normal logic output and ISERDES attached. When T=0 (output enabled), the ISERDES will see the data output to the pin. This enables you to use any unused IO pin with no hardware jumper required. Same works for OSERDES and normal logic input.

litghost:
> A bidirectional pin using IOBUF can have both a normal logic output and ISERDES attached. When T=0 (output enabled), the ISERDES will see the data output to the pin. This enables you to use any unused IO pin with no hardware jumper required. Same works for OSERDES and normal logic input.

Oooooh, that's cute!  I like it.

PR #1030, opened 2019-08-22
Minitest for ISERDES (networking SDR and DDR)

mkurc-ant:
This is a minitest for operation of ISERDES primitives. It tests networking SDR and DDR modes (see the README.md file).

Due to specific timing requirements of ISERDES primitives the design has to use BUFRs to generate divided clock. Generating this clock using logic led to faulty data reception.


litghost:
@mkurc-ant : Can you try the suggestion in https://github.com/SymbiFlow/prjxray/issues/996#issuecomment-529526113 and see if we can avoid the loopback!

mkurc-ant:
I've been trying it today while creating OSERDES minitest. But I'm still playing with BUFR etc. to make Vivado P&R successfully.

PR #1034, opened 2019-08-26
Change theme to SymbiFlow

mgielda:


docs/conf.py
```diff
@@ -113,7 +113,7 @@
 # The theme to use for HTML and HTML Help pages.  See the documentation for
 # a list of builtin themes.
 #
-html_theme = 'sphinx_rtd_theme'
+html_theme = 'sphinx_symbiflow_theme'
```
mithro: Does this need to be installed somehow?

mgielda: Yes, it's in the requirements.txt

mgielda: Just like sphinx_rtd_theme was. It's a 1:1 replacement

mithro: You need to change the requirements.txt file then?

mgielda: Yup, forgot to commit the requirements file, sorry

PR #1040, opened 2019-09-02
Fixing documentation cross-reference links

acomodi:

This PR solves some issues with cross-references between Markdown files (also between .md and .rst).


mithro:
@acomodi - We should split `docs/markdown_code_symlinks.py` out into its own PyPi module?

README.md
```diff
@@ -130,7 +130,7 @@ do.
 There are also "minitests" which are designs which can be viewed by a human in
 Vivado to better understand how to generate more useful designs.
 
-### [Experiments](experiments)
+### [Experiments](https://github.com/SymbiFlow/prjxray/blob/master/experiments)
```
mithro: Can we just provide a URL to the module which will prefix these bare links with?

IE We are providing the following config;
```python
    app.add_config_value(
        'recommonmark_config', {
            'github_code_repo': 'https://github.com/SymbiFlow/prjxray',
        }, True)
```

docs/db_dev_process/parts.rst
```diff
@@ -1,118 +1,3 @@
-
```
mithro: Out of interest, why is this file moving?

docs/markdown_code_symlinks.py
```diff
@@ -55,9 +66,21 @@ def relative(parent_dir, child_path):
     return os.path.relpath(child_path, start=parent_dir)
 
 
-class MarkdownCodeSymlinks(transform.AutoStructify, object):
+class PrjxrayDomain(Domain):
```
mithro: I think this should probably be a `MarkdownSymlinksDomain`?

docs/markdown_code_symlinks.py
```diff
@@ -1,7 +1,18 @@
 import logging
```
mithro: It would be really good if we could add some tests for this code in some way?

docs/markdown_code_symlinks.py
```diff
@@ -113,65 +157,80 @@ def find_links(cls):
...
+            scheme_known = bool(url_check.scheme)
+
+        is_dest_xref = url_check.fragment and destination[:1] != '#'
+
+        ref_node['refuri'] = destination
+        next_node = ref_node
+
+        # Adds pending-xref wrapper to unsolvable cross-references
+        if (is_dest_xref or not url_check.fragment) and not scheme_known:
+            wrap_node = addnodes.pending_xref(
```
mithro: Do links to headers work?

docs/db_dev_process/parts.rst
acomodi: So, there was a main issue in cross-referencing links in Markdown files that were directed to some repo's directory (e.g. https://github.com/SymbiFlow/prjxray/blob/master/README.md#minitests).
To solve this I have split the `parts.rst` into separate sections (minitests and fuzzers) and moved them into their corresponding folder.
After that I have created a symlink to the `minitests` and `fuzzers` folder in the `code` tree so that the `code2docs` `docs2code` mapping would contain also these cross-references used to solve the pending_xref nodes containing the wrong link to the `minitests` or `fuzzers` directory.

docs/markdown_code_symlinks.py
acomodi: By headers you mean links to other html pages without the use of fragments (e.g. a xref from CONTRIBUTING.md to SUPPORT.md)?
In this case yes, they do work

docs/markdown_code_symlinks.py
acomodi: Agreed, I will think on how to perform these tests.
I suppose that docutils or sphinx itself has some kind of check on whether links are broken.

So a possible test could be to produce a doc build and check that every link references to a valid destination.

acomodi:
> @acomodi - We should split `docs/markdown_code_symlinks.py` out into its own PyPi module?

Yeah, it could be worth doing this as this module could be potentially used in all the Symbiflow repos

docs/markdown_code_symlinks.py
acomodi: Done

README.md
acomodi: I have solved this in a similar way by creating an initializer function in the Domain class (this because I couldn't access the `recommonmark_config` settings from the Domain class itself.

acomodi:
> @acomodi - We should split `docs/markdown_code_symlinks.py` out into its own PyPi module?

I have created a new repo with the `markdown_code_symlinks.py` script here: https://github.com/SymbiFlow/markdown-code-symlinks.
Currently, to import it I have added a line to the requirements.txt linking directly to the repo.

acomodi:
@mithro CI went green

mithro:
@acomodi Is this ready to be merged now?

acomodi:
@mithro I need to go through another check actually as there were some issues with VtR docs generation for which the readthedocs build was failing. I still need to make sure that everything still works after the fixes I have done to let VtR docs correctly build.

acomodi:
@mithro: done, I have checked that ReadTheDocs builds successfully and as soon as CI gets green we can merge.

docs/db_dev_process/minitests/index
```diff
@@ -0,0 +1 @@
+../../../minitests/
```
mithro: What is this?

docs/db_dev_process/parts.rst
mithro: I'm not sure I understand this? The process to generate the link should be;
  1) Lookup src in `code2docs`
  2) Lookup dst in `code2docs`
  3) Calculate the relative locations of src and dst files - use this as the pending_xref value

docs/db_dev_process/minitests/index
acomodi: It is a "fake" symbolic link that is necessary in this particular case (as well as for the fuzzer sections) where the README.md file is absent from the [minitest](https://github.com/SymbiFlow/prjxray/tree/master/minitests) directory.
Without the sym-link the recommonmark package will treat [this link](https://github.com/SymbiFlow/prjxray/#minitests) as an absolute path (as it does not appear in the xref-links, nor in the known url schemas).

It can be removed as soon as a README.md file is created in both the directories.

docs/db_dev_process/parts.rst
acomodi: Partly explained it [here](https://github.com/SymbiFlow/prjxray/pull/1040#discussion_r327719854).
The process is the one that you described, the only issue is that there was no initial mapping in `code2docs` neither in `docs2code` for the minitest and fuzzers directories. To do so there was the need for a fake sym-link that had to be called `index`. 
Therefore I have split the fuzzers and minitests in separate directories.

PR #1046, opened 2019-09-09
fuzzers: 038-cfg: Add always on bit for Zynq

tmichalak:
This PR is a remedy to the current situation in which 3 bits which are reported as unknown (#746) and thus show up in the fasm file keep the harness generation for Zynq from being successful and block the following PR: https://github.com/SymbiFlow/symbiflow-arch-defs/pull/500.

### DIFF
--- a/fuzzers/038-cfg/add_constant_bits.py
+++ b/fuzzers/038-cfg/add_constant_bits.py
@@ -1,0 +1,11 @@
+import sys
+
+constant_bits = {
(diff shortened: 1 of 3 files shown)


fuzzers/038-cfg/add_constant_bits.py
```diff
@@ -0,0 +1,11 @@
+import sys
```
litghost: Please a comment on how we got to this point.

tmichalak: Done

Issue #1050, opened 2019-09-12
db-extras-zynq7 writing files to wrong path


mithro:
https://github.com/SymbiFlow/prjxray/blob/6d21194b569b2e320d7b1c4fe87c8db814a823b1/Makefile#L136-L140

Should be `database/zynq7/harness/` **not** `database/artix7/harness`.

mithro:
```patch
diff --git a/Makefile b/Makefile
index 12be9a41..948fc7f0 100644
--- a/Makefile
+++ b/Makefile
@@ -137,7 +137,7 @@ db-extras-zynq7:
        +source minitests/roi_harness/zybo-swbut.sh && $(MAKE) -C fuzzers part_only
        +source minitests/roi_harness/zybo-swbut.sh && \
                $(MAKE) -C minitests/roi_harness \
-                       HARNESS_DIR=database/artix7/harness/zybo/swbut run
+                       HARNESS_DIR=database/zynq7/harness/zybo/swbut run
 
 db-check:
        @true
```

tmichalak:
@mithro 
Created a PR with a fix (#1051), however I see you have applied the required changes for Zynq7 manually in the database.

tmichalak:
The PR has been merged. The issue has been fixed.

Issue #1049, opened 2019-09-12
Zynq harness not building on Zynq kokoro?


mithro:
Looking at the output of the latest Kokoro run for the Zynq.

https://console.cloud.google.com/storage/browser/symbiflow-prjxray/artifacts/prod/foss-fpga-tools/prjxray/continuous/database/zynq7/271/20190911-091600

If you look at the diff at https://storage.cloud.google.com/symbiflow-prjxray/artifacts/prod/foss-fpga-tools/prjxray/continuous/database/zynq7/271/20190911-091600/database/diff.html it doesn't look like the harness for zynq was actually built?

tmichalak:
@mithro Looks to me that it is caused by #1050. The PR with the fix is in #1051

tmichalak:
The PR has been merged and the issue is gone.

PR #1051, opened 2019-09-12
Fix HARNESS_DIR path for db-extras-zynq7

tmichalak:
This commit fixes issue #1050 and most probably #1049 as well.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -137,7 +137,7 @@ db-extras-zynq7:
 	+source minitests/roi_harness/zybo-swbut.sh && $(MAKE) -C fuzzers part_only
 	+source minitests/roi_harness/zybo-swbut.sh && \
 		$(MAKE) -C minitests/roi_harness \
-			HARNESS_DIR=database/artix7/harness/zybo/swbut run
+			HARNESS_DIR=$(XRAY_DATABASE_DIR)/zynq7/harness/zybo/swbut run
 
 db-check:
 	@true


tmichalak:
@mithro 
I didn't find anything in the failing fuzzer. It passed locally even when when I downloaded the failed fuzzer that is packed on error by kokoro.
I rerun the CI and it passed so maybe this was some kind of a glitch?

Makefile
```diff
@@ -137,7 +137,7 @@ db-extras-zynq7:
 	+source minitests/roi_harness/zybo-swbut.sh && $(MAKE) -C fuzzers part_only
 	+source minitests/roi_harness/zybo-swbut.sh && \
 		$(MAKE) -C minitests/roi_harness \
-			HARNESS_DIR=database/artix7/harness/zybo/swbut run
```
mithro: Thus should probably be `run copy`?

tmichalak: Well, I was about to ask about that.
The minitests/roi_harness/Makefile is:
<pre>
copy: $(HARNESS_FILES)
    mkdir -p $(HARNESS_DIR)
    cp $(HARNESS_FILES) $(HARNESS_DIR)

run:
    make clean
    make copy
</pre>
So why do we additionally explicitly call ``run copy`` in artix7's db-extras?

mithro: Dunno...

tmichalak: @litghost I think you added the ``run copy`` to db-extras-artix7`` (correct me if I am wrong). Maybe we should remove the redundant ``copy`` there and this one is good to merge.

tmichalak:
@litghost @mithro Do you insist on adding the `run copy` instead of simply having `run`?
If not then there is no other reason keeping this small PR from being merged.

PR #1057, opened 2019-09-18
utils: Add Spartan6 bitstream analyzer tool

tmichalak:
This PR adds a Spartan 6 bitstream analyzer tool.
It has been used to create the tools from PR #1025. 
The script reads a Spartan6 bitstream and prints out some useful information.
It can also create a frames file with the configuration data words.
The bitstream is analyzed word by word and interpreted according to
the UG380 Configuration User Guide.
The tool can be used to derive the initialization, startup and finalization
sequence as well as the configuration data. The latter is written to a frames
file which can be used by the bitstream tools such as frames2bit to generate
a valid bitstream.

### DIFF
--- a/utils/sp6_bitstream_analyzer.py
+++ b/utils/sp6_bitstream_analyzer.py
@@ -1,0 +1,420 @@
+#!/usr/bin/env python3
+'''
+Spartan 6 bitstream analyzer tool.
+
+This script reads a Spartan6 bitstream and prints out some useful information.
+It can also create a frames file with the configuration data words.
+The bitstream is analyzed word by word and interpreted according to
+the UG380 Configuration User Guide.
+
+The tool can be used to derive the initialization, startup and finalization
+sequence as well as the configuration data. The latter is written to a frames
+file which can be used by the bitstream tools such as frames2bit to generate
+a valid bitstream.
+'''
+
+import argparse
+from io import StringIO
+
+conf_regs = {
+    0: "CRC",
+    1: "FAR_MAJ",
+    2: "FAR_MIN",
+    3: "FDRI",
+    4: "FDRO",
+    5: "CMD",
+    6: "CTL",
+    7: "MASK",
+    8: "STAT",
+    9: "LOUT",
+    10: "COR1",
+    11: "COR2",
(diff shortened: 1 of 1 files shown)


tmichalak:
@mithro This PR fails CI as the script uses f-strings hence requires Python 3.6. Can we bump Python for CI or should I falback to str.format()?

mithro:
Please send a separate pull request to bump conda to the latest Python version.

litghost:
> @mithro This PR fails CI as the script uses f-strings hence requires Python 3.6. Can we bump Python for CI or should I falback to str.format()?

Let's just avoid f-strings and use str.format?

tmichalak:
@mithro @litghost I added a commit that converts f-strings into str.format() so that we don't have to do the bump now. Once we do we can just revert this one commit.

utils/sp6_bitstream_analyzer.py
```diff
@@ -0,0 +1,422 @@
...
+    11: "SHUTDOWN",
+    13: "DESYNC",
+    14: "IPROG"
+}
+
+opcodes = ("NOP", "READ", "WRITE", "UNKNOWN")
+
+
+def KnuthMorrisPratt(text, pattern):
+    '''Yields all starting positions of copies of the pattern in the text.
```
mithro: This is not formatted correctly.

tmichalak: Do you mean the docstring, dictionary or tuple? I do recall runinng format-py on this script...

mithro:
```python
def KnuthMorrisPratt(text, patter):
    ''''Yields all starting positions of copies of the pattern in the text.
    
    Calling conventions are similar to string.find, but its arguments can be
    lists or iterators, not just strings, it returns all matches, not just
    the first one, and it does not need the whole text in memory at once.
    Whenever it yields, it will have read the text exactly up to and including
    the match that caused the yield.
    '''
```

tmichalak: Fixed.

tmichalak:
@mithro Yes, it is.

PR #1071, opened 2019-09-27
README/Dependencies - Updated "Quickstart Guide"

JakeMercer:
Updated the Quickstart Guide section of the README as some deviations were
required when following this with a fresh Ubuntu 16.04 install.  Fixes to
''requirements.txt'' were required to remove the "futures" dependency which is
in the python3 standard library and is only intended for use with python2;
"fasm" now goes by "pyfasm".

Signed-off-by: Jake Mercer <jmercer04@qub.ac.uk>

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,26 +46,32 @@ Build the C++ tools:
     make build
 
 ### Step 5: ###
+Alias python to python3 to avoid using python2 accidentally:
+
+    alias python=python3
+
+Choose one of the following options:
+
 (Option 1) - Install the Python environment locally
 
-    sudo apt-get install virtualenv python3-virtualenv python3-yaml
+    sudo apt-get install virtualenv python3 python3-pip python3-virtualenv python3-yaml
     make env
 
 (Option 2) - Install the Python environment globally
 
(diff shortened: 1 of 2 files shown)


requirements.txt
```diff
@@ -1,5 +1,4 @@
-fasm
-futures
+pyfasm
```
litghost: This is incorrect, as fasm is referring to https://github.com/SymbiFlow/fasm not to https://github.com/srounet/pyfasm.  Maybe updated it something like https://stackoverflow.com/questions/16584552/how-to-state-in-requirements-txt-a-direct-github-source ?

README.md
```diff
@@ -46,26 +46,32 @@ Build the C++ tools:
     make build
 
 ### Step 5: ###
+Alias python to python3 to avoid using python2 accidentally:
+
+    alias python=python3
```
litghost: Why was this required?  Are there commands in prjxray that still reference python (rather than python3)?

JakeMercer: You're right, I don't think this is necessary; I will remove it.

PR #1078, opened 2019-10-03
Merge bitstream tools for Series7 and Spartan6

tmichalak:
This PR supersedes #1025 
Spartan6 support has been added with the use of a modified bitstream tools API.
The modifications of the API make it possible to add support for new architectures without much code duplication. 
Unit tests for Series7 and Spartan6 have been adapted to the new API.


tmichalak:
@litghost I created this PR to start the code review discussion as soon as possible.
Just while creating this PR I started thinking about whether this PR should be divided into several smaller ones:
1. Bumping C++ standard to C++17 - Travis CI failed immediately because of this
2. Changing the bitstream tools library to a more generic API (templating)
3. Adding support for Spartan6
Once those are done I could continue the work on:
4. Adding support for other architectures

litghost:
> this PR should be divided into several smaller ones

Please split this into small PR's, this one is very difficult to review.

tmichalak:
@mithro @litghost How do we update GCC to version 7 in CI?

litghost:
> 1. Bummping C++ standard to C++17

Why is this required again?

tmichalak:
@litghost I needed this for the std::variant implementation. Other than that why not use the latest stable c++ standard? We could get rid of the absl dependency we use now for `optional` for example.

litghost:
> @litghost I needed this for the std::variant implementation. Other than that why not use the latest stable c++ standard? We could get rid of the absl dependency we use now for `optional` for example.

There are couple reasons not to immediately use the new C++ standard, but mostly importantly is the same reason that you are running into with Travis.  Older GCC versions are going to be prevalent, and requiring developers to use GCC 7 to compile this code limits our developer field.  VTR for example supports GCC 5/6/7/8 and clang 6 and 7.

Given that we already have the absl dependency, which contains `absl::variant`, I would prefer we use the `absl` variant over requiring an update to a newer GCC.

mithro:
I believe it is okay for us to require C++14. C++17 is still too new for us to require it.

litghost:
> I believe it is okay for us to require C++14. C++17 is still too new for us to require it.

+1, I agree.

PR #1079, opened 2019-10-04
Add pin functions column to package pins output.

litghost:
This is required to know which pin is a PUDC pin, which requires special
handling.


mithro:
Looks like it broke Kintex 7?

litghost:
> Looks like it broke Kintex 7?

Ya, I see that.  I'll re-run and see if it is consistent.

PR #1088, opened 2019-10-07
Bitstream Tools: Add Spartan6 support

tmichalak:
This PR supersedes #1025 and depends on #1091  
Spartan6 support has been added with the use of a more generic bitstream tools API.
Created a Spartan6 version of the unit tests that have been adapted to the new API.

### DIFF
--- a/lib/include/prjxray/xilinx/frames.h
+++ b/lib/include/prjxray/xilinx/frames.h
@@ -9,7 +9,6 @@
 
 #include <absl/strings/str_split.h>
 #include <prjxray/xilinx/architectures.h>
-#include <prjxray/xilinx/xc7series/ecc.h>
 
 namespace prjxray {
 namespace xilinx {
@@ -67,13 +66,18 @@ int Frames<ArchType>::readFrames(const std::string& frm_file_str) {
 		std::vector<std::string> frame_data_strings =
 		    absl::StrSplit(frame_delta.second, ',');
 
-		if (frame_data_strings.size() != ArchType::words_per_frame) {
-			std::cerr << "Frame " << std::hex << frame_address
-			          << ": found " << std::dec
-			          << frame_data_strings.size()
-			          << " words instead of "
-			          << ArchType::words_per_frame << std::endl;
-			continue;
+		// Spartan6's IOB frames can have different word count
+		if (!std::is_same<ArchType, Spartan6>::value) {
+			if (frame_data_strings.size() !=
+			    ArchType::words_per_frame) {
+				std::cerr
+				    << "Frame " << std::hex << frame_address
+				    << ": found " << std::dec
+				    << frame_data_strings.size()
+				    << " words instead of "
+				    << ArchType::words_per_frame << std::endl;
+				continue;
+			}
 		}
 
 		FrameData frame_data(frame_data_strings.size(), 0);
--- a/lib/include/prjxray/xilinx/spartan6/block_type.h
+++ b/lib/include/prjxray/xilinx/spartan6/block_type.h
@@ -1,0 +1,37 @@
+#ifndef PRJXRAY_LIB_XILINX_SPARTAN6_BLOCK_TYPE_H_
+#define PRJXRAY_LIB_XILINX_SPARTAN6_BLOCK_TYPE_H_
+
+#include <ostream>
+
+#include <yaml-cpp/yaml.h>
+
+namespace prjxray {
+namespace xilinx {
+namespace spartan6 {
+
+// According to UG380 pg. 97 there are 3 types of configuration frames:
+// Type 0: Core: CLB, DSP, input/output interconnect (IOI), clocking
+// Type 1: Block RAM
+// Type 2: IOB
+enum class BlockType : unsigned int {
+	CLB_IOI_CLK = 0x0,
+	BLOCK_RAM = 0x1,
+	IOB = 0x2,
+};
+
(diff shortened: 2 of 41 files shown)


litghost:
CI failure looks like a formatting issue.

tmichalak:
@litghost 
I would like to merge #1090 before this one. Adding Spartan6 and then other architectures won't need explicit template instantiations in the source files anymore.

litghost:
> @litghost
> I would like to merge #1090 before this one. Adding Spartan6 and then other architectures won't need explicit template instantiations in the source files anymore.
> Of course I will need to rebase on the future new master.

#1090 is merged, let's rebase this PR and get it green on CI.

litghost:
Let's rebase this one again.  Can any of the changes in this PR take notes from the changes made to the US or US+ changes?

tmichalak:
Ha, you beat me with the comment about rebasing by 9 seconds:)

tools/bittool.cc
```diff
@@ -170,14 +170,28 @@ struct DeviceIdGetter {
 			            ArchType::ConfRegType::IDCODE);
 		    });
 		if (idcode_packet != reader->end()) {
-			if (idcode_packet->data().size() != 1) {
-				std::cerr
-				    << "Write to IDCODE with word_count != 1"
-				    << std::endl;
-				return 1;
+			if (std::is_same<ArchType, xilinx::Spartan6>::value) {
```
litghost: This seems weird.  Can think of a way to make a better way to do this in bittool and bitread (which both have a similar block)?

tmichalak: At the time when this was done, I tried to leave the most of the untouched and focused on the relevant part. 
However, this part, and a probably also some other of the original code could use some refactor.

tmichalak:
> Can any of the changes in this PR take notes from the changes made to the US or US+ changes?
I don't think there is much more we can gain with further refactor, at least with the current software architecture. The differences between Spartan6 and Series7, US and US+ are much bigger than the rest among themselves.

lib/include/prjxray/xilinx/frames.h
```diff
@@ -67,13 +66,18 @@ int Frames<ArchType>::readFrames(const std::string& frm_file_str) {
...
 		    absl::StrSplit(frame_delta.second, ',');
 
-		if (frame_data_strings.size() != ArchType::words_per_frame) {
-			std::cerr << "Frame " << std::hex << frame_address
-			          << ": found " << std::dec
-			          << frame_data_strings.size()
-			          << " words instead of "
-			          << ArchType::words_per_frame << std::endl;
-			continue;
+		// Spartan6's IOB frames can have different word count
```
litghost: Should each arch have a function that checks this?  On xc7, US and US+, it will be a straight forward check, and spartan6 will have a data lookup?

lib/include/prjxray/xilinx/spartan6/block_type.h
```diff
@@ -0,0 +1,33 @@
...
+
+#include <ostream>
+
+#include <yaml-cpp/yaml.h>
+
+namespace prjxray {
+namespace xilinx {
+namespace spartan6 {
+
+enum class BlockType : unsigned int {
```
litghost: Comment where this comes from

lib/include/prjxray/xilinx/spartan6/command.h
```diff
@@ -0,0 +1,34 @@
+#ifndef PRJXRAY_LIB_XILINX_SPARTAN6_COMMAND_H_
+#define PRJXRAY_LIB_XILINX_SPARTAN6_COMMAND_H_
+
+namespace prjxray {
+namespace xilinx {
+namespace spartan6 {
+
+enum class Command : uint32_t {
```
litghost: Comment where this comes from

lib/xilinx/spartan6/frame_address.cc
```diff
@@ -0,0 +1,89 @@
...
+
+namespace prjxray {
+namespace xilinx {
+namespace spartan6 {
+
+FrameAddress::FrameAddress(spartan6::BlockType block_type,
+                           uint8_t row,
+                           uint8_t column,
+                           uint16_t minor) {
+	address_ = bit_field_set(0, 31, 28, block_type);
```
litghost: Comment where this comes from

lib/include/prjxray/xilinx/frames.h
tmichalak: What kind of check do you have in mind for Spartan6? Do you mean something like checking if the current frames is indeed an IOB frame and if not perform the same check as for all other architectures?

litghost: Sort of.  Is the IOB frame fully variable, or just a different number?

litghost: Let me rephrase.  Is there a way to write this check that is valid for all spartan 6 frame addresses?  Why not implement that check?

tmichalak: Unsure. I just read that the IOB frame length can differ, but it didn't say exactly how and when. In my particular case it had something around 20 when the regular frames is 65. But I can't tell what the number is gonna be for other parts. So at this moment I can't really think of any generic check...

litghost: Ok, let's file an issue to investigate and fix this later.

tmichalak: Filed #1113.

PR #1090, opened 2019-10-10
Bitstream tools: Move template methods from source to header files

tmichalak:
This PR refactors the bitstream tools api by mostly moving template methods to the header files.
This way there is no need to instantiate the templates in the source file for each architecture we will add support for.


lib/xilinx/xc7series/ecc.cc
```diff
@@ -36,6 +38,21 @@ uint32_t icap_ecc(uint32_t idx, uint32_t data, uint32_t ecc) {
...
+static uint32_t calculateECC(const std::vector<uint32_t>& data) {
+	uint32_t ecc = 0;
+	for (size_t ii = 0; ii < data.size(); ++ii) {
+		ecc = xc7series::icap_ecc(ii, data[ii], ecc);
+	}
+	return ecc;
+}
+
+void updateECC(std::vector<uint32_t>& data) {
+	assert(data.size() != 0);
```
litghost: Let's actually change this `assert(data.size() >= 0x32)`?
Also lets move 0x32 into a `constexpr size_t`

lib/include/prjxray/xilinx/frames.h
```diff
@@ -29,15 +33,88 @@ class Frames {
...
+
+		FrameData frame_data(frame_data_strings.size(), 0);
+		std::transform(frame_data_strings.begin(),
+		               frame_data_strings.end(), frame_data.begin(),
+		               [](const std::string& val) -> uint32_t {
+			               return std::stoul(val, nullptr, 16);
+		               });
+
+		// Spartan6 doesn't have ECC
+		if (std::is_same<ArchType, Series7>::value) {
```
litghost: Just have `Spartan6` specialized updateECC to a `nop`.

lib/xilinx/xc7series/ecc.cc
tmichalak: Done

lib/include/prjxray/xilinx/frames.h
tmichalak: In this PR I removed the type checking and added a generic updateECC template method.
I will add the specialization in the Spartan6 related PR - see [commit](https://github.com/SymbiFlow/prjxray/pull/1088/commits/083363f5806e5b405cd61d8b9869b37986f32a87)

PR #1091, opened 2019-10-10
Bitstream Tools: Add UltraScale devices support

tmichalak:
This PR adds initial support for the UltraScale devices to the bitstream tools.
The bitstream used for testing the tools was for xcku040-ffva1156-2 which is featured on KCU105 Kintex UltraScale (https://www.xilinx.com/products/boards-and-kits/kcu105.html) board.
The output bitstream hasn't been tested on real hardware yet.


litghost:
Please rebase

lib/include/prjxray/xilinx/architectures.h
```diff
@@ -37,12 +38,24 @@ class Series7 : public Architecture {
...
 };
 
+class UltraScale : public Series7 {
+       public:
+	UltraScale() : name_("UltraScale") {}
+	const std::string& name() const override { return name_; }
+	static constexpr int words_per_frame = 123;
+
+       private:
+	std::string name_;
```
litghost: Should be `const std::string`?

lib/include/prjxray/xilinx/bitstream_reader.h
```diff
@@ -98,6 +98,32 @@ BitstreamReader<Series7>::InitWithBytes(T bitstream) {
 	return BitstreamReader<Series7>(std::move(words));
 }
 
+template <>
+template <typename T>
+absl::optional<BitstreamReader<UltraScale>>
+BitstreamReader<UltraScale>::InitWithBytes(T bitstream) {
```
litghost: `<UltraScale>::InitWithBytes` and `<Series7>::InitWithBytes` should probably just use a template function?  They seem identical except for `s/Series7/UltraScale/`?

lib/xilinx/configuration.cc
```diff
@@ -41,6 +41,31 @@ Configuration<Series7>::createType2ConfigurationPacketData(
...
+		          std::back_inserter(packet_data));
+
+		auto next_address = part->GetNextFrameAddress(frame.first);
+		if (next_address &&
+		    (next_address->block_type() != frame.first.block_type() ||
+		     next_address->is_bottom_half_rows() !=
+		         frame.first.is_bottom_half_rows() ||
+		     next_address->row() != frame.first.row())) {
+			packet_data.insert(packet_data.end(),
+			                   UltraScale::words_per_frame * 2, 0);
```
litghost: Please make a constant of the `2`.  Is it just `sizeof(uint16_t)`, etc?

tmichalak: The `2` here has nothing to do with the word size. It's has just been empirically proven that at the end of the Type 2 packet there are 2 full (2 * ArchType::words_per_frame) frames with zeros. Adding the `sizeof(uint16_t)` would be misleading as it would imply that the number of zero frames has something to do with the data type which is not the case.

litghost: Ok sure, so put the details you just wrote into a comment above a new constant with a good name.

lib/include/prjxray/xilinx/bitstream_reader.h
tmichalak: Yes, `Configuration<UltraScale>::InitWithPackets` along with `Configuration<UltraScale>::createType2ConfigurationPacketData` are other candidates for such a change.

lib/include/prjxray/xilinx/architectures.h
tmichalak: Yes, corrected that.

lib/include/prjxray/xilinx/configuration.h
```diff
@@ -65,144 +65,37 @@ class Configuration {
...
-			}
-			default:
-				break;
+template <typename ArchType>
+typename Configuration<ArchType>::PacketData
+Configuration<ArchType>::createType2ConfigurationPacketData(
+    const typename Frames<ArchType>::Frames2Data& frames,
+    absl::optional<typename ArchType::Part>& part) {
+	PacketData packet_data;
+	static const int kZeroFramesSeparatorWords =
```
litghost: Comment on this constant please.

lib/include/prjxray/xilinx/configuration.h
```diff
@@ -286,7 +179,7 @@ Configuration<UltraScalePlus>::InitWithPackets(
 					start_new_write = false;
 				}
 
-				// UltraScale frames are 93-words long.  Writes
+				// 7-series frames are 101-words long.  Writes
```
litghost: Comment should now be series agnostic?

PR #1092, opened 2019-10-10
Updated wrong string path added in syspath before import

lromor:


minitests/roi_harness/demo_sw_led_fasm.py
```diff
@@ -2,7 +2,7 @@
 
 import sys
 import os
-sys.path.append(os.getenv("XRAY_DIR") + "/tools")
+sys.path.append(os.getenv("XRAY_DIR") + "/utils")
```
litghost: Should probably try `XRAY_UTILS_DIR`?

PR #1098, opened 2019-10-14
Add pin functions to tilegrid.

litghost:
- Add support to emit PUDC_B pullup if unused (for A7 and Z7 fabrics).

### DIFF
--- a/utils/fasm2frames.py
+++ b/utils/fasm2frames.py
@@ -54,9 +54,55 @@ def dump_frm(f, frames):
             '0x%08X ' % addr + ','.join(['0x%08X' % w for w in words]) + '\n')
 
 
-def run(db_root, filename_in, f_out, sparse=False, roi=None, debug=False):
+def find_pudc_b(db):
+    """ Find PUDC_B pin func in grid, and return the tile and site prefix.
+
+    The PUDC_B pin is a special 7-series pin that controls unused pin pullup.
+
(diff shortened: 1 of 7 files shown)


utils/fasm2frames.py
```diff
@@ -54,10 +54,56 @@ def dump_frm(f, frames):
             '0x%08X ' % addr + ','.join(['0x%08X' % w for w in words]) + '\n')
 
 
-def run(db_root, filename_in, f_out, sparse=False, roi=None, debug=False):
+def find_pudc_b(db):
+    """ Find PUDC_B pin func in grid, and return the tile and site prefix.
+
+    The PUDC_B pin is a special 7-series pin that controls unused pin pullup.
+
+    If the PUDC_B is unused, it is configured as an input with a PULLUP.
```
mithro: This is my understanding;

 * The PUDC_B pin controls unused pin pullup state by the value being samples during start up sequence. 
 * If you don't have an external pullup or pulldown on this pin you need to configure it with an internal PULLUP / PULLDOWN so that it isn't randomly floating when sampled (and hence you get a repeatable unused pin pullup state).

litghost: Not exactly.  The documentation specifically says to **physically** pull up or pull down the signal.  It is not exactly clear why Vivado emits bitstreams with a PULLUP, but I'd guess power savings.

litghost:
I'm gonna wait for green CI, and then merge.

PR #1099, opened 2019-10-14
Enable ppips on aliased segbit files.

litghost:
This is required for emitting `[LR]IOI3_SING` frames in the ROI breakout.


mithro:
If others say LGTM, then I agree :-)

PR #1103, opened 2019-10-15
Bitstream Tools: Add UltraScalePlus support

tmichalak:
This PR adds initial support for the UltraScale+ devices to the bitstream tools.
The bitstream used for testing the tools was for Zynq UltraScale+ XCZU7EV-2FFVC1156 MPSoC which is featured on the ZCU104 evaluation kit (https://www.xilinx.com/products/boards-and-kits/zcu104.html).
The bitstream generated by the patcher and frames2bit tools were both successfully tested on hardware.

### DIFF
--- a/lib/CMakeLists.txt
+++ b/lib/CMakeLists.txt
@@ -6,6 +6,7 @@ add_library(libprjxray
 	xilinx/configuration_packet.cc
 	xilinx/configuration_register.cc
 	xilinx/configuration.cc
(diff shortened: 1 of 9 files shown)


tmichalak:
@mithro 
Definitely, I consider this change as a "cherry on top" type of commit, but will definitely create a PR with it once we merge all other PRs with adding new architectures to the bitstream tools.

PR #1112, opened 2019-10-20
Whitespace

### LINKED ISSUE #758: Check for trailing whitespace in files
Travis should check that people are not adding trailing whitespace in files. Should include the documentation as well as the code.

JakeMercer:
Fixes #758


Makefile
```diff
@@ -53,7 +53,7 @@ test-cpp:
 
 # Auto formatting of code.
 # ------------------------
-FORMAT_EXCLUDE = $(foreach x,$(ALL_EXCLUDE),-and -not -path './$(x)/*')
+FORMAT_EXCLUDE = $(foreach x,$(ALL_EXCLUDE),-and -not -path './$(x)/*') -and -not -name *.bit
```
tmichalak: @JakeMercer why is that needed? I don't recall us keeping any .bit files in git?

JakeMercer: There are at least two .bit files in `lib/test_data` that are tracked by Git so I figured it would be better to be safe on the off-chance those changed.  Though maybe they should be generated instead of tracked?

Another thing I was thinking that we should probably read the `.gitignore` into a Make variable and use it for exclusions for `make format` also to help speed it up.

tmichalak: Oh, yes, you are right about the test_data bit files. They are used in the unit tests. They can indeed be excluded.
I guess the idea with .gitignore also makes sense.

JakeMercer: Had a quick look at the `.gitignore` idea, would need some thought as there are multiple things in the file that would have to be dealt with (comments, non-glob patterns, etc.) [1](http://schacon.github.io/git/gitignore.html#_pattern_format).  Probably not worth it for this PR.

Rebased on master already, so ready to go.

PR #1116, opened 2019-10-23
Revert "Whitespace"

litghost:
Reverts SymbiFlow/prjxray#1112
FYI @JakeMercer 

This causes make format to take >10 minutes on an active tree.  Something about the filtering is broken and causes it to touch built files.


JakeMercer:
No worries, I'll take a look at it, might be better to whitelist patterns instead of blacklisting.  Or filtering based on what Git is tracking.

PR #1121, opened 2019-10-24
Expand README a bit to include also Spartan6 and UltraScale series

plaes:
Signed-off-by: Priit Laes <plaes@plaes.org>

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,16 +4,20 @@
 ![Documentation Status](https://readthedocs.org/projects/pyxray/badge/?version=latest)
 ![License](https://img.shields.io/github/license/SymbiFlow/prjxray.svg)
 
-Documenting the Xilinx 7-series bit-stream format.
+Documenting the Xilinx FPGA bit-stream formats.
 
-This repository contains both tools and scripts which allow you to document the
-bit-stream format of Xilinx 7-series FPGAs.
+This repository contains both tools and scripts used to document the
+bit-stream formats of various Xilinx FPGA families:
+ * Xilinx 7-Series
+ * Xilinx Spartan-6
+ * Xilinx UltraScale and UltraScalePlus
+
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -4,16 +4,20 @@
...
 
-This repository contains both tools and scripts which allow you to document the
-bit-stream format of Xilinx 7-series FPGAs.
 
 More documentation can be found published on [prjxray ReadTheDocs site](http://prjxray.readthedocs.io/en/latest/) - this includes;
  * [Highlevel Bitstream Architecture](http://prjxray.readthedocs.io/en/latest/architecture/overview.html)
  * [Overview of DB Development Process](http://prjxray.readthedocs.io/en/latest/db_dev_process/overview.html)
 
-# Quickstart Guide
+# Quickstart Guide (for Spartan 7-Series)
```
litghost: Should be `Xilinx 7-Series` not `Spartan 7-Series`, yes?

plaes: Yes, indeed :(

mithro:
Hi @plaes,

Firstly, thank you for trying to make our docs better, it is highly appreciated!

For the moment, I think adding this stuff around Spartan 6 and Ultrascale parts into to the README might be misleading and give people the wrong message.

While a bunch of the bitstream tools **do** support Spartan 6 and Ultrascale bitstream processing there **isn't** anything hugely useful you can do with that at the moment. Without fuzzers and the rest of the stuff that makes up Project X-Ray it would be a bit misleading to suggest that this repo covers those parts.

The bitstream work was done mainly because it was pretty easy to do (due to similar bitstream formats) and there is a hope of **eventually** doing something like the Project X-Ray fuzzers for both these series. If Spartan 6 and/or Ultrascale actually progresses further, then we will most certainly update the documentation in the way you are suggesting.

We would love help improving some of the other Project X-Ray README files. The fact that some of the fuzzers are missing README files is not great!

Look forward to your future contributions and thank you once again!

Tim 'mithro' Ansell

Issue #1141, opened 2019-11-12
Incorrect PLL IN_USE feature


litghost:
https://github.com/SymbiFlow/symbiflow-arch-defs/pull/1150 failed as a result of a bad segbit definition:

```
FAILED: cd /tmpfs/src/github/symbiflow-arch-defs-presubmit-xc7/build/xc7/tests/pll && /usr/bin/cmake -E env PYTHONPATH=/tmpfs/src/github/symbiflow-arch-defs-presubmit-xc7/build/env/conda/lib/python3.7/site-packages:/tmpfs/src/github/symbiflow-arch-defs-presubmit-xc7/third_party/prjxray:/tmpfs/src/github/symbiflow-arch-defs-presubmit-xc7/third_party/prjxray/third_party/fasm /tmpfs/src/github/symbiflow-arch-defs-presubmit-xc7/build/env/conda/bin/python3 /tmpfs/src/github/symbiflow-arch-defs-presubmit-xc7/third_party/prjxray/utils/fasm2frames.py --db-root /tmpfs/src/github/symbiflow-arch-defs-presubmit-xc7/third_party/prjxray-db/artix7 --sparse --emit_pudc_b_pullup /tmpfs/src/github/symbiflow-arch-defs-presubmit-xc7/build/xc7/tests/pll/pll_ext_basys3_x1y0/artix7-xc7a50t-roi-virt-xc7a50t-test/top.fasm /tmpfs/src/github/symbiflow-arch-defs-presubmit-xc7/build/xc7/tests/pll/pll_ext_basys3_x1y0/artix7-xc7a50t-roi-virt-xc7a50t-test/top.frames
Traceback (most recent call last):
  File "/tmpfs/src/github/symbiflow-arch-defs-presubmit-xc7/third_party/prjxray/utils/fasm2frames.py", line 199, in <module>
    main()
  File "/tmpfs/src/github/symbiflow-arch-defs-presubmit-xc7/third_party/prjxray/utils/fasm2frames.py", line 195, in main
    emit_pudc_b_pullup=args.emit_pudc_b_pullup)
  File "/tmpfs/src/github/symbiflow-arch-defs-presubmit-xc7/third_party/prjxray/utils/fasm2frames.py", line 123, in run
    assembler.parse_fasm_filename(filename_in, extra_features=extra_features)
[...]

PR #1142, opened 2019-11-14
Update to 032 which fixes conflict with 034

mkurc-ant:
Fixes https://github.com/SymbiFlow/prjxray/issues/1141

### DIFF
--- a/fuzzers/032-cmt-pll/generate.py
+++ b/fuzzers/032-cmt-pll/generate.py
@@ -10,9 +10,6 @@ def bitfilter(frame, word):
     if frame < 28:
         return False
 
-    if frame == 25 and word == 3121:
-        return False
-
     return True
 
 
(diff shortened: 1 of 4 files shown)


fuzzers/032-cmt-pll/generate.py
```diff
@@ -10,7 +10,15 @@ def bitfilter(frame, word):
     if frame < 28:
         return False
 
-    if frame == 25 and word == 3121:
+    # Those bits are in conflict with fuzzer 034.
+    if (frame, word) in (
+        (28, 10),
```
litghost: I'm not a huge fan of hard coding conflicts.  Any way we can vary the input stimulus to avoid this kind of filter list?  034 clearly toggles these bits, can we modify 032 to do the same?

mkurc-ant: That would probably require routing manually through some pips. I can take a look at it.

mkurc-ant:
@litghost I added forced routing through some wires and the conflicting bits in `IN_USE` disappeared.

I had to disable outputting of the following features as they were composed of routing bits (presumably) solved by 034:
```
COMPENSATION.INTERNAL
COMPENSATION.BUF_IN_OR_EXTERNAL_OR_ZHOLD_CLKIN_BUF
```

fuzzers/032-cmt-pll/generate.py
litghost: Thanks, I think this is fuzzer is better for those changes!

PR #1149, opened 2019-11-19
001-part-yaml: Add iobanks information to part's json

tmichalak:
This PR adds the IO banks information to the part's json which is generate in the `001-part-yaml` fuzzer.
This information is needed for the xdc plugins (https://github.com/SymbiFlow/yosys-symbiflow-plugins/pull/1) to work and is a prerequisite for (https://github.com/SymbiFlow/symbiflow-arch-defs/pull/1158)

### DIFF
--- a/fuzzers/001-part-yaml/add_iobanks.py
+++ b/fuzzers/001-part-yaml/add_iobanks.py
@@ -1,0 +1,30 @@
+#!/usr/bin/env python3
+'''
+Script for adding the IO Banks information to the Part's generated JSON.
+'''
+import argparse
+import json
+
+
+def main(argv):
+    with open(args.part_json) as json_file, open(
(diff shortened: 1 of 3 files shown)


fuzzers/001-part-yaml/add_iobanks.py
```diff
@@ -0,0 +1,24 @@
+#!/usr/bin/env python3
+'''
+Script for adding the IO Banks information to the Part's generated JSON.
+'''
+import sys
+import json
+
+if len(sys.argv) != 3:
```
litghost: Please give this a `def main()`, etc

tmichalak: Done, will merge when green.

PR #1157, opened 2019-11-29
Minitest for different IOSTANDARDs

mkurc-ant:
This is a minitest that allows to check which fasm features are enabled for which IO standard. It also allows to identify features that are active for unused IOBs when another IOB is active in the same bank. Moreover there is an utility script that automatizes translation of IOB verilog parameters to fasm features.

Signed-off-by: Maciej Kurc <mkurc@antmicro.com>

### DIFF
--- a/minitests/iostandard/dump_iobs.tcl
+++ b/minitests/iostandard/dump_iobs.tcl
@@ -1,0 +1,29 @@
+proc dump_iobs {file_name} {
+
+    set fp [open $file_name w]
+    puts $fp "tile,site_name,site_type,clock_region,bank,is_bonded,is_clock,is_global_clock,is_vref"
+
+    foreach tile [get_tiles *IOB33*] {
+        foreach site [get_sites -of_objects $tile] {
+            set site_type [get_property SITE_TYPE $site]
+            set clock_region [get_property CLOCK_REGION $site]
+            set is_bonded [get_property IS_BONDED $site]
+
+            set pin [get_package_pins -of_objects $site]
+            set bank [get_property BANK $pin]
+            set is_clock [get_property IS_CLK_CAPABLE $pin]
+            set is_global_clock [get_property IS_GLOBAL_CLK $pin]
+            set is_vref [get_property IS_VREF $pin]
+
+            puts $fp "$tile,$site,$site_type,$clock_region,$bank,$is_bonded,$is_clock,$is_global_clock,$is_vref"
+        }
+    }
+
+    close $fp
+}
+
+create_project -force -in_memory -name dump_iobs -part $::env(VIVADO_PART)
+set_property design_mode PinPlanning [current_fileset]
+open_io_design -name io_1
+
+dump_iobs "iobs-$::env(VIVADO_PART).csv"
--- a/minitests/iostandard/dump_iobs.tcl
+++ b/minitests/iostandard/dump_iobs.tcl
@@ -1,7 +1,7 @@
 proc dump_iobs {file_name} {
 
     set fp [open $file_name w]
-    puts $fp "tile,site_name,site_type,clock_region,bank,pkg_pin,is_bonded,is_clock,is_global_clock,is_vref"
+    puts $fp "tile,site_name,site_type,clock_region,bank,pkg_pin,is_bonded,is_clock,is_global_clock,is_vref,pin_func"
 
     foreach tile [get_tiles *IOB33*] {
         foreach site [get_sites -of_objects $tile] {
@@ -15,8 +15,9 @@ proc dump_iobs {file_name} {
             set is_clock [get_property IS_CLK_CAPABLE $pin]
             set is_global_clock [get_property IS_GLOBAL_CLK $pin]
             set is_vref [get_property IS_VREF $pin]
+            set pin_func [get_property PIN_FUNC $pin]
 
-            puts $fp "$tile,$site,$site_type,$clock_region,$bank,$pkg_pin,$is_bonded,$is_clock,$is_global_clock,$is_vref"
+            puts $fp "$tile,$site,$site_type,$clock_region,$bank,$pkg_pin,$is_bonded,$is_clock,$is_global_clock,$is_vref,$pin_func"
         }
     }
 
--- a/minitests/iostandard/dump_iobs.tcl
+++ b/minitests/iostandard/dump_iobs.tcl
@@ -11,11 +11,12 @@ proc dump_iobs {file_name} {
 
             set pin [get_package_pins -of_objects $site]
             set bank [get_property BANK $pin]
+            set pkg_pin [get_property NAME $pin]
             set is_clock [get_property IS_CLK_CAPABLE $pin]
             set is_global_clock [get_property IS_GLOBAL_CLK $pin]
             set is_vref [get_property IS_VREF $pin]
 
-            puts $fp "$tile,$site,$site_type,$clock_region,$bank,$is_bonded,$is_clock,$is_global_clock,$is_vref"
+            puts $fp "$tile,$site,$site_type,$clock_region,$bank,$pkg_pin,$is_bonded,$is_clock,$is_global_clock,$is_vref"
         }
     }
 
--- a/minitests/iostandard/dump_iobs.tcl
+++ b/minitests/iostandard/dump_iobs.tcl
@@ -1,7 +1,7 @@
 proc dump_iobs {file_name} {
 
     set fp [open $file_name w]
-    puts $fp "tile,site_name,site_type,clock_region,bank,is_bonded,is_clock,is_global_clock,is_vref"
+    puts $fp "tile,site_name,site_type,clock_region,bank,pkg_pin,is_bonded,is_clock,is_global_clock,is_vref"
 
     foreach tile [get_tiles *IOB33*] {
         foreach site [get_sites -of_objects $tile] {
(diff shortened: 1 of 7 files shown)


minitests/iostandard/features/README.md
```diff
@@ -0,0 +1,41 @@
+# IOSTANDARD feature correlation minitest
+
+This test checks which fasm features present in the db are set for given IO settings (IOSTANDARD + DRIVE + SLEW) and for which IOB type (IBUF, OBUF, IOBUF). It also checks what features are set on unused IOBs of a given bank in which active IOBs are instantiated. To avoid conflicts, a single iosettings combination is used within an IO bank.
```
litghost: Should add differential signalling examples (OBUFDS, OBUFTDS, IOBUFDS, etc)?

minitests/iostandard/features/analyze.py
```diff
@@ -0,0 +1,168 @@
+import os
```
litghost: File level comment

minitests/iostandard/features/Makefile
```diff
@@ -0,0 +1,76 @@
+PART=xc7a50tfgg484-1
```
litghost: Should probably be `PART?=` to enable command line overrides?

Why is `PART` different than `VIVADO_PART`?  Maybe add a comment?
Should VIVADO_PART simply be `VIVADO_PART?=$(PART)`?

minitests/iostandard/features/Makefile
```diff
@@ -0,0 +1,76 @@
...
+VIVADO_PART=xc7a50tfgg484-1
+
+BIT2FASM_ARGS= --part "$(XRAY_DIR)/database/artix7/$(PART)" --verbose
```
litghost: `artix7` should probably be `XRAY_DATABASE`?  See https://github.com/SymbiFlow/prjxray/blob/master/settings/artix7.sh#L1

minitests/iostandard/dump_iobs.tcl
```diff
@@ -0,0 +1,30 @@
...
+            set site_type [get_property SITE_TYPE $site]
+            set clock_region [get_property CLOCK_REGION $site]
+            set is_bonded [get_property IS_BONDED $site]
+
+            set pin [get_package_pins -of_objects $site]
+            set bank [get_property BANK $pin]
+            set pkg_pin [get_property NAME $pin]
+            set is_clock [get_property IS_CLK_CAPABLE $pin]
+            set is_global_clock [get_property IS_GLOBAL_CLK $pin]
+            set is_vref [get_property IS_VREF $pin]
```
litghost: Should also dump `PIN_FUNCTION`

minitests/iostandard/features/analyze.py
```diff
@@ -0,0 +1,168 @@
...
+            parts = feature.split(".", maxsplit=3)
+            if len(parts) >= 3 and parts[1] == "IOB_Y0":
+                features.append(".".join(parts[2:]))
+
+    return features
+
+
+def correlate_features(features, tile, site, set_features):
+    """
+    Correlate each feature with the fasm disassembly for given tile/site.
```
litghost: Please document arguments and return, and provide a longer description of what is happening.

minitests/iostandard/features/analyze.py
```diff
@@ -0,0 +1,168 @@
...
+
+from collections import defaultdict
+
+from prjxray import util
+import fasm
+
+# =============================================================================
+
+
+def load_iob_segbits():
```
litghost: Why not use the prjxray python library?  Then you don't have to assume anything w.r.t. `LIOB33`/`RIOB33`.  You also have to deal with the fact that `_SING` tiles have a different offset, which would be handled if you used the library decode functions?

minitests/iostandard/features/generate.py
```diff
@@ -0,0 +1,491 @@
...
+def load_iob_sites(file_name):
+    """
+    Loads IOB site dump from the given CSV file.
+    """
+    iob_sites = defaultdict(lambda: [])
+
+    with open(file_name, "r") as fp:
+        fp.readline()
+
+        for line in fp:
```
litghost: Just use `csv.DictReader`?  More flexible in the event of more columns, and less code, and it does the same thing?

minitests/iostandard/features/generate.py
```diff
@@ -0,0 +1,491 @@
...
+        #        'HSTL_II',
+        #        'HSTL_II_18',
+        #        'HSUL_12',
+        #        'MOBILE_DDR',
+    )
+
+    DRIVES = defaultdict(lambda: [None])
+    DRIVES.update(
+        {
+            "LVTTL": [4, 8, 12, 16, 24],
```
litghost: Any way we could document the DRIVE strengths via a tcl output rather than coding it this way?

minitests/iostandard/features/generate.py
```diff
@@ -0,0 +1,491 @@
...
+        num_out = 0
+        num_ino = 0
+
+        # Generate clock regions
+        region_data = []
+        for region in sorted(list(iob_sites.keys())):
+
+            # Get IO settings
+            try:
+                iosettings = next(iosettings_gen)
```
litghost: Why not use:

`for region, iosettings = sorted(zip(iob_sites, gen_iosettings), key=lambda x: x[0]):`

litghost: Can you describe in words what you were trying to do with: https://github.com/SymbiFlow/prjxray/pull/1157/files?short_path=02164c0#diff-84790ccfe259a048566f5d1695b477ddR156-R162 ?

minitests/iostandard/features/snippets.py
```diff
@@ -0,0 +1,209 @@
+import os
```
litghost: Top level comment

minitests/iostandard/features/snippets.py
```diff
@@ -0,0 +1,209 @@
...
+
+    # Open the file
+    with open(fname, "r") as fp:
+
+        # Header
+        line = fp.readline()
+        features = line.strip().split(",")[1:]
+
+        # Data
+        for line in fp:
```
litghost: Again, `csv.DictReader`

minitests/iostandard/features/README.md
mkurc-ant: It does. I just forgot to update the README. Now its fixed.

minitests/iostandard/features/analyze.py
mkurc-ant: Fixed.

minitests/iostandard/features/Makefile
mkurc-ant: Ok, I've made those as parameters and made the VIVADO_PART equal to PART. I can't recall where but I've seen that in some cases the XRay and Vivado parts to have different suffixes.

minitests/iostandard/features/Makefile
mkurc-ant: Fixed.

minitests/iostandard/dump_iobs.tcl
mkurc-ant: But why? It's not needed in the test. Actually what is needed are pkg_pin, is_vref and site name to pkg_pin mapping.

minitests/iostandard/features/analyze.py
mkurc-ant: Added more detailed docstring to the function.

minitests/iostandard/features/analyze.py
mkurc-ant: But I'm not really decoding anything here. I just need a list of fasm features of an IOB site.

minitests/iostandard/features/generate.py
mkurc-ant: Right, switched to `csv.DictReader`

minitests/iostandard/features/generate.py
mkurc-ant: Need to check that. Those values are taken from the official documentation: https://www.xilinx.com/support/documentation/user_guides/ug471_7Series_SelectIO.pdf

minitests/iostandard/features/generate.py
mkurc-ant: So there is a generator of IO settings combinations `gen_iosettings()`. But since there is 5 IO banks on the Artix7 part I need to get 5 of those settings per design. And off course the numer of IO settings combinations may not be dividable by 5 so the last design may have less than 5 banks used.

And this is implemented as a while loop that generates design. Inside it there is a loop over clock regions / IO banks which may terminate if we run out of IO setting combinations.

minitests/iostandard/features/snippets.py
mkurc-ant: Added.

minitests/iostandard/features/snippets.py
mkurc-ant: Fixed.

minitests/iostandard/dump_iobs.tcl
litghost: You will want to have special handling for at least one pin function, PUDC, see: https://github.com/SymbiFlow/prjxray/blob/master/utils/fasm2frames.py#L125

minitests/iostandard/features/Makefile
litghost: Did you forget to push?  I still see the old version?

mkurc-ant:
@litghost Yep, I committed but forgot to push... I did now.

minitests/iostandard/dump_iobs.tcl
mkurc-ant: Added skipping of PUDC pin site(s)

PR #1162, opened 2019-12-03
Tilegrid generation for Zynq 7020

tmichalak:
This PR enables the generation of the tilegrid for Zynq 7020 and makes it the default part for the Zynq database.


settings/zynq7.sh
```diff
@@ -1,5 +1,5 @@
 export XRAY_DATABASE="zynq7"
-export XRAY_PART="xc7z010clg400-1"
+export XRAY_PART="xc7z020clg400-1"
```
litghost: I don't think we want to only output tile grid for the xc7z020clg400.  In the long run we need tilegrid for multiple classes of parts.  Do you want to propose something akin to https://github.com/SymbiFlow/prjxray/blob/master/Makefile#L165 allow for tilegrids of multiple parts within the same family?

tmichalak: Sure, we definitely need support per part. This WIP PR is currently for testing how well suited are our fuzzers currently for 7020 parts.

tmichalak:
We have currently 9 failing fuzzers:
- [ ] 015-clb-nffmux
- [ ] 016-clb-noutmux
- [ ] 018-clb-ram
- [ ] 032-cmt-pll
- [ ] 035-iob-ilogic
- [ ] 035a-iob-idelay
- [ ] 036-iob-ologic
- [ ] 037-iob-pips
- [ ] 039-hclk-config

For iobs the fixes should similar to 030-iob.

acomodi:
`kokoro:force-run` seems not to be working.

acomodi:
I believe that the problem could lie in the roi_harness generation. I am unsure why it fails for artix, as this PR shouldn't touch anything related to it.

tmichalak:
All the fuzzers finish successfully when run locally. On CI the run can't even finish to produce a log.
@mithro Do you happen to have access to the build artifacts, I get the following message:

mithro:
@tmichalak What URL are you seeing that at?

tmichalak:
@mithro: it's [this](https://00e9e64bac21272e103168e8bd70afda18299e9e2b733d26ee-apidata.googleusercontent.com/download/storage/v1/b/kokoro_build_logs/o/kokoro-sisyphus%2Fresultstore%2Fprod%2Ffoss-fpga-tools%2Fprjxray%2Fpresubmit%2Fdatabase%2Fzynq7%2F9fdbb9b0-785b-482b-8ce7-f202e8b13dd8%2Fbuild.log?qk=AD5uMEvBcJHi-5yLHMcV1JQFI6A2Vc4X0KTIRDTVELW6_yolWwidBZ3GxBXc4tEGmbFsMKY6iQP7iTy_nZRM5cx14XWcEyUQCJfoUOmwDRHgT1_rz8gOPRtsi-Zzgr82Hv8m_zTq87OxNTZQK5hLIUTQPt1w6unM12fBw8Dbx74a3L0KtTRfYDhOrrFsMOmJoAcnus3ILno0RrkJM2vh8xNCSBLt2gOMLLb3Q8kJh77D2Kg3AXLJJndw57Vka_4FUfbwbgFEaL0cLNwFcC_x-XXKGrc5RqPxX57ZrLyNiJGtMvmpU8OEan4h8bGfgpiLci62PNw6YDL9sl_Fs2-WJ5sC_izPFChHmvvkFqs86InYCYkVgOP_lNotfktNJaNjnR4blDTAsrxZWmq37EvtpSdiCbzDqa51dYx1bCVJTGpZYLH7HJK7hJXhG_0XS2Ue9ARvxmmESGM2iGxUUlyA0I4HZdq5sveLvJdrmBQfhWYX2W1ZkIqjefnImu4kbHVyukCtaUxFv57ea9lgSslmTdbAC207DYG7E_pKigfvH2n3iiWlNZySVZUxYVpAaWg6X4GiUNQuQTTX2T1Ox6MocFNpSLuE5d-BLRZYX87lQ8enZTkJRyF0uO8nTxcuXZuobL1GOC8oKO1D30t8W4RXHgrSwGz5eGMKsEE6v0M7_tpp1gpdFknsr941G91qgMQoJh4iwQmDE-G-p_6zYQrI-zso0S_nRtthZc_qdMmdk4OaJwhu2ndmFIPibyX9_jn0-KcKpUWdfQVKVZOhvFHD34IMKR_KBlVU9MXKBA6iCzKRUqHr71IOV5tHI2wo0xwmrBMkKruIiSTJWW_RDXc9P8JI8X2T58CHMS4oXf2X8hVzcKV8jHrwz0U3Hl4pyhoZvIoCVuA72qP7WTWk7R9AZOUI5LUjt3NGrlPTpgyPJd8gSJpY7cMev9LNpHC4Le3olwdiJ7eO_rPg)

mithro:
@tmichalak But how did you get that URL?

fuzzers/074-dump_all/zynq7_ignored_wires.txt
```diff
@@ -1,44 +1,632 @@
...
-RIOI3_TBYTESRC_X31Y81/RIOI_I2GCLK_TOP1
-RIOI3_TBYTETERM_X31Y13/RIOI_I2GCLK_TOP1
-RIOI3_TBYTESRC_X31Y93/RIOI_I2GCLK_TOP1
-RIOI3_TBYTESRC_X31Y57/RIOI_I2GCLK_BOT1
-RIOI3_TBYTETERM_X31Y13/RIOI_I2GCLK_BOT1
-RIOI3_X31Y83/RIOI_I2GCLK_BOT1
-CMT_TOP_L_UPPER_B_X119Y31/CMT_PHASER_UP_DQS_TO_PHASER_D
-RIOI3_X31Y33/RIOI_I2GCLK_BOT1
-RIOI3_TBYTESRC_X31Y7/RIOI_I2GCLK_TOP1
+IO_INT_INTERFACE_L_X0Y5/L_INT_INTER_DQS_IOTOPHASER
```
mithro: I think this file should be sorted?

mithro:
@tmichalak / @acomodi - Can you rebase this onto master?

tmichalak:
@mithro I tried to download the log from [this](https://source.cloud.google.com/results/invocations/54639d83-f668-466e-b0ce-a7b01d360207/artifacts/) page.

litghost:
There is a suspicious diff from the A7 CI:

```
-            CMT_TOP_L_UPPER_T.CMT_TOP_R_UPPER_T_PLLE2_CLKFBIN.CMT_TOP_L_UPPER_T_FREQ_BB1 !28_11 !28_43 !28_44 !29_10 !29_11 !29_42 !29_43            
+            CMT_TOP_L_UPPER_T.CMT_TOP_R_UPPER_T_PLLE2_CLKFBIN.CMT_TOP_L_UPPER_T_FREQ_BB1 !28_11 !28_43 28_44 29_10 !29_11 !29_42 !29_43
```

That diff looks wrong.  Can you please investigate if this is a regression?  I believe @mkurc-ant worked on the `FREQ_BB1 -> CLKFBIN` pips.

litghost:
The above diff is present in all three A7, K7 and Z7.  I've also noticed that Z7 is solving for the IMUX features being solved in https://github.com/SymbiFlow/prjxray/pull/1185 , but neither A7 nor K7 is getting those solutions.

There is also the broader question of Z7010 support, but I guess with #1196 we can add support for both Z7010 and Z7020?

fuzzers/005-tilegrid/ps7_int/top.py
```diff
@@ -39,9 +39,9 @@ def run():
     params = {}
 
     # NOTE: The INT_L tile has been hardcoded and it works only for the part specified in the assertion
-    assert os.getenv('XRAY_PART') == "xc7z010clg400-1"
+    assert os.getenv('XRAY_PART') == "xc7z020clg484-1"
     for isone in util.gen_fuzz_states(1):
-        params['INT_L_X0Y50'] = isone
+        params['INT_L_X18Y100'] = isone
```
litghost: Is there a way to not hardcode this?

fuzzers/034-cmt-pll-pips/Makefile
```diff
@@ -3,13 +3,9 @@ PIP_TYPE?=cmt_top
...
 TODO_RE=".*CMT_TOP_[LR]_UPPER_T_PLLE2_CLK(IN1|IN2|FBIN)\.CMT_TOP_[LR]"
 
-ifneq (${XRAY_DATABASE}, zynq7)
 MAKETODO_FLAGS=--sides "r_upper_t,l_upper_t" --pip-type ${PIP_TYPE} --seg-type cmt_top --re $(TODO_RE)
-else
-MAKETODO_FLAGS=--sides "l_upper_t" --pip-type ${PIP_TYPE} --seg-type cmt_top --re $(TODO_RE)
-endif
 
-N = 100
+N = 200
```
litghost: This seems pretty high, are you sure this is needed?  Should there be better input stimulus rather than generating this many samples?

fuzzers/074-dump_all/zynq7_ignored_wires.txt
litghost: Please do sort this file

tmichalak:
> There is a suspicious diff from the A7 CI:

It's most probably due to the fact that I increased the number of specimen in `034-cmt-pll` to 200. So this can be actually the correct result? I mean I increasing the specimen should add a time penalty, but not influence the correctness of the results.

fuzzers/034-cmt-pll-pips/Makefile
tmichalak: I didn't notice much difference in the runtime of CI. But it's because it tooks around 5 hours already. I had to increase the number of specimen as otherwise I got conflicts between iterations as in Zynq7020 there is only one clock region with `CMT_TOP_R_UPPER` tiles compared to 3 clock regions in Artix.

tmichalak:
> There is also the broader question of Z7010 support, but I guess with #1196 we can add support for both Z7010 and Z7020?

Yes, the plan is to merge #1196 first, so that we don't break zyqn 7010 support in arch-defs. Next we should merge this PR and eventually proceed to adding other [zynq board support](https://github.com/SymbiFlow/symbiflow-arch-defs/pull/1260)

fuzzers/005-tilegrid/ps7_int/top.py
tmichalak: Since this fuzzer eventually returns the base address of a regular interconnect tile I don't think it's possible to distinguish it from all other interconnect tiles in the tilegrid. Hence, not hardcoding would require some messing with the basicdb tilegrid and then the `add_tdb.py` or `generate_full.py` scripts.

litghost: You misunderstand.  Is there a general relationship between the PS7 location and the interconnect choice you are making?  E.g. tile_at[PS7_tile.x + delta_x][PS7_tile.y + delta_y]?

fuzzers/034-cmt-pll-pips/Makefile
litghost: What I mean is that it doesn't make much sense for a N this high, it likely means that there is insufficient stimulus in the `top.py`, and rather than cranking the N higher, we should add better stimulus, either through `route_via` in the `generate.tcl` or modifying probabilities in `top.py`.

fuzzers/005-tilegrid/ps7_int/top.py
tmichalak: There is no ps7 tile from tilegrid's point of view. The ps7 tile is off-grid. It does not have a base address. Think of it as if you cut the fabric at the INT_L column and connected exposed wires to another chip which is the ps7 ARM SoC. It is only an interconnect and some pips that we know belong/are activated when the ps7 gets instantiated.

fuzzers/034-cmt-pll-pips/Makefile
tmichalak: I understand your point and I would also much rather improve the stimulus which seems satisfactory for artix and kintex, but isn't good enough solely for Zynq. Would it be an option at this point to set `N=200` only for Zynq and open an issue about this to deal with it a bit later?

acomodi:
I have rebased this PR and did the following:
- use xc7020 as main part for fuzzing;
- use xc7010 as extra part generation;

mithro:
The Zynq run took for 6 hr, 29 min, 3 sec (compared to O(4h) for both Artix and Kintex parts).

@litghost - Do we think that is reasonable or not?

litghost:
@tmichalak  / @acomodi  Please open a PR to add back the Artix 200T

PR #1163, opened 2019-12-03
Zynq7 EMIO minitest

mkurc-ant:
This is a minitest for the Zynq7 EMIO interface between PS (ARM) and PL (FPGA fabric).


kgugala:
LGTM

PR #1169, opened 2019-12-09
Support for required fasm features

mkurc-ant:
This is a draft PR for adding support of required fasm features that has to be set for a given part.

The idea is to add a JSON file describing them to the prjxray db root directory. The `Database` class of the prjxray API would provide the method `get_required_fasm_features(part)` which would return a set of fasm lines that need to be always appended to a fasm file for the given part.

The structure of the JSON file might look like this:
```
{
  "all": [
    "FEATURE_A",
    "FEATURE_B",
    "FEATURE_C"
  ],
  "xc7zXXXX": [
    "-FEATURE_A",
    "FEATURE_D"
  ],
  "xc7zYYYY": [
    "-FEATURE_B",
    "-FEATURE_C",
    "FEATURE_E"
  ],
}
```

The root element is a dict indexed by part names. Under each part name there is a list of fasm lines that need to be appended. The one special list `all` contains features common to all parts. Then, for each part there is a list of additional features to be appended (along with the `all` ones). If a feature name starts with the `-` sign then this one should be removed from the `all` set for that part.


mkurc-ant:
@litghost I have some questions:
 - How do we include a hand-written JSON file into the prjxray db?
 - Is it assumed that the database "root" folder contains data for a single part only?

mithro:
I suggest we just have FASM files that can be merged?

litghost:
> structure of the JSON

Rather that have a special key in the parts `dict`, make a root dictionary, with a key `parts` and a key `always_required`.

>     * How do we include a hand-written JSON file into the prjxray db?

Make a fuzzer that writes a JSON to the database folder, leave the JSON empty for parts that don't require anything?

>     * Is it assumed that the database "root" folder contains data for a single part only?

What do you mean by root?  `database/artix7`?  Or `database`?

mkurc-ant:
@litghost 
> What do you mean by root? database/artix7? Or database?

I mean `database/artix7`, `database/zynq7` etc. Could you write an example skeleton structure of the JSON that you are proposing?

litghost:
> Could you write an example skeleton structure of the JSON that you are proposing?

Rather than:
```
{
  "all": [
    "FEATURE_A",
    "FEATURE_B",
    "FEATURE_C"
  ],
  "xc7zXXXX": [
    "-FEATURE_A",
    "FEATURE_D"
  ],
  "xc7zYYYY": [
    "-FEATURE_B",
    "-FEATURE_C",
    "FEATURE_E"
  ],
}
```

do:

```
{
  "always_required": [
    "FEATURE_A",
    "FEATURE_B",
    "FEATURE_C"
  ],
  "per_part_required": {
    "xc7zXXXX": [
      "-FEATURE_A",
      "FEATURE_D"
    ],
    "xc7zYYYY": [
      "-FEATURE_B",
      "-FEATURE_C",
      "FEATURE_E"
    ]
  }
}
```

litghost:
> Is it assumed that the database "root" folder contains data for a single part only?

No.  In fact something we want to do soon is start to emit tilegrid/tileconn for multiple parts in the same family.

mkurc-ant:
I changed the approach a little bit. Instead of a JSON file for all parts there is one file per part named `{PART_NAME}_required_features.fasm`.

I also added the `--part` parameter to `fasm2frames` which value defaults to `XRAY_PART`. I'm not sure though whether any common required features should be emitted when the `--part` is not given.

prjxray/db.py
```diff
@@ -85,8 +87,33 @@ def __init__(self, db_root):
...
+                    for line in fp:
+                        line = line.strip()
+                        if len(line) > 0:
+                            features.append(line)
+
+                    per_part_features[part] = set(features)
+
         self.tile_types_obj = {}
 
+        # Find common required features for all supported parts
```
litghost: Why separate into common and part specific at all at this point?

utils/fasm2frames.py
```diff
@@ -165,7 +170,13 @@ def main():
         db_root_kwargs['required'] = False
         db_root_kwargs['default'] = os.path.join(database_dir, database)
 
+    def_part = os.getenv("XRAY_PART")
```
litghost: Copy the logic from above about making `--part` required if the env var is not defined.

mkurc-ant: Done.

prjxray/db.py
mkurc-ant: Fixed this. For now if the part is not explicitly given, no required features are returned.

prjxray/db.py
```diff
@@ -147,3 +161,16 @@ def get_tile_segbits(self, tile_type):
...
+        Assembles a set of required fasm features for given part. Returns a list
+        of fasm features.
+        """
+
+        # No required features in the db or part not given, return empty list
+        if self.required_features is None or part is None:
+            return set()
+
+        # Return list of part specific features
+        return self.required_features[part]
```
litghost: Maybe make this `self.required_features.get(part, set())`, so if a part has not required features there isn't a problem?

utils/fasm2frames.py
```diff
@@ -120,6 +121,10 @@ def check_for_pudc_b(set_feature):
             extra_features = fasm.parse_fasm_string(
                 '\n'.join(roi_j['required_features']))
 
+    # Get required extra features for the part
+    required_features = db.get_required_fasm_features(part)
```
litghost: Have you tested this logic when `get_required_fasm_features()` returns an empty set?

prjxray/db.py
mkurc-ant: Fixed.

utils/fasm2frames.py
mkurc-ant: Yes, I did. Actually the `test_fasm2frames.py` invokes fasm2frames on a fake db when there are no required features defined.

mkurc-ant:
Looks like Vivado crash on 045 for Kintex7:
```
...
pure virtual method called
terminate called without an active exception
make[3]: *** [vivado.ok] Error 134
make[2]: *** [build/7/specimen_015/OK] Error 2
make[2]: *** Waiting for unfinished jobs....
Parent process (pid 62143) has died. This helper process will now exit
make[1]: *** [run] Error 2
```

PR #1170, opened 2019-12-09
PS7 cell definition extraction

mkurc-ant:
This set of scripts allows to extract the PS7 cell definition from Vivado. All the PS7 pins are written to a JSON file which is pushed to the database.


litghost:
Should this be published into the zynq database?  If so, please update https://github.com/SymbiFlow/prjxray/blob/master/Makefile#L165

mkurc-ant:
I added generation of techmap that ties all unused input ports to GND.
@litghost I don't know whether any products of this tool should be published along with the db. Suggestions?

litghost:
> @litghost I don't know whether any products of this tool should be published along with the db. Suggestions?

We've been talking about this in other PR's, but shouldn't all generated outputs from this PR do to the database?

Issue #1151, opened 2019-11-22
Integrate LiteX minitests into Project X-Ray


tmichalak:
Add to `minitest/litex` all the LiteX related minitests.

 * `min` - Minimal configuration - just a CPU + uart targeting the Basys + Arty boards.

 Basys
  - [ ] CPU configured in "lite" mode? -- IE No DSPs or anything.
  - [ ] CPU configured with divider + multiplier + etc -- with DSPs enabled.

 Arty
  - [ ] CPU configured in "lite" mode? -- IE No DSPs or anything.
  - [ ] CPU configured with divider + multiplier + etc -- with DSPs enabled.

 * `base` - SoC config with DDR and Ethernet targeting the Arty.
  - [ ] CPU configured in Linux mode but with no DSPs.
  - [ ] CPU configured in Linux mode but with DSPs.

Add a README which describes how to generate these outputs or even add a script which generates them. 
We should also capture the git revisions of everything used to generate this output.

PR #1171, opened 2019-12-10
Integrate Base LiteX into Project X-Ray 

tmichalak:
This PR adds a minitest for the Base Litex design configuration (VexRiscv CPU in linux variant, DDRs and Ethernet).
The migen script used for generating the design has been added to the scripts directory.
The parent issue is #1151 

There are two variants: for Vivado only flow and for Yosys+Vivado flow. In order to run one of them enter the specific directory and run make.
Once the bitstream is generated and loaded to the board, we should see the test result on the terminal connected to one of the serial ports.


tmichalak:
This project is not the place for this design, we already have it in arch-defs and symbiflow-examples.

PR #1176, opened 2019-12-12
bitstream_writer: fix 7 series and US+ headers

kgugala:
This PR fixes bitstream header generation for series 7 and US+. Without this fix the bistreams could not be loaded from U-Boot or Linux on both Zynq-7000 and ZynqMP. The loading was failing on header check. The headers were fixed to follow the definitions from:

- Zynq7000: https://github.com/Xilinx/u-boot-xlnx/blob/master/drivers/fpga/zynqpl.c#L46
- ZynqMP: https://github.com/Xilinx/u-boot-xlnx/blob/master/drivers/fpga/zynqmppl.c#L18

Signed-off-by: Karol Gugala <kgugala@antmicro.com>


litghost:
Have you tested the new 7-series header with the existing working parts (e.g. basys3 and arty)?

kgugala:
> Have you tested the new 7-series header with the existing working parts (e.g. basys3 and arty)?

yes, it works fine. I tested it with xc3prog tool

PR #1184, opened 2019-12-18
037-iob-pips: removed IMUX22_1 from excluded re

acomodi:

This solves ILOGIC_CLKB pips that are missing and required for https://github.com/SymbiFlow/symbiflow-arch-defs/pull/1220


mkurc-ant:
@acomodi The fuzzer timed out, couldn't solve those two:
```
LIOI3.IOI_ILOGIC0_CLKB.IOI_IMUX22_1
RIOI3.IOI_ILOGIC0_CLKB.IOI_IMUX22_1
```

mithro:
I assume that is why it was excluded...
Should check the `git blame` output

acomodi:
I could solve those bits locally, but they were the only ones missing from the db. I think that, if the fuzzer wants to solve all of the other required features, the noise is too high to solve also IMUX22_1. I suppose a way to overcome this problem is to separate the fuzzer and solve the IMUXes features for the clock in that one only.

mithro:
Sounds like that could be a good option.

tmichalak:
Created a PR with the 2 iob-pips fuzzers instead of one - #1185

tmichalak:
Superseded by #1185

PR #1186, opened 2019-12-20
Grouping of IN_TERM features

mkurc-ant:
Added grouping of IN_TERM features so they can be decoded unambiguously by bit2fasm.

The utility `group.py` is a slightly modified version of `fuzzers/034-cmt-pll-pips/fixup_and_group.py`


fuzzers/030-iob/group.py
```diff
@@ -0,0 +1,220 @@
+#!/usr/bin/env python3
```
litghost: Does this file belong in `030-iob` versus `<root>/utils`?

fuzzers/030-iob/Makefile
```diff
@@ -13,7 +13,8 @@ build/segbits_xiob33.rdb: $(SPECIMENS_OK)
 
 build/segbits_xiob33.db: build/segbits_xiob33.rdb process_rdb.py bits.dbf
 	python3 process_rdb.py build/segbits_xiob33.rdb > build/segbits_xiob33_processed.rdb
-	${XRAY_DBFIXUP} --db-root build --zero-db bits.dbf --seg-fn-in build/segbits_xiob33_processed.rdb --seg-fn-out $@
+	python3 group.py -g tag_groups.txt -i build/segbits_xiob33_processed.rdb -o build/segbits_xiob33_processed.rdb2
+	${XRAY_DBFIXUP} --db-root build --zero-db bits.dbf --seg-fn-in build/segbits_xiob33_processed.rdb2 --seg-fn-out $@
```
litghost: Should the feature set of `group.py` be part of `XRAY_DBFIXUP`?

mithro:
What is the status of this pull request?

fuzzers/030-iob/Makefile
mkurc-ant: I moved the `group.py` to the `utils` folder. Do you want this tool to be a part of the dbfixup script? At first I wanted to be that way but I found it difficult to implement the functionality in the dbfixup script architecture.

litghost: Is there a way to have it be a part of the dbfixup script, but only in the lightest sense?  Basically don't try to integrate it complete, but make it an isolated pass within the script?

The idea is future fuzzers might use either the zero-db or the group style?  Thoughts?

mkurc-ant: Hmmm, Ok. I can make it as an additional pass to the db_fixup. So my idea is to keep `group.py` as a separate script but call it from within the db_fixup.

litghost:
@mkurc-ant What is the status of this PR?  The requested change was fairly small?

mkurc-ant:
@litghost I'm integrating the `group.py` with the `dbfixup.py`. Give me time until tomorrow for that.

mkurc-ant:
@litghost I ported the functionality of the `group.py` to `dbfixup.py`. I left the previously written `group.py` script.

PR #1196, opened 2020-01-10
Allow generation of multiple-part db

acomodi:

This PR is an initial WIP to allow the generation of multiple xc7 parts definitions for each family.

The idea is to have the part-dependent fuzzers acting differently from the rest of the fuzzers, and so run them against all the selected parts for each family.

In this way we can have a multiple-parts support for each FPGA family to be used in symbiflow-arch-defs


litghost:
In words, what is your plan for this PR?  I'm not sure the direction it is going is the right one, but I don't understand what you are trying to accomplish yet either.

acomodi:
@litghost the basic idea is to have prjxray to geberate part-dependent files (yaml, tilegrid, tileconn, etc) for each part specified in the settings.sh.

To do so, each part-dependent fuzzer is run against all the parts that are specified, and each output is put in the specific part directory.

The implementation I am following is similar to the general Makefile in the top directory of the fuzzers.
I have created a function to be called with the correct part name, that contains the targets to generate the various intermediate files (specimens, segdata, etc.).

The fuzzers that need modifications are:
- 001-yaml
- 005-tilegrid
- 072-ordered-wires
- 074-tileconn
- 075-pins

acomodi:
I have modified my initial implementation path.
With the new code, the additional part-specific files are generated during the `db-extras` step.

This requires lesser changes to the various fuzzers, but the addition of the settings of every single part.

acomodi:
IMO, to speed up CI, we could set up another build that deals with all the `db-extras`, so that they can start building alongside with all the rest of the fuzzers.

litghost:
> I have modified my initial implementation path.
> With the new code, the additional part-specific files are generated during the `db-extras` step.
> 
> This requires lesser changes to the various fuzzers, but the addition of the settings of every single part.

Much better, thanks.

acomodi:
> let's see how much extra time will be needed for Artix to finish now it has the 200T tilegrid to do.

It has been running for almost 20 hours now... I suppose that all the time is spent on the 074-dump_all fuzzer. I am testing locally as well.

I am also trying to get all the part-specific generated files to be located in the respective directories. This is done so that we can avoid re-running the xc50at part-only fuzzers.

The local change I am about to push changes the location of the tilegrid.json and tileconn.json (needed by all the various script, namely, fasm2frames, fasm2pips, etc.).

Makefile
```diff
@@ -139,9 +139,31 @@ $(foreach DB,$(DATABASES),$(eval $(call database,$(DB))))
...
+KINTEX_PARTS=kintex70t
+
+XRAY_PARTS=${ARTIX_PARTS} ${ZYNQ_PARTS} ${KINTEX_PARTS}
+
+define multiple-parts
+
+# $(1): PART to be used
+
+db-part-only-$(1):
+	+source settings/$(1).sh && $(MAKE) -C fuzzers part_only
```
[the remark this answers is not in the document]
acomodi: Done

litghost:
> It has been running for almost 20 hours now

Is the time spent in Vivado, or in the post processing?

prjxray/util.py
```diff
@@ -73,7 +80,7 @@ def slice_xy():
 def get_roi():
     from .db import Database
     (x1, x2), (y1, y2) = roi_xy()
-    db = Database(get_db_root())
+    db = Database(get_db_root(), get_part())
```
tmichalak: Since get_part() is basically only getting an environment variable why can't this be done in the Database class. This would save many calls of this function for example in the `top.py` scripts in many fuzzers

acomodi: The fact is that the Database is accessed also outside of prjxray, where no env variable is set (e.g. in arch-defs). Therefore, there is the need to get it as input to create the Database object.

acomodi: Of course I could avoid adding that to the prjxray function calls (as the Database init function could look for the env variable, and if it is not present, and the argument is also `None`, than it could throw an error).

tmichalak: I know @litghost is not a big fan of environment variables so maybe it's better to leave it the way you have it currently. I just see that like with any API changes, the number of fuzzers that require adjustment is quite big. Other than that I don't see any problem with having to call `get_part()`.

fuzzers/005-tilegrid/generate_full.py
```diff
@@ -395,15 +395,10 @@ def propagate_IOI_Y9(database, tiles_by_grid):
...
-    arch = os.getenv('XRAY_DATABASE')
-    if arch in 'artix7':
-        tiles = ['RIOI3_X43Y9', 'LIOI3_X0Y9']
-    elif arch in 'kintex7':
-        tiles = ['LIOI3_X0Y9']
-    elif arch in 'zynq7':
-        tiles = ['RIOI3_X31Y9']
-    else:
-        assert False, "Unsupported architecture"
+    ioi_tiles = os.getenv('XRAY_IOI3_TILES')
```
litghost: This should be changed to just keep of the pin function (e.g. avoid the PUDC), rather than an env var.

fuzzers/005-tilegrid/generate_tiles.tcl
```diff
@@ -3,6 +3,7 @@ source "$::env(FUZDIR)/util.tcl"
 proc write_tiles_txt {} {
     # Get all tiles, ie not just the selected LUTs
     set tiles [get_tiles]
+    set not_allowed_sites [get_sites -of_objects [get_pblocks exclude_roi]]
```
litghost: What is the purpose of the exclude_roi?  I haven't seen it before.

fuzzers/072-ordered_wires/Makefile
```diff
@@ -2,8 +2,8 @@
 N := 1
 SPECIMENS := $(addprefix build/specimen_,$(shell seq -f '%03.0f' $(N)))
 SPECIMENS_OK := $(addsuffix /OK,$(SPECIMENS))
-MAX_VIVADO_PROCESS ?= 4
-MAX_PIPS_INSTANCE ?= 340000
+MAX_VIVADO_PROCESS ?= 8
+MAX_PIPS_INSTANCE ?= 680000
```
litghost: Are you sure about increasing this?  Vivado has some kind of memory leak and/or slowdown as this number gets higher.

fuzzers/072-ordered_wires/run_fuzzer.py
```diff
@@ -20,6 +22,25 @@ def start_pips(argList):
         str(blockID) + " " + str(start) + " " + str(stop),
         shell=True)
 
+    uphill_wires = "wires/uphill_wires_{}.txt".format(blockID)
```
litghost: Can you explain the reasoning behind this?

fuzzers/072-ordered_wires/run_fuzzer.py
```diff
@@ -3,13 +3,15 @@
...
 import signal
-from multiprocessing import Pool
+from multiprocessing import Pool, Lock
 from itertools import chain
 import argparse
 
 # Can be used to redirect vivado tons of output
 # stdout=DEVNULL in subprocess.check_call
 
+MP_LOCK = Lock()
```
litghost: Probably better to pass the lock as part of the arglist than to use a global.

fuzzers/074-dump_all/Makefile
```diff
@@ -2,9 +2,9 @@
 N := 1
 SPECIMENS := $(addprefix build/specimen_,$(shell seq -f '%03.0f' $(N)))
 SPECIMENS_OK := $(addsuffix /OK,$(SPECIMENS))
-MAX_VIVADO_PROCESS ?= 4
-MAX_TILES_INSTANCE ?= 300
-MAX_NODES_INSTANCE ?= 30000
+MAX_VIVADO_PROCESS ?= 8
+MAX_TILES_INSTANCE ?= 600
+MAX_NODES_INSTANCE ?= 60000
```
litghost: Are you sure about increasing this? Vivado has some kind of memory leak and/or slowdown as this number gets higher.

fuzzers/074-dump_all/jobnodes.tcl
```diff
@@ -13,10 +13,24 @@ set root_fp [open "root_node_${blocknb}.csv" w]
 
 set nodes [get_nodes]
 
-for {set j $start } { $j < $stop } { incr j } {
+create_pblock exclude_roi
```
litghost: This exclude_roi seems problematic.  Why is it being done?

fuzzers/Makefile
```diff
@@ -127,11 +127,41 @@ $(eval $(call fuzzer,100-dsp-mskpat,005-tilegrid))
...
+# an roi harness.
+roi_only:
 	+$(MAKE) -C 001-part-yaml run
 	+$(MAKE) -C 075-pins run
 
-.PHONY: all clean clean_fuzzers clean_logs quick part_only
+# Part only targets
+# -----------------
+
+part_only_yaml:
```
litghost: Why is this written to use recursive make rather than targets with dependencies?

minitests/timing/Makefile
```diff
@@ -27,7 +27,6 @@ build/$(2)_$(1)/timing_$(2)_$(1).json5: build/.touch runme.tcl $(3)
 build/timing_$(2)_$(1).xlsx: build/$(2)_$(1)/timing_$(2)_$(1).json create_timing_worksheet_db.py
 	python3 create_timing_worksheet_db.py \
 		--timing_json build/$(2)_$(1)/timing_$(2)_$(1).json \
-		--db_root ${XRAY_DATABASE_DIR}/${XRAY_DATABASE} \
```
litghost: No, add both the `--db_root` argument and `--part` argument

prjxray/util.py
litghost: @acomodi made the right change, and the amount of churn within prjxray isn't that concerning.

fuzzers/005-tilegrid/generate_tiles.tcl
acomodi: The problem is that, for some parts (200T, for instance) there are some CLBs that cannot be used (they appear to be locked), and would generate errors in Vivado. With the exclude ROI, if set, we can exclude the problematic tiles from the tilegrid.

fuzzers/072-ordered_wires/Makefile
acomodi: Actually it is a leftover, I tried to increase this, as locally building the 200T took a huge amount of time (there are sth like 80 milion pips)

fuzzers/005-tilegrid/generate_tiles.tcl
litghost: But can the interconnect around them be used?  If so, we need to include the INT tiles in the tilegrid and tileconn.

Can you post the error from Vivado for reference?

If we keep the excluded ROI idea, there needs to be a large comment somewhere explaining why this happens.

fuzzers/072-ordered_wires/Makefile
litghost: Something to note is that Kokoro sets `MAX_VIVADO_PROCESS`, https://github.com/SymbiFlow/prjxray/blob/master/.github/kokoro/db-full.sh#L55

fuzzers/072-ordered_wires/run_fuzzer.py
acomodi: Here there is a disk utilization optimization only. The fact is that, for big parts, this fuzzer leaves a lot of temporary data. With this change, each process spawned by the pool, updates the final files and deletes the intermediate files, saving disk space. 

Mainly it is done to be able to run this fuzzer locally if the machine does not have enough disk space

fuzzers/074-dump_all/jobnodes.tcl
acomodi: This is a follow up to the usage of exclude ROI in the 005 fuzzer. Tileconn needs to avoid building the tilegrid with the locked tiles as it would conflict with the original tilegrid

fuzzers/072-ordered_wires/run_fuzzer.py
litghost: Only, add a comment to that effect.

acomodi: Done

fuzzers/072-ordered_wires/run_fuzzer.py
acomodi: So, the lock cannot be zipped  with the other arguments. Instead, I have kept this variable global, but I have encapsulated it in an init function that is passed at the creation time of the Pool of workers.

fuzzers/Makefile
acomodi: Fixed

minitests/timing/Makefile
acomodi: Fixed

fuzzers/005-tilegrid/generate_full.py
acomodi: The fact is that the PUDC IOB tiles are not connected to the ones that are expanded by one frame in the `propagate_IOI_Y9` function.

In the `pin_func.txt` file produced during the tiles extraction reports that the only IOB with the PUDC function is `IOB_X0Y44` (for artix 50T), while the IOI tiles with the extra frame address are `X0Y9` and `X43Y9`

fuzzers/072-ordered_wires/Makefile
acomodi: Fixed, restored as it was previously

fuzzers/074-dump_all/Makefile
acomodi: Fixed

acomodi:
```
FATAL ERROR: CALL_AND_RETRY_LAST Allocation failed - JavaScript heap out of memory
 1: node::Abort() [node]
 2: 0x13740dc [node]
 3: v8::Utils::ReportOOMFailure(char const*, bool) [node]
 4: v8::internal::V8::FatalProcessOutOfMemory(char const*, bool) [node]
 5: v8::internal::Factory::NewRawOneByteString(int, v8::internal::PretenureFlag) [node]
 6: v8::internal::Factory::NewStringFromOneByte(v8::internal::Vector<unsigned char const>, v8::internal::PretenureFlag) [node]
 7: v8::internal::Factory::NumberToString(v8::internal::Handle<v8::internal::Object>, bool) [node]
 8: v8::internal::Runtime_NumberToString(int, v8::internal::Object**, v8::internal::Isolate*) [node]
 9: 0x1dc75d4840dd
github/symbiflow-prjxray-presubmit-db-artix7/.github/kokoro/db-full.sh: line 161: 20166 Aborted                 diff2html --summary=open --file diff.html --format html -- --irreversible-delete --find-renames --find-copies --ignore-all-space origin/master
```

Apparently there was an issue while calculating the diff.
I would suppose that the addition of the 200T makes the diff pretty huge, therefore resulting in `JavaScript heap out of memory`.

mithro:
@acomodi - Ignore the diff error.

mithro:
For the diff problem, I opened https://github.com/SymbiFlow/prjxray/pull/1205

acomodi:
@mithro @litghost The extra-part for artix (200T) seems to be taking a huge amount of time to build.
Being it independent from the full build of the db, I am considering whether it is worth setting up an additional CI run (one for each family) to build the extra parts. It will probably still take a high amount of time. We can verify the exact amount as soon as the current run completes.

The issue is in the 074-dump_all fuzzer, which does not scale linearly. I need to analyze if there are possible optimizations to be made on the scripts, even though great part of the time is spent dumping the data.

litghost:
> @mithro @litghost The extra-part for artix (200T) seems to be taking a huge amount of time to build.
> Being it independent from the full build of the db, I am considering whether it is worth setting up an additional CI run (one for each family) to build the extra parts. It will probably still take a high amount of time. We can verify the exact amount as soon as the current run completes.
> 
> The issue is in the 074-dump_all fuzzer, which does not scale linearly. I need to analyze if there are possible optimizations to be made on the scripts, even though great part of the time is spent dumping the data.

I've asked this before, which part is taking the time?  The Vivado part?  Or the reduction part?

acomodi:
@litghost Missed that. Anyway, I need to run it locally again to get the exact run-time information. From what I can tell for now, both of the vivado and reduction parts should take at least a couple of hours each.

mithro:
@acomodi The diff problem should be solved now.

acomodi:
@mithro Perfect, I will rebase. Thanks

@litghost Got the run-time info:
```
Vivado start time:
01:14:28 CET
Vivado end time and Reduction start time:
07:22:55 CET
Reduction end time:
09:44:51 CET
```

This was run locally, and the we can confirm that the Vivado part is occupying much of the run-time (~6 hours) while the reduction time is ~2 hours.

For now I will disable the generation of the extra parts, so that we do not let this PR hang, therefore letting also https://github.com/SymbiFlow/prjxray/pull/1162 land.

litghost:
@acomodi Can you open a PR that re-enables the extra parts?

acomodi:
Sure, I am also preparing a PR for arch-defs that enables the usage of the new structure of prjxray

PR #1215, opened 2020-01-29
fuzzers: Add support for KiB, MiB and GiB units

tmichalak:
The `run_fuzzers.py` script assumes that the memory reported by `free` is in GB, MB or kB.
This PR adds GiB, MiB and KiB support.

### DIFF
--- a/fuzzers/run_fuzzer.py
+++ b/fuzzers/run_fuzzer.py
@@ -287,6 +287,9 @@ def mem_convert(s):
     62.0
     """
     units = {
+        'Gi': pow(2, 30),
+        'Mi': pow(2, 20),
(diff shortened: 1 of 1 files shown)


fuzzers/run_fuzzer.py
```diff
@@ -287,6 +287,9 @@ def mem_convert(s):
     62.0
     """
     units = {
+        'Gi': pow(2, 30),
+        'Mi': pow(2, 20),
+        'Ki': pow(2, 10),
```
mithro: Minor nit - feel free not to change.

I find the `pow(2, 10)` form a bit weird. I would have written `1024`, `1024*1024` and `1024*1024*1024` personally.

PR #1218, opened 2020-02-03
Added UART DDR minitest

acomodi:

This PR adds the minimal UART-DDR test, produced using https://github.com/enjoy-digital/arty_ddr3_prjxray


minitests/litex/uart_ddr/arty/src.yosys/missing_bit_report.py
```diff
@@ -0,0 +1,81 @@
+""" Generates a missing feature/bit report for LiteX design.
```
kgugala: I think this one should be implemented as more generic tool. Leaving it here will shortly end up in notworking tool nobody uses.

minitests/litex/uart_ddr/arty/README.md
```diff
@@ -0,0 +1,90 @@
+# LiteX UART DDR minitest
```
kgugala: Please add info about the exact versions of Litex repos (Litex and Litedram) used to generate this minitest

minitests/litex/uart_ddr/arty/scripts/Makefile
```diff
@@ -0,0 +1,5 @@
+litex/litex/tools/litex_client.py:
+	git clone https://github.com/enjoy-digital/litex.git
```
kgugala: should we point to exact commit?

litghost:
Some something we did in the VTR fork we should start doing here is putting generated files in a `gen` folder to distiguish them.  Sounds good?

minitests/litex/uart_ddr/arty/src.yosys/missing_bit_report.py
acomodi: Hmm, I believe this is not needed, it is actually a leftover from my local branch.

This script should be relative to only HCLK_IOI and IOI3 tiles unknown features, which are solved by now.

minitests/litex/uart_ddr/arty/README.md
acomodi: Done

minitests/litex/uart_ddr/arty/scripts/Makefile
acomodi: I think it is not strictly necessary, but I have added a `git checkout` to the current commit, to be sure.

acomodi:
> Some something we did in the VTR fork we should start doing here is putting generated files in a `gen` folder to distiguish them. Sounds good?

@litghost You mean the files under the `verilog` directory? Because those are the ones that have already been generated and effectively used for the synthesys step.
I believe also the top.xdc in the `src.yosys` and `src.vivado` directories.

minitests/litex/uart_ddr/arty/src.vivado/Makefile
```diff
@@ -0,0 +1,33 @@
...
+	@rm -f *.bit
+	@rm -f *.bin
+	@rm -f *.bits
+	@rm -f *.fasm
+	@rm -f *.log
+	@rm -rf build
+
+.PHONY: all clean
+
+top.bit: $(VIVADO) $(SOURCES) top.xdc top.tcl
```
kgugala: dependency of this target is `top.tcl` and in the target's command you use `../top.tcl`

acomodi: I've checked, the command first gets to the build directory, therefore this should be fine

PR #1220, opened 2020-02-04
Optimize 074 dump all

acomodi:
This PR tries to optimize fuzzer `074-dump_all` by moving timing model calculations after the tile have been reduced.

### DIFF
--- a/fuzzers/074-dump_all/jobtiles.tcl
+++ b/fuzzers/074-dump_all/jobtiles.tcl
@@ -23,11 +23,6 @@ foreach roi "$::env(XRAY_EXCLUDE_ROI_TILEGRID)" {
 set not_allowed_sites [get_sites -of_objects [get_pblocks exclude_roi]]
 set not_allowed_tiles [get_tiles -of_objects $not_allowed_sites]
 
-# Convert DRIVE from ??? units to 10^(-3 to -6) Ohms
-set MAGIC 0.6875
-
-set speed_model_index_map [dict create]
-
 for {set j $start } { $j < $stop } { incr j } {
 
     set tile [lindex $tiles $j]
--- a/fuzzers/074-dump_all/jobtiles.tcl
+++ b/fuzzers/074-dump_all/jobtiles.tcl
@@ -27,26 +27,6 @@ set not_allowed_tiles [get_tiles -of_objects $not_allowed_sites]
 set MAGIC 0.6875
 
 set speed_model_index_map [dict create]
-set speed_model_name_map [dict create]
-
-proc lookup_speed_model_index {index} {
-    upvar speed_model_index_map speed_model_index_map
-    if { ![dict exists $speed_model_index_map $index] } {
-        set model [get_speed_models -filter "SPEED_INDEX == $index"]
-        dict set speed_model_index_map $index $model
-    }
-    return [dict get $speed_model_index_map $index]
-}
-
-proc lookup_speed_model_name {name} {
-    upvar speed_model_name_map speed_model_name_map
-    if { ![dict exists $speed_model_name_map $name] } {
-        set model [get_speed_models -filter "NAME == $name"]
-        dict set speed_model_name_map $name $model
-    }
-
-    return [dict get $speed_model_name_map $name]
-}
 
 for {set j $start } { $j < $stop } { incr j } {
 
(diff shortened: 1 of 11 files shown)


fuzzers/074-dump_all/jobtiles.tcl
```diff
@@ -27,26 +27,6 @@ set not_allowed_tiles [get_tiles -of_objects $not_allowed_sites]
 set MAGIC 0.6875
 
 set speed_model_index_map [dict create]
```
litghost: Lines 27-29 should also be deleted?

litghost: Comment on line 26 didn't make it to the new file

acomodi: Done

acomodi: @litghost Just to know, where does this magic value originally comes from?

litghost: Empirical measurement, hence why I called it "MAGIC".  It is a constant that seems to generate valid values, but I don't really know what it is.  Hence the comment about converting from ??? to Ohm like values.

litghost:
Please remove `speed_model_index` once merged:
```
@@ -16,6 +16,7 @@
             "is_directional": "1",
             "is_pass_transistor": 0,
             "is_pseudo": "0",
+            "speed_model_index": "472",
             "src_to_dst": {
                 "delay": [
                     "0.057",

```

acomodi:
> Please remove `speed_model_index` once merged:

Done

litghost:
Did you change how the JSON were written:
```
--- a/artix7/tile_type_BRAM_L.json
+++ b/artix7/tile_type_BRAM_L.json
@@ -19721,639 +19721,767 @@
                 "ADDRARDADDRL0": {
                     "cap": "0.000",
                     "delay": [
-                        "0.000",
-                        "0.000",
-                        "0.000",
-                        "0.000"
+                        [
+                            "0.000",
+                            "0.000",
+                            "0.000",
+                            "0.000"
+                        ]
                     ],
                     "wire": "BRAM_FIFO36_ADDRARDADDRL0"
```

Please fix, I'd like to avoid diffs where no exist.

litghost:
Also should we add the A200T back to this PR to see how long it is taking?

litghost:
> Did you change how the JSON were written:
> 
> ```
> --- a/artix7/tile_type_BRAM_L.json
> +++ b/artix7/tile_type_BRAM_L.json
> @@ -19721,639 +19721,767 @@
>                  "ADDRARDADDRL0": {
>                      "cap": "0.000",
>                      "delay": [
> -                        "0.000",
> -                        "0.000",
> -                        "0.000",
> -                        "0.000"
> +                        [
> +                            "0.000",
> +                            "0.000",
> +                            "0.000",
> +                            "0.000"
> +                        ]
>                      ],
>                      "wire": "BRAM_FIFO36_ADDRARDADDRL0"
> ```
> 
> Please fix, I'd like to avoid diffs where no exist.

What this data is actually different, [0,0,0,0] is now [[0,0,0,0]]!  This should be fixed.

acomodi:
> What a minute, this data is actually different. [0,0,0,0] is now [[0,0,0,0]]! This should be fixed.

Done, there was a bug in the back-annotation of site_pins

> Also should we add the A200T back to this PR to see how long it is taking?

Done, there is something that I am thinking that may improve further run-time.
I noticed that both in 072 and 074, each `job.tcl` does an expensive task, namely getting all pips or nodes:
https://github.com/SymbiFlow/prjxray/blob/bdb876caec1385953821bc8b5f28ac395c944ebe/fuzzers/074-dump_all/jobnodes.tcl#L14 https://github.com/SymbiFlow/prjxray/blob/bdb876caec1385953821bc8b5f28ac395c944ebe/fuzzers/072-ordered_wires/job.tcl#L12
For large parts, like the 200T, this operation takes on my local machine ~30 seconds, which, multiplied by all the jobs it rounds up to ~1 hour each. Maybe this could be avoided by extracting those info once and split them among different jobs.

acomodi:
Looks like we have a nice performance increase.
CI run-times of the `074-dump_all`'s Vivado on artix7 50T build:
- with this PR: `~25m49s`
- previous build without these changes: `~1h24m`

Hopefully this scales up for bigger parts.

acomodi:
There seem to be an issue with back annotation of wires capacitance and resistance values, as there are differences reported in the kintex diff.

I need to fix that

litghost:
So something I've noticed is that you don't start the 200T build until after the entire A7 build is done.  Can this be fixed?

litghost:
Another thing I've noticed is we have memory and CPU to spare:
```
2020-02-07T21:15:29.076241 - 072-ordered_wires         - 3h25m:                              `-python3
2020-02-07T21:15:39.108198 - 072-ordered_wires         - 3h25m: Still running (1m:3.97%, 5m:4.22%, 15m:4.25% Mem:5.6Gi used, 347.0Gi free).
2020-02-07T21:15:39.108198 - 072-ordered_wires         - 3h25m: make---make---bash---python3-+-3*[python3---sh---vivado.sh---vivado---loader---vivado]
```

We could consider increasing the number of parallel vivado jobs to run, assuming they are not disk limited.

acomodi:
> So something I've noticed is that you don't start the 200T build until after the entire A7 build is done. Can this be fixed?

Exactly, the main problem is that the 200T cannot start until the part-independent fuzzers have finished for the 50T. This because temporary files could be overwritten during the process.

For instance, if 50T 074-dump_all fuzzer is running and we start the 072-ordered_wires for the 200T, than the `uphill_wires` and `downhill_wires` would get overwritten, generating issues with the 50T run.

I suppose we can fix this by enabling parallel CIs that only build the extra-parts.

acomodi:
I have gathered some data when running the 200T with this PR locally:
```
YAML start time:                11:23:00
----------------------------------------
TILEGRID start time:            11:24:07
----------------------------------------
ORDERED WIRES start time:       12:07:23
----------------------------------------
GET_COUNTS start time:          14:33:29
----------------------------------------
TILECONN timing: 
    Vivado start time:          14:34:48
    Reduction start time:       17:25:42
    Reduction end time:         19:28:37
----------------------------------------
PINS start time:                19:28:38
----------------------------------------
Flow end time:                  19:28:58

Full flow time:                 8h 5m 58s
```

The flow ended successfully, therefore I am unsure why the CI is hanging on the 200T.

There seems also to be a zynq7 instability of some bits

litghost:
> The flow ended successfully, therefore I am unsure why the CI is hanging on the 200T.

```
074-dump_all/build/specimen_001/INT_R_X99Y234/.NW6BEG2.json5.hiiHLc" failed: No space left on device (28)
```

litghost:
It actually succeeds:
```
2020-02-07T01:26:21.126485 - 074-dump_all              - 7h19m: 
2020-02-07T01:26:21.126485 - 074-dump_all              - 7h19m: --------------------------------------------------------------------------
2020-02-07T01:26:21.126485 - 074-dump_all              - 7h19m: Succeeded! @ 2020-02-07T01:26:20.969799
```

But then fails when trying to collect results from the job.  The solution is to remove the intermediate outputs once done!

litghost:
I noticed another issue, I believe you are running with the default MAX_VIVADO_PROCESS (4) rather than a higher number, see https://github.com/SymbiFlow/prjxray/pull/1227/commits/b0a7f8892ef4b2745b6bdb2fccc4b243c07df619

acomodi:
@litghost yep missed that, that'll fix the issue, let's see how much time will it take now

litghost:
Superceded by #1227

PR #1226, opened 2020-02-10
Fixing the sorting of database files

mithro:
- Rework how the `json` files are sorted (numbers are treated as numerics).
 - Sort `csv` and `txt` files.
 - Sort `segbits.*origin_info.db` files.
 - Sort the grid file.

How this changes the output can be seen in https://github.com/SymbiFlow/prjxray-db/pull/11/files

PR #1230, opened 2020-02-12
Fixing the sorting of database files

litghost:
Copy of #1226, remove use of xjson from `074`
>  - Rework how the `json` files are sorted (numbers are treated as numerics).
>  - Sort `csv` and `txt` files.
>  - Sort `segbits.*origin_info.db` files.
>  - Sort the grid file.
> 
> How this changes the output can be seen in https://github.com/SymbiFlow/prjxray-db/pull/11/files

### DIFF
--- a/fuzzers/074-dump_all/create_node_tree.py
+++ b/fuzzers/074-dump_all/create_node_tree.py
@@ -5,8 +5,7 @@ import os.path
 import prjxray.lib
 import pickle
 import collections
-
-from utils import xjson
+import json
 
 
 def build_node_index(fname):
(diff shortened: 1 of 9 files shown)


mithro:
@litghost Yeah - I was thinking that xjson might be the issue on that script.

fuzzers/074-dump_all/create_node_tree.py
```diff
@@ -273,7 +272,7 @@ def main():
 
     print('{} Writing node tree'.format(datetime.datetime.now()))
     with open(os.path.join(args.output_dir, 'node_tree.json'), 'w') as f:
-        xjson.pprint(f, nodes)
+        json.dump(nodes, f, indent=2, sort_keys=True)
```
mithro: Since we are going to sort + indent things at a later date, should we just skip indent / sort_keys?

litghost: It's still convenient for debugging to have the `indent=2`, and the sort_keys should also be fast.

litghost:
@mithro looks like we are in business

mithro:
I have pushed an updated database based on this change at https://github.com/SymbiFlow/prjxray-db/commit/db494e9d5d71edd3c8b6558f7eec12e163aef2d8

PR #1231, opened 2020-02-12
Add xc7a35tftg256 package to extra output.

litghost:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -160,6 +160,8 @@ endef
 $(foreach PART,$(XRAY_PARTS),$(eval $(call multiple-parts,$(PART))))
 
 db-extras-artix7-parts: $(addprefix db-part-only-,$(ARTIX_PARTS))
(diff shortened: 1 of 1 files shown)


ssb83:
the above "Database - Artix 7" build failed due to:

`set_property -dict "PACKAGE_PIN $::env(XRAY_PIN_00) IOSTANDARD LVCMOS33" [get_ports clk]
ERROR: [Common 17-69] Command failed: 'E22' is not a valid site or package pin name.`

which is because XRAY_PIN_00 .. XRAY_PIN_06  in settings/artix7.sh mapped to pins that are not bonded in the ftg256 package.

litghost:
> the above "Database - Artix 7" build failed due to:
> 
> `set_property -dict "PACKAGE_PIN $::env(XRAY_PIN_00) IOSTANDARD LVCMOS33" [get_ports clk] ERROR: [Common 17-69] Command failed: 'E22' is not a valid site or package pin name.`
> 
> which is because XRAY_PIN_00 .. XRAY_PIN_06 in settings/artix7.sh mapped to pins that are not bonded in the ftg256 package.

Fixed

PR #1235, opened 2020-02-13
Fix some bugs present timing analysis script.

litghost:
These fixes were required to allow initial analysis of https://github.com/SymbiFlow/symbiflow-arch-defs/files/4200259/timing.zip


minitests/timing/runme.tcl
```diff
@@ -1,95 +1,4 @@
...
-
-        puts $fp "},"
-    }
-
-    puts $fp "\]"
-    close $fp
-}
+source $::env(XRAY_UTILS_DIR)/write_timing_info.tcl
 
 proc create_design {design_name sig_mask verilogs} {
```
acomodi: Maybe it would be useful to include also additional tcl files (such as the one generated in arch-defs vivado target) to use this runme.tcl more generically, to produce the json5 timings file of an implemented design.

litghost: In the case of implemented design, `runme.tcl` isn't required, just:
```
open_checkpoint <file>
source ${PATH_TO_UTILS}/write_timing_info.tcl
write_timing_info timing.json5
```

So I don't believe we need to modify `runme.tcl`

PR #1237, opened 2020-02-14
Do not sort already ordered data.

litghost:

PR #1238, opened 2020-02-14
Remap some timing

litghost:
This is an attempt at solving https://github.com/SymbiFlow/symbiflow-arch-defs/issues/1301

This PR builds onto of #1237, only new commit is e6836c68b45748023bd86a5a5e88bf73736d727a


litghost:
Good news is this fix appears to resolve the timing error in https://github.com/SymbiFlow/symbiflow-arch-defs/issues/1301, except that the long wire capacitance is wrong.  I'll make an update to `create_timing_worksheet_db.py` and then examine if the rr graph import needs the same change.

fuzzers/074-dump_all/jobtiles.tcl
```diff
@@ -48,6 +48,22 @@ proc lookup_speed_model_name {name} {
     return [dict get $speed_model_name_map $name]
 }
 
+# For BSW_INT_LONG_MUX, use the model from BSW_INT_HLONG_MUX.
```
mithro: Could you add a comment about why there isn't a `BSW_INT_LONG_MUX` model and thus why we have to use the `BSW_INT_HLONG_MUX` model?

utils/create_timing_worksheet_db.py
```diff
@@ -125,7 +125,13 @@ def __init__(self, net):
 
     def extend_rc_tree(self, ws, current_rc_root, timing_lookup, node):
         rc_elements = []
-        for wire in node['wires']:
+
+        # LV nodes have a workaround applied because of a working in the
```
mithro: I don't understand this comment?
> LV nodes have a workaround applied because **of a working** in the pip timing.

litghost: Ya, fixed.

fuzzers/074-dump_all/jobtiles.tcl
litghost: Fixed

Issue #1236, opened 2020-02-14
New sort_db is broken


litghost:
```
  File "../../utils/sort_db.py", line 463, in <module>
    sys.exit(main(sys.argv))
  File "../../utils/sort_db.py", line 458, in main
    sort_dir('.')
  File "../../utils/sort_db.py", line 450, in sort_dir
    sort_file(n)
  File "../../utils/sort_db.py", line 422, in sort_file
    x = sort_json(n)
  File "../../utils/sort_db.py", line 385, in sort_json
    xjson.pprint(f, d)
  File "/tmpfs/src/github/symbiflow-prjxray-presubmit-db-artix7/utils/xjson.py", line 125, in pprint
    data = sort(data)
  File "/tmpfs/src/github/symbiflow-prjxray-presubmit-db-artix7/utils/xjson.py", line 117, in sort
    return rsorter(data)
  File "/tmpfs/src/github/symbiflow-prjxray-presubmit-db-artix7/utils/xjson.py", line 82, in rsorter
    nitems.append((key(k), k, rsorter(v)))
  File "/tmpfs/src/github/symbiflow-prjxray-presubmit-db-artix7/utils/xjson.py", line 108, in rsorter
    nlist.sort(key=lambda n: n[0])
TypeError: unorderable types: tuple() < NoneType()
```

mithro:
Any idea what the input to `xjson.py` was there?

litghost:
Doesn't matter?

litghost:
I'll also point out that sorting ordered constructions (tuples and list) can break the data?

mithro:
In Python 3, you can't sort something which has a None object in it...

```
In [1]: ['a', None, '1'].sort()
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-1-46c34c36c41c> in <module>()
----> 1 ['a', None, '1'].sort()

TypeError: '<' not supported between instances of 'NoneType' and 'str'
```

litghost:
> In Python 3, you can't sort something which has a None object in it...
> 
> ```
> In [1]: ['a', None, '1'].sort()
> ---------------------------------------------------------------------------
> TypeError                                 Traceback (most recent call last)
> <ipython-input-1-46c34c36c41c> in <module>()
> ----> 1 ['a', None, '1'].sort()
> 
> TypeError: '<' not supported between instances of 'NoneType' and 'str'
> ```

[...]

PR #1243, opened 2020-02-15
Explicitly set which lists in JSON files should be sorted

mithro:
This is an alternative to https://github.com/SymbiFlow/prjxray/pull/1237

The primary difference is that it keeps all the sorting behaviour in the `sort_db.py` script.

The `sort` function now takes two functions which check if a list should be sorted or not. Any list which isn't **explicitly** marked as should be sorted or should be ordered will raise an error (which is why one function is not just the inverse of the other). This means the "hack" around two length lists and `tileconn.json` can be removed.

Fixes https://github.com/SymbiFlow/prjxray/issues/1236.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,5 @@
 SHELL = bash
-ALL_EXCLUDE = third_party .git env build docs/env
+ALL_EXCLUDE = third_party .git env build docs/venv docs/env
 
 # Check if root
 ifeq ($(shell id -u),0)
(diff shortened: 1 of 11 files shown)


litghost:
It makes very little sense to maintain a global registery of exceptions to a dangerous transformation.  Only a white list or local changes make sense from a maintainability standpoint.  The benefit gains from stable diffs is not large enough to break outputs at a great distance.  For example, if a developer only runs their fuzzer, they won't see `sort_db.py` breaking their output until CI.

Only "safe" (e.g. reversable) transformations are acceptable in `sort_db.py`.

mithro:
Reopening for now to be able to look at the CI results.

PR #1245, opened 2020-02-18
Fix for the lost IOB bits

mkurc-ant:
Fixed the 030 fuzzer to automatically detect where the PUDC_B site is.


litghost:
Did this work locally?  I expect that it will.

mkurc-ant:
@litghost I've checked locally only for Zynq 7020 and it did solve the missing bits. I left checks for others to the CI.

PR #1247, opened 2020-02-18
Couple of small cleanup changes

mithro:
These changes should change should fix all the warnings and make the current `make test-py` output from;
```
=================== 29 passed, 4 warnings in 1.65s ===================
```
to
```
=================== 29 passed, 0 warnings in 1.65s ===================
```

### DIFF
--- a/utils/environment.python.sh
+++ b/utils/environment.python.sh
@@ -1,0 +1,6 @@
+# FIXME: fasm should be installed into the running Python environment.
+export PYTHONPATH="${XRAY_DIR}:${XRAY_DIR}/third_party/fasm:$PYTHONPATH"
+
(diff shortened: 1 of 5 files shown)


utils/environment.python.sh
```diff
@@ -0,0 +1,6 @@
+# FIXME: fasm should be installed into the running Python environment.
+export PYTHONPATH="${XRAY_DIR}:${XRAY_DIR}/third_party/fasm:$PYTHONPATH"
```
litghost: This shouldn't be needed per https://github.com/SymbiFlow/prjxray/pull/1247/files#diff-b4ef698db8ca845e5845c4618278f29aR1

PR #1257, opened 2020-02-26
Fix warnings in documentation

rw1nkler:
This commit resolves all warnings in the project documentation.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -35,6 +35,8 @@ from markdown_code_symlinks import LinkParser, MarkdownSymlinksDomain
 # Add any Sphinx extension module names here, as strings. They can be
 # extensions coming with Sphinx (named 'sphinx.ext.*') or your custom
 # ones.
+
+# yapf: disable
 extensions = [
     'sphinx.ext.autodoc',
     'sphinx.ext.autosummary',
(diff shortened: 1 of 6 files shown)


docs/db_dev_process/fuzzers/index.rst
```diff
@@ -57,7 +57,6 @@ Programmable Interconnect Points (PIPs)
    :maxdepth: 1
```
mithro: Why are these others being removed?

docs/conf.py
```diff
@@ -100,7 +100,10 @@
 # List of patterns, relative to source directory, that match files and
 # directories to ignore when looking for source files.
 # This patterns also effect to html_static_path and html_extra_path
-exclude_patterns = ['_build', 'Thumbs.db', '.DS_Store']
+exclude_patterns = [
+    '_build', 'architecture/copying.md', 'db_dev_process/minitests/index/**',
```
mithro: One per line would be better.

docs/db_dev_process/fuzzers/index.rst
rw1nkler: Those fuzzers do not have `README.md` file.

docs/conf.py
rw1nkler: It was like that, but the format auto-check on Travis complained about the formatting.
I pushed again the previous version.

rw1nkler:
I restored the previous version obtained using `make format` command. This ensures that the formatting is valid in CI.

One per line formating is not allowed by CI, @mithro:
format1
format2

tmichalak:
@mithro Do we merge this one now and then open an issue to fix the formatter?

mithro:
You should be able to add markers to the file to tell yapf to ignore this structure.

rw1nkler:
@mithro, I've added the yapf markers

rw1nkler:
Could we merge this? @mithro

PR #1268, opened 2020-03-09
Convert CLB/CLB_INT tilegrid fuzzer to workaround prohibited locations.

litghost:


litghost:
@mithro FYI, this fixes the failure discovered in https://github.com/SymbiFlow/symbiflow-arch-defs/pull/1360

The prohibited sites are something we will likely want to not use in any P&R flow, but it is good to have them documented.

litghost:
> how is this related to SymbiFlow/symbiflow-arch-defs#1360?

Vpr wanted to route through an int tile in excluded range, and fasm2bits errored because we were missing the base address for the tile.

PR #1270, opened 2020-03-13
Add prohibited site list to tilegrid.

litghost:
Prohibited sites are sites that Vivado will not place at.


acomodi:
LGTM, @mithro @tmichalak, I would proceed with the merge as this is needed for https://github.com/SymbiFlow/symbiflow-arch-defs/pull/1360.
The db diff is huge, but it regards the addition of the prohibited keys in the various tilegrids as I understand from the code.

Issue #1264, opened 2020-03-06
htmlgen - KeyError: 'XRAY_ROI'


mithro:
```
source $(dirname ${BASH_SOURCE[0]})/../utils/environment.sh
--------------------------------------------
Loading tilegrid.
Writing ./html/artix7/index.html
Traceback (most recent call last):
  File "/tmp/tmp.0LNdLtQCmx/prjxray/htmlgen/htmlgen.py", line 1017, in <module>
    main()
  File "/tmp/tmp.0LNdLtQCmx/prjxray/htmlgen/htmlgen.py", line 1012, in main
    allow_fake=args.allow_fake,
  File "/tmp/tmp.0LNdLtQCmx/prjxray/htmlgen/htmlgen.py", line 980, in run
    mk_tilegrid_page(dbstate, output, grid)
  File "/tmp/tmp.0LNdLtQCmx/prjxray/htmlgen/htmlgen.py", line 387, in mk_tilegrid_page
    get_setting("XRAY_PART"), get_setting("XRAY_ROI"),
  File "/tmp/tmp.0LNdLtQCmx/prjxray/htmlgen/htmlgen.py", line 28, in <lambda>
    return lambda name: settings[name]
KeyError: 'XRAY_ROI'
The command "./.travis/generate-html.sh" exited with 1.
```

mithro:
https://travis-ci.com/SymbiFlow/prjxray-db/builds/152196782?utm_medium=notification&utm_source=email

mithro:
This is still happening on head - https://travis-ci.com/github/SymbiFlow/prjxray-db/builds/157300050?utm_medium=notification&utm_source=email
```
source $(dirname ${BASH_SOURCE[0]})/../utils/environment.sh
--------------------------------------------
Loading tilegrid.
Writing ./html/artix7/index.html
Traceback (most recent call last):
  File "/tmp/tmp.nNvTwaA29R/prjxray/htmlgen/htmlgen.py", line 1017, in <module>
    main()
  File "/tmp/tmp.nNvTwaA29R/prjxray/htmlgen/htmlgen.py", line 1012, in main
    allow_fake=args.allow_fake,
  File "/tmp/tmp.nNvTwaA29R/prjxray/htmlgen/htmlgen.py", line 980, in run
    mk_tilegrid_page(dbstate, output, grid)
[...]

PR #1287, opened 2020-04-01
htmlgen: use XRAY_ROI_TILEGRID instead of XRAY_ROI

acomodi:

This PR fixes https://github.com/SymbiFlow/prjxray/issues/1264.

It changes to the usage of XRAY_ROI_TILEGRID instead of XRAY_ROI. This is more accurate as it corresponds to the whole device (or at least the one described by tilegrid.json). 

In addition XRAY_ROI_TILEGRID is mandatory to have in the settings files


mithro:
LGTM.

mithro:
@acomodi Can you look at getting htmlgen run as part of the CI for this repository?

acomodi:
> @acomodi Can you look at getting htmlgen run as part of the CI for this repository?

Done, I have added it in Travis CI, as well as a script to run the htmlgen on all the setting files

acomodi:
Hmm actually this won't work in Travis, unless I add also the download of the db in the script, to get the tilegrid info. Probably needs to be added in Kokoro?

mithro:
@acomodi Kokoro or Travis is fine. Testing with `make download-db` could also be fine?

acomodi:
@mithro Sure, I'll go with the make download-db solution then

mithro:
Maybe as a separate travis build stage or something?

acomodi:
So, re-thinking at this I would say that, in case there is a PR that adds an additional part, with a new tilegrid, the database would be incomplete, therefore the test would result in a failure. 
I believe we should add this test as a Kokoro stage at the end of the build

mithro:
@acomodi Whatever you think is best.

acomodi:
@mithro All the CI build went green. I see that Kintex and Tests CI are marked as still running, but they actually went green (unsure why this happens).

mithro:
Okay merged it.

PR #1324, opened 2020-05-07
SSTL15 features

mkurc-ant:
This PR adds SSTL15 features to fuzzer 030. No new bits were found for that IO standard. Some features changed names though.

Apart from that, the IOSTANDARD minitest has been updated to also check SSTL15 and set INTERNAL_VREF wherever appropriate. The latter change eliminated missing bits that the test was reporting previously.

### DIFF
--- a/fuzzers/030-iob/generate.py
+++ b/fuzzers/030-iob/generate.py
@@ -26,7 +26,7 @@ def drives_for_iostandard(iostandard):
         drives = [4, 8, 12, 16, 24]
     elif iostandard == 'LVCMOS12':
         drives = [4, 8, 12]
-    elif iostandard == 'SSTL135':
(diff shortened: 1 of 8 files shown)


mkurc-ant:
I've noticed that now there is a feature `LVCMOS15_SSTL15.DRIVE.I16_I_FIXED`. Since SSTL15 seems to use the 16mA drive only would it make sense to remove the `I_FIXED` part to have `LVCMOS15_SSTL15.DRIVE.I16` ?

litghost:
> Now I see that you don't actually specify DRIVE for SSTL15 so maybe the `I_FIXED` should stay?

Yes, because SSTL15 only supports one drive settings, hence the "fixed" bit.

Issue #1285, opened 2020-03-30
Missing INIT string data in FASM for selected BRAM sites


nelsobe:
When generating a large memory in Verilog, in some cases, parts of the BRAM INIT information is missing when the .bit file is converted to FASM.

Here is an example (memory was initialized to all 1's in Verilog, BRAM INIT string should be 256 1's but some get left off):

BRAM_R_X37Y35.RAMB18_Y0.INIT_3F[127:0] = 128'b11111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111

There are 3 other such INIT strings in this design out of many that have this issue.  This is for the xc7a50 part.

The common thread is the location of the particular BRAM cells -
X37Y35 and X37Y40 are the only locations identified thus far where the information drops occur.  Have run 100's of other test cases which map BRAMs to other locations without problems (cannot say have tried every other single site, however).

Additional info:
- This memory uses 32 RAMB36's.  Only 2 of the 32 have this problem.
- Always is at BRAM_R_X37Y35 and BRAM_R_X47Y40. Other locations in that column are just fine.

To re-create:
1. Generate a bitstream for the attached .sv code for the Artix-7 50T part.

2. Run $XRAY_BIT2FASM on the resulting .bit file.

Here is the test code:

[128kb8.zip](https://github.com/SymbiFlow/prjxray/files/4405050/128kb8.zip)

litghost:
If you run $XRAY_BIT2FASM --verbose, and grep the output for "unknown", do you find anything?

nelsobe:
Yes, lots of such messages.  I just dumped everything it output to a file, it is attached.   
[128kb8.txt](https://github.com/SymbiFlow/prjxray/files/4405115/128kb8.txt)

nelsobe:
This issue (#1285) suggests that there are some holes in the prjxray bitstream db as it relates to BRAMs.  Interestingly, it only occurs for _some_ BRAMs in one column of the device - the bitstream is correct for others.  

Would like to track down and propose a fix for this, and have some new students coming on-line soon to help.  A bit of direction from the experts out there could be helpful to jump start this.

mithro:
@litghost, @nelsobe could use some advice on the best way to look into this further.

litghost:
@nelsobe Have you confirmed that the `INIT_xx` does from the FASM do **not** match the `INIT_xx` parameter from the relevant RAMBx cells in Vivado?

[...]

PR #1335, opened 2020-05-19
bitread: Don't overwrite last frame

tmichalak:
This PR targets the following issue https://github.com/SymbiFlow/prjxray/issues/1285
There seemed to be an empty frame at the end of the configuration data which was used to overwrite the contents of the last configuration frame address of which happened to belong to the BRAM_R tile.
The fix makes sure the last frame does not get overwritten.
Remaining work or this PR is adding a test.


litghost:
So I don't really understand what the original problem was.  Can you make a better description for the issue and the solution?  Maybe some diagrams?

tmichalak:
@litghost I added more details to the PR description.

litghost:
> Bug
>
> Due to the fact that the there is no write to the CMD register the zero frame is used to overwrite the content of the last configuration frame.

I think I know what is going on, and it has to do with how bitread models the FAR address register.  I believe you analysis is not correct.  If my analysis is correct, then the solution present in this PR is not correct.  I'll take a look at the FAR address modelling code, and come back with a comment in a bit.

litghost:
@tmichalak:

I believe the bug is actually located here: https://github.com/SymbiFlow/prjxray/blob/master/lib/include/prjxray/xilinx/configuration.h#L253

This switch statement likely needs to advance "frame_address_register" by 1 frame when the CRC command is recieved.  I'll annotate your example:

```
 [Write Type=1 Address= 1 Length=         1 Reg="Frame Address"]
Data in hex:
       c00000  << BOTTOM Row= 0 Column= 0 Minor= 0 Type=Block RAM
```

FAR = 0xc00000 because it was set via the "Frame Address" write

```
[Write Type=1 Address= 2 Length=       101 Reg="Frame Data Input"]
Data in hex:
       0        0        0        0 
       (...) << configuration words for frame c00000
       0 
```

FAR = 0xc00001, because "Frame Data Input" write advanced it.

```
[Write Type=1 Address= 1 Length=         1 Reg="Frame Address"]
Data in hex:
       c00000 << BOTTOM Row= 0 Column= 0 Minor= 0 Type=Block RAM
```

FAR = 0xc00000 because it was set via the "Frame Address" write.

```
[Write Type=1 Address= 0 Length=         1 Reg="CRC"]
Data in hex:
1e640f57 
```

FAR = 0xc00001 because it was set via the "Frame Address" write.  Not that next write is a "Frame Data Input", because CRC has already advanced the FAR to the correct location!

```
[Write Type=1 Address= 2 Length=       101 Reg="Frame Data Input"]
Data in hex:
       0        0        0        0 
       (...) << configuration words for frame c00001
       0 
[Write Type=1 Address= 1 Length=         1 Reg="Frame Address"]
Data in hex:
       c00001 
[Write Type=1 Address= 0 Length=         1 Reg="CRC"]
Data in hex:
12d781cb 

(packets for frames between c00002 and 0xc0017e)

[Write Type=1 Address= 2 Length=       101 Reg="Frame Data Input"]
Data in hex:
       0        0        0        0 
    8000        0        0        0  
   (...) << configuration words for frame c0017f
       0 
[Write Type=1 Address= 1 Length=         1 Reg="Frame Address"]
Data in hex:
  c0017f 
[Write Type=1 Address= 0 Length=         1 Reg="CRC"]
Data in hex:
f8a17b70 
```

At this point the FAR = 0xc00180 because the CRC command advanced it 1 frame!  So this zero frame is writing into 0xc00180, **not** overwriting 0xc0017f!  This is where I believe the bug occurred.

```
[Write Type=1 Address= 2 Length=       101 Reg="Frame Data Input"]
Data in hex:c0017f
       0        0        0        0 
      (...) << zero frame
       0 
[Write Type=1 Address= 1 Length=         1 Reg="Frame Address"]
Data in hex:
  c0017f 
[Write Type=1 Address= 0 Length=         1 Reg="CRC"]
Data in hex:
22075150 
```

tmichalak:
@litghost 
First of all this example is only for the last row, but such scenario applies to all other rows as well. I guess a broader picture is needed for this. So for the entire bitstream the critical frames look as follows:
```
WCFG Command

FAR 0
FDRI writes for FAR 0
FAR 0
CRC for FAR 0

FDRI writes for 1
FAR 1
CRC for FAR 1

FDRI writes n-1
FAR n-1
CRC for n-1

FDRI write n
FAR n
CRC for n

FDRI write zero frame
FAR n
CRC for zero frame and FAR n

WCFG command

FAR 40000 (0 in next row)
FDRI writes for FAR 40000
FAR 40000
CRC for FAR 40000

FDRI writes for 40001
FAR 40001
CRC for FAR 40001

FDRI writes 40002
....
```
> FAR = 0xc00001 because it was set via the "Frame Address" write. Not that next write is a "Frame Data Input", because CRC has already advanced the FAR to the correct location!

This is true, but only for the first frame in a configuration row and this is how it is implemented. For the rest it is advanced by the FDRI write. Why? Because of how the configuration data is layed out in the bitstream (presented in the snippet above) and the fact that Type 2 writes don't have explicit FAR writes in between frames (it is auto-incremented).

> This switch statement likely needs to advance "frame_address_register" by 1 frame when the CRC command is recieved.

Won't work for Type 2 writes where there is CRC only at the end of the packet.

> At this point the FAR = 0xc00180 because the CRC command advanced it 1 frame! So this zero frame is writing into 0xc00180, not overwriting 0xc0017f! This is where I believe the bug occurred.

Incorrect. First of all there is no 0xc00180 frame in the FPGA, the FAR counter in this particular case advances to 0 since it reaches the end of the available address range. Secondly look at the broader snippet and all above points.

This being said I stand by my point. Maybe I should be more precise in the bug root cause which should say: 
Due to the fact that the there is no **WCFG** write to the CMD register the zero frame is used to overwrite the content of the last configuration frame.

This entire discussion proves that there should be an extension to the bitstream format chapter of the documentation where all of the above should go into.

litghost:
> https://github.com/SymbiFlow/prjxray/pull/1335#issuecomment-634476283

Your more detailed example doesn't change my point.  The FAR is clearly advanced when the CRC command is issued, and if the bitread modeled that, no special logic would be required.

litghost:
Type 2 packets remember the previous register target (e.g. FDRI vs CRC vs CMD), but the FAR register behavior is likely independent of the Type 1/2 behavior.

tmichalak:
Analyzed the case with bits2rbt and bitread (with aux support required for bits2rbt) ([PR](https://github.com/SymbiFlow/prjxray/pull/1503))
To compare the original RBF (vivado generated) with the reconstructed (bits2rbt generated).

Without the fix, ie where the `bits` files are equal the RBF test fails: 
```
573766c573766
< 00000000000000001000000000000000
---
> 00000000000000000000000000000000
573866c573866
< 11111000101000010111101101110000
---
> 00100010000001110101000101010000
bits2rbt_bram_xc7 FAIL
```
clearly there is one missing bit (`bit_00c0017f_004_15`) which causes the CRC value to be invalid

With the fix the bit `bit_00c0017f_004_15` is in bits, but the RBT comparison fails on different CRCs:
```
573812c573812
< 00000000000000000000000000000000
---
> 00000000000000000001001110101111
573866c573866
< 11111000101000010111101101110000
---
> 01000011101111001000011010011110
573871a573872
> 00000000000000001000000000000000
573917,573918c573918
< 00000000000000000000000000000000
< 00000000000000000000000000000000
---
> 00000000000000000001001110101111
573972c573972
< 00100010000001110101000101010000
---
> 01000011101111001000011010011110
```

PR #1338, opened 2020-05-22
Added support for xc7z020clg400-1 and added an openocd config file for pynq-z1

andrewb1999:
Added generation of the xc7z020clg400-1, which is required for using the pynq-z1 board.  Also, created a working openocd configuration for the pynq-z1.


mithro:
@andrewb1999 - Could you also add this to our `openocd` conda packages?
 - https://github.com/litex-hub/litex-conda-prog/tree/master/prog/openocd
 - https://github.com/SymbiFlow/conda-packages/tree/master/openocd

Would be good to try and get it upstream into openocd -- http://openocd.org/doc/doxygen/html/patchguide.html

Issue #1246, opened 2020-02-18
Zynq7 build unstable


acomodi:
There are some instabilities regarding the tilegrid fuzzer that should be solved.

For reference, here is an error of a particular run: https://github.com/SymbiFlow/prjxray/pull/1227#issuecomment-587601068

acomodi:
There have been periodic instabilities on the zynq7 build. The ones that did show up for now are the following:

Bit collisions:
- https://github.com/SymbiFlow/prjxray/pull/1253#issuecomment-590810303

Fuzzer failures:
- https://github.com/SymbiFlow/prjxray/pull/1227#issuecomment-587601068

The latter one is most probably already solved by increasing the number of specimens.

mkurc-ant:
Again, instability: https://github.com/SymbiFlow/prjxray/pull/1307

tcal-x:
Also, [#1313](https://github.com/SymbiFlow/prjxray/pull/1313).

mithro:
```
Checking zynq7 database
============================
Checking individual .db...
segbits_*.db: 33 okay
mask_*.db: 39 okay
Checking aggregate dir...
ERROR: 1 collisions
  0042229a_044_27: had INT_R_X69Y22.INT_R.IMUX25.LOGIC_OUTS18, got CLBLM_R_X69Y22.CLBLM_R.SLICEL_X1.BLUT.INIT[16]
Traceback (most recent call last):
  File "utils/checkdb.py", line 190, in <module>
    main()
  File "utils/checkdb.py", line 186, in main
    run(args.db_root, args.part, verbose=args.verbose)
  File "utils/checkdb.py", line 172, in run
    check_tile_overlap(db, verbose=verbose)
  File "utils/checkdb.py", line 148, in check_tile_overlap
    raise ValueError("%s collisions" % len(collisions))
ValueError: 1 collisions
Makefile:172: recipe for target 'db-check-zynq7' failed
make: *** [db-check-zynq7] Error 1
```

PR #1360, opened 2020-06-11
Conditionally use MAP_POPULATE to support non-Linux OSes

HackerFoo:


litghost:
I almost wonder if we need MAP_POPULATE at all?

Should we use CMake's "try_compile" for this?

HackerFoo:
It's not clear what will happen with `MAP_PRIVATE | MAP_POPULATE` before 2.6.23. It could be a runtime error, or maybe it will just be ignored.

This change seems like the safest option, and also documents where MAP_POPULATE will apply. Otherwise I'd use:

```c
#ifndef MAP_POPULATE
#define MAP_POPULATE 0
#endif
```

litghost:
> It's not clear what will happen with `MAP_PRIVATE | MAP_POPULATE` before 2.6.23

Fair.  I'm wondering if we should just drop `MAP_POPULATE`, if the performance improvement is small.

Maybe just drop it and see what happens to the CI build time.

litghost:
> Looks like there was a tool failure in kokoro, re-running CI now.

Master is green again, re-running CI.

HackerFoo:
I don't understand how this PR could change anything, since it shouldn't affect recent Linux systems at all, but I got this error for the Zynq database:

```
ERROR: 1 collisions
  0040209a_006_27: had CLBLM_R_X65Y53.CLBLM_R.SLICEL_X1.BLUT.INIT[16], got INT_R_X65Y53.INT_R.IMUX25.LOGIC_OUTS18
```

litghost:
> I don't understand how this PR could change anything, since it shouldn't affect recent Linux systems at all, but I got this error for the Zynq database:
> 
> ```
> ERROR: 1 collisions
>   0040209a_006_27: had CLBLM_R_X65Y53.CLBLM_R.SLICEL_X1.BLUT.INIT[16], got INT_R_X65Y53.INT_R.IMUX25.LOGIC_OUTS18
> ```

That Zynq failure has been showing up intermittently: https://github.com/SymbiFlow/prjxray/issues/1246 , definitely not your issues.

litghost:
This PR has been superceeded by #1361

PR #1368, opened 2020-06-16
Revert "Move fasm2frames to prjxray"

litghost:
Reverts SymbiFlow/prjxray#1347

Caused:
```
+ python3 /tmpfs/src/github/symbiflow-prjxray-presubmit-db-artix7/prjxray/fasm2frames.py design.fasm design.frm
  File "/tmpfs/src/github/symbiflow-prjxray-presubmit-db-artix7/prjxray/fasm2frames.py", line 23, in <module>
    from . import fasm_assembler, util
SystemError: Parent module '' not loaded, cannot perform relative import
```


setup.py
```diff
@@ -30,5 +30,5 @@
         "Operating System :: OS Independent",
     ],
     entry_points={
-        'console_scripts': ['fasm2frames=prjxray.fasm2frames:main'],
+        'console_scripts': ['fasm2frames=utils.fasm2frames:main'],
```
acomodi, line 33: Given that this now has `utils` instead of `prjxray`, utils needs to be added to the packages parameter as well, as fasm2frames cannot be located otherwise when installing prjxray.

litghost:
PR #1347 broke the build, so this revert undoes the merge.  @acomodi you are free to revert the revert and get the CI green.

PR #1369, opened 2020-06-19
setup: temporarily add `utils` to the python pkg to enable fasm2frames

acomodi:


litghost:
I assume this can be closed, as we don't need it anymore?

acomodi:
> I assume this can be closed, as we don't need it anymore?

Indeed, closing.

PR #1383, opened 2020-06-30
Add dedicated port type to harness design.json

andrewb1999:
Signed-off-by: Andrew Butt <butta@seas.upenn.edu>

Allows a convergence of prjxray roi harnesses and a general partition region in https://github.com/SymbiFlow/symbiflow-arch-defs/pull/1561

### DIFF
--- a/minitests/roi_harness/create_design_json.py
+++ b/minitests/roi_harness/create_design_json.py
@@ -46,6 +46,15 @@ def main():
     design_json['info'] = {}
     with open(args.design_txt) as f:
         for d in csv.DictReader(f, delimiter=' '):
+            if d['name'].startswith('dout['):
+                d['type'] = 'out'
+            elif d['name'].startswith('din['):
+                d['type'] = 'in'
+            elif d['name'].startswith('clk'):
+                d['type'] = 'clk'
+            else:
+                assert False, d
+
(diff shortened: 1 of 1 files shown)


andrewb1999:
@litghost Any idea why this change would cause the zynq 7 database build to fail?

minitests/roi_harness/create_design_json.py
```diff
@@ -46,6 +46,15 @@ def main():
     design_json['info'] = {}
     with open(args.design_txt) as f:
         for d in csv.DictReader(f, delimiter=' '):
+            if d['name'].startswith('dout['):
+                d['type'] = 'in'
```
litghost, line 50: Ins't this backwards for the file format?  dout[x] is an output, not an input right?

andrewb1999: Ah yes, you're right.  Fixing now.

andrewb1999:
CI looks good now.  Good to go?

litghost:
@andrewb1999 Once CI runs on the new master, a new prjxray-db can be pushed.  Remind @mithro next Monday or Tuesday and they should be able to update it.

mithro:
New db has been pushed.

PR #1389, opened 2020-07-08
Fix SDF generation

### LINKED ISSUE #1388: SDF generation produces lowercase CELLTYPE for routing_bel in slicel/slicem
https://github.com/SymbiFlow/prjxray-db/blob/master/artix7/timings/slicel.sdf#L807

Unsure why this is happening as the python seems to convert all CELLTYPEs to uppercase.  Causes symbiflow-arch-defs CI to fail with new prjxray-db: https://github.com/SymbiFlow/symbiflow-arch-defs/pull/1571

andrewb1999:
Fixes #1388


litghost:
This seems okay, I'd like @kgugala or @acomodi to LGTM, but if the CI is green I'll merge it so we can unblock you.

andrewb1999:
@litghost This made the changes I expected.  I think this should fix the symbiflow prjxray-db issues.

PR #1390, opened 2020-07-10
Add overlay.py to allow for overlay architecture generation

andrewb1999:
Signed-off-by: Andrew Butt <butta@seas.upenn.edu>

Essentially an inverse roi.  Can accommodate cutting out multiple roi like regions.  Testing with symbiflow-arch-defs overlay generation and working.

### DIFF
--- a/prjxray/overlay.py
+++ b/prjxray/overlay.py
@@ -1,3 +1,13 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2017-2020  The Project X-Ray Authors.
+#
+# Use of this source code is governed by a ISC-style
+# license that can be found in the LICENSE file or at
+# https://opensource.org/licenses/ISC
+#
+# SPDX-License-Identifier: ISC
 class Overlay(object):
     """ Object that represents an overlay.
 
@@ -5,8 +15,8 @@ class Overlay(object):
 
     """
 
-    def __init__(self, db, region_dict):
-        self.grid = db.grid()
+    def __init__(self, grid, region_dict):
+        self.grid = grid
         self.region_dict = region_dict
 
     def tile_in_roi(self, grid_loc):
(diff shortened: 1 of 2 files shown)


prjxray/overlay.py
```diff
@@ -0,0 +1,60 @@
+class Overlay(object):
+    """ Object that represents an overlay.
+
+    Can be used to iterate over tiles and sites not inside a partition region.
+
+    """
+
+    def __init__(self, db, region_dict):
```
litghost, line 8: Why not just take the grid?

prjxray/overlay.py
```diff
@@ -0,0 +1,60 @@
+class Overlay(object):
```
litghost, line 1: Tests?

prjxray/overlay.py
```diff
@@ -0,0 +1,60 @@
...
+                continue
+
+            gridinfo = self.grid.gridinfo_at_loc(loc)
+
+            if tile_types is not None and gridinfo.tile_type not in tile_types:
+                continue
+
+            yield tile_name
+
+    def gen_sites(self, site_types=None):
```
litghost, line 41: This seems like a less useful method, as it is just an extension of `gen_tiles`.  Why have both?

prjxray/overlay.py
andrewb1999: Which grid are you talking about?  I'm trying to match the original roi input schema which was x_max, x_min, y_min, y_max.  Just abstracting that to take multiple regions.

prjxray/overlay.py
andrewb1999: Not sure what a proper test for this in prjxray would be.  Any suggestions?

prjxray/overlay.py
andrewb1999: Actually, I was just leaving in both the gen_tiles and gen_sites methods to match the roi interface.  I do not believe they are used in architecture generation.  Maybe I should just get rid of both methods?

litghost: Drop both

prjxray/overlay.py
litghost: Provide a grid and a region dict and ensure that things work as expected?

prjxray/overlay.py
litghost: You take the `db` argument, but immediately grab the grid, and discard the db.  Just take the grid argument.

prjxray/overlay.py
andrewb1999: Ok makes sense.  Should that be a new minitests subdirectory?

litghost: We have https://github.com/SymbiFlow/prjxray/tree/master/tests and I believe we are running them via pytest.

litghost: https://github.com/SymbiFlow/prjxray/blob/master/Makefile#L59-L60

andrewb1999:
@litghost This test case works on my local machine, but I guess it doesn't work here because the database isn't built.  What would you suggest doing?

prjxray/overlay.py
```diff
@@ -0,0 +1,30 @@
...
+# SPDX-License-Identifier: ISC
+class Overlay(object):
+    """ Object that represents an overlay.
+
+    Can be used to iterate over tiles and sites not inside a partition region.
+
+    """
+
+    def __init__(self, grid, region_dict):
+        self.grid = grid
```
litghost, line 19: Is `self.grid` even used anymore?

andrewb1999: Oh you're right, it's not.  That fixes the testing issue as well.

andrewb1999:
@litghost can this be merged?

PR #1429, opened 2020-09-01
Fix python on kokoro and break circular import through tile_segbits.

litghost:
@mkurc-ant This probably needs to be fixed in prjuray-tools too.  Another reason to converge the repos when we can.

### DIFF
--- a/.github/kokoro/steps/hostsetup.sh
+++ b/.github/kokoro/steps/hostsetup.sh
@@ -22,6 +22,7 @@ echo "Host adding PPAs"
 echo "----------------------------------------"
 wget -O - https://apt.kitware.com/keys/kitware-archive-latest.asc 2>/dev/null | sudo apt-key add -
 sudo apt-add-repository 'deb https://apt.kitware.com/ubuntu/ xenial main'
+sudo add-apt-repository ppa:deadsnakes/ppa
 echo "----------------------------------------"
 
 echo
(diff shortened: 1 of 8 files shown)


mkurc-ant:
@litghost Good catch. I've issued a similar PR to prjuray-tools.

mithro (approved):
Generally LGTM.

BTW Are we requiring Python 3.8 now? If so that should be in the README right?

litghost:
> Generally LGTM.
> 
> BTW Are we requiring Python 3.8 now? If so that should be in the README right?

We only really need Python 3.6, and the Xeniel default is Python 3.5.  We don't specifically require Python 3.5 really, but `python3.5 -mpip install numpy` now generates the error:

```
Collecting numpy (from -r requirements.txt (line 6))
  Downloading https://files.pythonhosted.org/packages/2c/2f/7b4d0b639a42636362827e611cfeba67975ec875ae036dd846d459d52652/numpy-1.19.1.zip (7.3MB)
    Complete output from command python setup.py egg_info:
      File "<string>", line 1, in <module>
      File "/tmpfs/tmp/pip-build-4efh__zv/numpy/setup.py", line 30, in <module>
        raise RuntimeError("Python version >= 3.6 required.")
    RuntimeError: Python version >= 3.6 required.
```

mithro:
We could potentially just move to bionic?

litghost:
Looks like the newer python consistently results in a DB conflict, investigating.

litghost:
> Looks like the newer python consistently results in a DB conflict, investigating.

I've replicated the issue locally, and have a proposed fix.  We'll see if it works on the CI.

PR #1446, opened 2020-10-01
Add missing FREQ_BB active feature

litghost:
This helps with https://github.com/SymbiFlow/symbiflow-arch-defs/issues/1536

This adds `CMT_TOP_R_UPPER_T.PLL_CLK_FREQ_BB[0-3]_NS_ACTIVE`


mkurc-ant (approved):
LGTM. So it seems that we had incorrect base address for PLL tile.

litghost:
> LGTM. So it seems that we had incorrect base address for PLL tile.

That too

PR #1454, opened 2020-10-06
Add missing file segbits to 000-init-db.

litghost:


mithro:
LGTM! Merge if CI says it's green.

PR #1468, opened 2020-10-23
Fixup doc requirements to work around https://github.com/pypa/pip/issues/9031

litghost:
See https://github.com/pypa/pip/issues/9031 for way too many details.


mithro (approved):
LGTM

PR #1474, opened 2020-10-28
Fix new BRAM36 features.

litghost:
This fixes the BRAM features for https://github.com/SymbiFlow/symbiflow-arch-defs/pull/1720


mithro (approved):
LGTM.

Issue #1476, opened 2020-10-29
Add part data for all parts


litghost:
Once #1475 is complete, populate part data for all parts that fabric documentation exists for.  #1475 should be done first to avoid duplicating all of the fabric data repeatedly.

dnltz:
@litghost - Before I will start I have some questions:
1) Where do I know the fabric documentation exists? Because the script in settings/ exists?
2) Where do I know devices share the same fabric?
3) Any restrictions on the parts or do you want all available packages, devices and speed grades?

litghost:
I would start with a family, e.g. "artix7".  Example:
```
get_parts -filter "FAMILY == artix7"
```

litghost:
> * Where do I know devices share the same fabric?

I think the default is each device is a different fabric, and add a file that specifies which devices are the same fabric.  Does that make sense?

dnltz:
Ah, that's a brilliant way to fetch all available parts, thanks!

Okay. I understood different devices share the same fabric as well. Makes absolutely sense to me now. That's also the reason for: https://github.com/SymbiFlow/prjxray/pull/1549

dnltz:
hey @litghost, I wrote a script to populate all supported parts. Should I also extend the Makefile to roi_only all parts during the CI runs?

litghost:
> hey @litghost, I wrote a script to populate all supported parts. Should I also extend the Makefile to roi_only all parts during the CI runs?

Yep, exactly!

Issue #1475, opened 2020-10-29
Refactor how fabric data is shared between parts


litghost:
Before adding data for all parts (e.g. pinouts), a refactoring of how fabric data is shared is required.   For terminology:

- **Fabric**
  - The interior tile and routing layout within a FPGA silicon
- **Device**
  - The name of the **fabric** used by Vivado
- **Package**
  - A standardized pin layout that can be used with one or more **devices**.
- **Part**
   - A **device** wire bonded to a **package**
- **Family**
  - A collection of **devices** and **parts** that share some common traits.

Currently prjxray-db only has **families** (e.g. artix7, kintex7, zynq7) and **parts** (e.g. xc7a35tcpg236-1).  Because prjxray-db is missing the explicit idea of a **device** and **fabric**, **fabric** data is currently duplicated between **parts** that have the same **device**.  To be very specific:

- The **fabric** for a **part** is defined in `prjxray-db/<family>/<part>` with the following files:
  - `node_wires.json`
  - `tilegrid.json`
  - `tileconn.json`

Because some **parts** share **fabrics** with multiple parts, these files are copied between the parts.  This is wasteful.  The solution is to:

- Add a small file describing:
  - The list of **parts** in each **family**
  - The **device** for each **part**
  - The speed grades for each **part**
  - The **fabric** used in each **device**

Then the **fabric** data (e.g. `tilegrid.json`, etc) should be moved to **fabric** folder rather than a **part** folder.  The `prjxray` library should be updated to transparently handle the **fabric**/**device** split from the **part**.

This change is required before adding **part** data for all parts.

hansfbaier:
In order to help with this, where are the data sources they are derived from?
datasheets, scripts, parts of vivado?

litghost:
When Vivado opens a design, it has both the `part`, the `device` and `package`.  Here is a small TCL example:
```
link_design -part xc7a35tcpg236-1
set design [get_design]
set part [get_parts [get_property PART $design]]
set package [get_property PACKAGE $part]
set device [get_property DEVICE $part]
```

From the device you can find all of the packages for that device:
```
get_parts -filter "DEVICE == xc7a35t"
```

You can also just get all parts that Vivado supports, but you'd want to filter by family:
```
get_parts -filter "FAMILY == artix7"
```

hansfbaier:
@litghost Where do we get the fabric from?

dnltz:
@hansfbaier - Did you start working on this issue?

[...]

PR #1486, opened 2020-11-06
Doc for adding a new part to PrjXray.

tcal-x:
Signed-off-by: Tim Callahan <tcal@google.com>


NEWPART.md
```diff
@@ -0,0 +1,203 @@
+
```
litghost, line 1: Should be in `docs/`?

NEWPART.md
```diff
@@ -0,0 +1,203 @@
...
+```
+make -j32 MAX_VIVADO_PROCESS=32 db-part-only-artix7_100t
+```
+
+* It should run fuzzers 000, 001, 005, 072, 073, 074, and 075.
+                
+* 005 will take a long time unless you throw a lot of cores at it.  32-64 cores suggested.
+
+* You _may_ get a spurious failure, since the parallel runs may violate ordering dependencies among the fuzzers.  If this happens, clean the fuzzer that failed (e.g.  `make -C fuzzers/005-tilegrid clean`), and then rerun the top-level make.
```
litghost, line 74: There shouldn't be ordering failing in 005?

NEWPART.md
```diff
@@ -0,0 +1,203 @@
...
+
+
+
+### Step 2 ###
+
+Add a new settings file.  Usually you will start with an existing settings file and modify it.
+Assuming you're in prjxray/,
+
+```
+cp settings/artix7_200t.sh settings/artix7_100t.sh
```
litghost, line 34: Please change to two instances of what needs to be changed.  One that makes clear the old and new, and an example.  So for example this block would be:

```
cp settings/<baseline settings>.sh settings/<new settings>.sh
git add settings/<new settings>.sh
```

Example:

```
cp settings/artix7_200t.sh settings/artix7_100t.sh
git add settings/artix7_100t.sh
```

NEWPART.md
```diff
@@ -0,0 +1,203 @@
...
+```
+
+### Step 3 ###
+
+Edit the top Makefile
+
+* Add the new part to the correct list.   This is the basename of the new settings file that you just created. In this case, we added `artix7_100t`:
+
+```
+ARTIX_PARTS=artix7_200t artix7_100t
```
litghost, line 58: Add a github permalink to the lines you are talking about here to the doc.  This step is a little hard to follow unless you first read the Makefile.  Maybe setup what this line is doing, and why you need to change it?

NEWPART.md
```diff
@@ -0,0 +1,203 @@
...
+```
+
+
+### Step 4 ###
+
+Make the following target, adapted for your new part, and setting the parallelism factor appropriate for your machine:
```
litghost, line 64: Explain what this is going to do.  The is also a good example of where the template comment from above is important.  In this case, the template version is:

```
make -j32 MAX_VIVADO_PROCESS=32 db-part-only-<new settings>
```

And make clear that, the name of the settings sh files and the change to the Makefile and the name here all need to be the same.

NEWPART.md
```diff
@@ -0,0 +1,203 @@
...
+    * Run `python3 ../analyze_errors.py --output_ignore_list  > new-ignored`
+    * Inspect and compare against existing ignored wire files
+    * IF IT LOOKS GOOD, copy it to an appropriately-named file: `cp new-ignored ../ignored_wires/artix7/xc7a100tfgg676-1_ignored_wires.txt`
+    * Add it: `git add ../ignored_wires/artix7/xc7a100tfgg676-1_ignored_wires.txt`
+
+* Return to prjxray/ directory, and clean up 074 to prepare for the rerun: `make -C fuzzers/074-dump-all clean`
+
+* Rerun the top make command, `make -j32 MAX_VIVADO_PROCESS=32 db-part-only-artix7_100t`
+
+### Step 5 ###
```
litghost, line 87: When this is committed, add a comment to https://github.com/SymbiFlow/prjxray/issues/1476 to remove this step once we have updated the part processing stuff.  The instructions here also need to be updated as part of https://github.com/SymbiFlow/prjxray/issues/1475 too

NEWPART.md
```diff
@@ -0,0 +1,203 @@
...
+```
+git push origin master  // or whatever branch you're using
+```
+
+Then make a pull request.  
+Navigate to the GitHub page for your fork, and click the "New pull request" button.
+Just making the request will kick off Kokoro runs.
+
+
+# BUT WAIT, WHAT ABOUT THE DB #
```
litghost, line 171: This section could use a tidy.

NEWPART.md
```diff
@@ -0,0 +1,203 @@
...
+git add Makefile settings/artix7_100t.sh
+git status
+git commit --signoff 
+```
+
+### Step 8 ###
+
+Push to GitHub:
+```
+git push origin master  // or whatever branch you're using
```
litghost, line 163: I generally recommend that users create a new branch when submitting PRs with a useful branch name.

NEWPART.md
```diff
@@ -0,0 +1,203 @@
...
+_If you are **just** adding a new package for a device that is already handled, you can skip Steps 2..4!_
+
+
+### Step 0 ###
+Fork a copy of https://github.com/SymbiFlow/prjxray on GitHub (go to the page, click "Fork" button, select your own workspace).
+
+Clone your fork (replace "tcal-x" with your GitHub ID):
+
+```
+git clone git@github.com:tcal-x/prjxray.git
```
mkurc-ant, line 16: You may just point to the SymbiFlow/prjxray. There will be no need for instructing a user to replace your github id.

mkurc-ant (commented):
You have written this doc more like a report of what you did and how. In my opinion an instruction should be written in a more impersonal way.

NEWPART.md
tcal-x: Yeah, makes sense.   The question is, /only/ in docs as .rst, or also have an .md version?   I noticed that the getting started readme exists in both formats.

NEWPART.md
tcal-x: 001 must produce the parts yaml for 005.    But yeah, now when I run the target `db-part-only-<part>`, I see that 005 waits for 001 to finish.  Do you think that respecting the dependency was added recently, or did I just see some fluke, maybe due to a restart, where 005 started when the parts yaml wasn't there?   If this should never happen, I'll take out the section.

litghost: I believe there shouldn't be an issue here, and if there was we'd want to file a bug, replicate the issue and then fix it.

NEWPART.md
tcal-x: I'm updating the instructions to create a branch with a useful name.

NEWPART.md
tcal-x: I've replaced "tcal-x" with "<yourGithubId>", since I don't want them to start hacking directly on their clone of SymbiFlow's repo.   Yes, if they know enough git they will know how to update things so they can commit to their own fork later, but it seems simpler if they start with their own fork...

NEWPART.md
tcal-x: I'll ask for re-reviews after updating the content here, and then when it looks good, I'll convert it to an .rst file under docs/.

NEWPART.md
tcal-x: Yes, good catch.   I'll leave this comment unresolved, and add a comment to those PRs when this doc's final location is determined.

tcal-x: I also need to check if https://github.com/SymbiFlow/symbiflow-arch-defs/pull/1791 affects my instructions (under "Database Updates")

NEWPART.md
tcal-x: @mkurc-ant , @litghost , the new doc (docs/newpart.rst) is on readthedocs, ready for review.   Ignore the NEWPART.md file; I'll delete it before merging.   Start here:  [https://symbiflow--1486.org.readthedocs.build/projects/prjxray/en/1486/db_dev_process/index.html](https://symbiflow--1486.org.readthedocs.build/projects/prjxray/en/1486/db_dev_process/index.html) (to see where I've put the new page in the hierarchy), then click on "Guide to adding...".

README.md
```diff
@@ -200,6 +200,10 @@ We have also started experimenting with the Kintex-7 parts.
 The aim is to eventually document all parts in the Xilinx 7-series FPGAs but we
 can not do this alone, **we need your help**!
 
+## Adding a new part to an existing family
+
+We have written a [detailed guide](./NEWPART.md) that walks through the process
```
litghost, line 205: This should reference the RST version of the doc?

NEWPART.md
litghost: Is it time to delete this file?

docs/db_dev_process/newpart.rst
```diff
@@ -0,0 +1,313 @@
...
+   git clone git@github.com:<yourUserID>/prjxray.git
+   cd prjxray
+   git checkout -b <new_branch_name>
+
+Step 1
+~~~~~~
+
+Follow the Project X-Ray developer setup instructions in the
+`documentation <https://symbiflow.readthedocs.io/en/latest/prjxray/docs/db_dev_process/readme.html>`__,
+up through Step 7, Option 1. That is, source the
```
litghost, line 48: I don't believe it sources this particular script, it just invokes it.

litghost (commented):
Overall looks great!  Let's cleanup the handful of comments, and get this merged!

litghost (approved):
LGTM

PR #1492, opened 2020-11-10
Partial bitstream support

tmichalak:
This PR adds partial bitstream generation support for Series7.


lib/include/prjxray/xilinx/configuration.h
```diff
@@ -46,6 +46,15 @@ class Configuration {
 	    const PacketData& packet_data,
 	    absl::optional<typename ArchType::Part>& part);
 
+	// Creates the configuration package for partial configuration
+	// which is later on used by the bitstream writer to generate
+	// the partial bitstream file.
+	static void createPartialConfigurationPackage(
```
litghost, line 52: This comment is missing some important information.  For example why is `far_address` even an argument if `PacketData` contains the `far_address`.  Does this function work if `PacketData` spans multiple frames?  If `PacketData` spans multiple frames, what is the right thing to do?

kowalewskijan: There are some more changes I've pushed recently:
* I removed the `far_address` from arguments because we should handle the FAR addreses for `CLB_IO_CLK` blocks and `BRAMs` separately, in the result we pass a map with data and address for them. This step is handled [here](https://github.com/SymbiFlow/prjxray/pull/1492/files#diff-48d090fe4cd12b7012bdf68a59cd76e3b76e1d7726e9ff33876cd9b5f487d30eR78-R104) and later we use this data to generate proper FAR/FDRI for each of them [here](https://github.com/SymbiFlow/prjxray/pull/1492/files#diff-2f23cb85515d06553f4907136937359a07cb057275d2c8de6c0e1382569079f1R360-R379).
* I added a new definition of [addMissingFrames](https://github.com/SymbiFlow/prjxray/pull/1492/files#diff-5727fcabeaf95213aea76bb536fcb114eee598735fc6e96d6697c2fbac9a5ddcR144-R149) function to handle address gaps in a partial reconfiguration region.
* To handle the PR region correctly `CLB_IO_CLK` and `BRAM` address ranges should be defined as [arguments](https://github.com/SymbiFlow/prjxray/pull/1492/files#diff-5727fcabeaf95213aea76bb536fcb114eee598735fc6e96d6697c2fbac9a5ddcR143-R149) to `xc7frames2bit`

Issue #1504, opened 2020-11-24
Inconsistent project documentation?


mithro:
The prjxray sphinx docs should be identical when viewed;
 - Directly via prjxray
 - When part of the symbiflow documentation

The documentation when viewed at https://symbiflow.readthedocs.io/projects/prjxray/en/latest/# has a fairly short list under items;

![Screenshot from 2020-11-24 09-45-46](https://user-images.githubusercontent.com/21212/100132175-e54d6400-2e39-11eb-8762-5538bebb0a00.png)

Yet, when you view the documentation at https://symbiflow.readthedocs.io/en/latest/ you get a much longer output;

![Screenshot from 2020-11-24 09-46-24](https://user-images.githubusercontent.com/21212/100132222-f4ccad00-2e39-11eb-97d1-41482b5cdebc.png)

mithro:
@tcal-x - This is what you discovered?

rw1nkler:
Probably this is due to missing links in the documentation when building on RTD:
https://github.com/SymbiFlow/symbiflow-docs/blob/master/source/conf.py#L348

I will check that

tcal-x:
@mithro It was more that on Prjxray docs, after "fuzzers", "minitests", "tools", then my new "Guide.." appeared.    On Symbiflow docs, my new "Guide..." doesn't appear at the same point.    So perhaps I need to find another file in which to add a link to the new guide.

tcal-x:
To be clear, this issue is unrelated to my original issue, which is that the link for "Guide to adding a new device to an existing family" is missing from the Symbiflow version of the docs.   The underlying reason for my issue may be https://github.com/SymbiFlow/symbiflow-docs/issues/395 , which blocks the prjxray submodule update on symbiflow-docs.

Looking at _this_ PR, now prjxray doc pages have the same section repeated at different points in the hierarchy -- is this intended?   Specifically, https://symbiflow--1505.org.readthedocs.build/projects/prjxray/en/1505/db_dev_process/readme.html is linked both from under "Getting Started" and also from under "Database Development Process".

However, without this PR (that is, on the current prjxray docs "latest"), this material  ("Quickstart Guide", "C++ Development", "Process", ...) doesn't appear _anywhere_ that I can see.

So we are jumping from not having the material at all, to having it appear at two places in the hierarchy.   I will need to look at it more, and maybe look at the history, to understand what is going on.

tcal-x:
[...]

PR #1505, opened 2020-11-24
Fix links generation for markdown-code-symlinks

rw1nkler:
This PR fixes the missing symbolic links generation during building the Sphinx documentation
Resolves https://github.com/SymbiFlow/prjxray/issues/1504

https://github.com/SymbiFlow/prjxray/pull/1513 should be merged first.

### DIFF
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -69,5 +69,5 @@ links: fuzzers-links minitests-links
 
 # Catch-all target: route all unknown targets to Sphinx using the new
 # "make mode" option.  $(O) is meant as a shortcut for $(SPHINXOPTS).
-%: Makefile links
+%: Makefile
 	@$(SPHINXBUILD) -M $@ "$(SOURCEDIR)" "$(BUILDDIR)" $(SPHINXOPTS) $(O)
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -30,6 +30,8 @@ import recommonmark
 #
 import os
 import sys
+import subprocess
+
 sys.path.insert(0, os.path.abspath('.'))
 from markdown_code_symlinks import LinkParser, MarkdownSymlinksDomain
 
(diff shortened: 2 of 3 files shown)


mithro (approved):
Can you use the same approach RTD environment detection for `make links` that other repo does?

rw1nkler:
> Can you use the same approach RTD environment detection for make links that other repo does?

The other repo calls `make-links` target directly (in the same way as in this PR):
https://github.com/SymbiFlow/symbiflow-docs/blob/master/source/conf.py#L346

rw1nkler:
@tcal-x can you confirm that this PR generates the documentation as you expected?

tcal-x (approved):
As I mentioned in the connected issue, there is some duplication here (copied from the Symbiflow version).   But that can be addressed separately, or ignored (sometimes it makes sense to duplicate).   And this is much improved, since it fixes a page that is empty except for title and sidebar.

lib/include/prjxray/xilinx/bitstream_reader.h
```diff
@@ -93,8 +93,7 @@ class BitstreamReader {
```
mithro, line 93: Why are there changes to this file?

mithro (commented):
There seem to be two random changes in this pull request? Otherwise LGTM.

lib/include/prjxray/xilinx/bitstream_reader.h
tcal-x: Hi @rw1nkler , I see CI had started to fail due to formatting after you rebased on main.   Did the format issue come from main?  And if so, how did it get there?   Or is it a new formatting check?

rw1nkler: This is due to the wrong formatting merged in the current master.
I will prepare a separate PR with the changes.

rw1nkler: https://github.com/SymbiFlow/prjxray/pull/1513

mithro (approved):
Merge if CI is green.

There was an alternative rst render that @xobs was using (rather than recommonmark) which seemed like potentially a better option we should investigate.

PR #1506, opened 2020-11-27
Fixes get_node_for_wire by initializing self.node

syed-ahmed:
The method `get_node_for_wire` fails in `node_model.py` when called for the first time and errors out with:
```
AttributeError: 'NoneType' object has no attribute 'items'
```
Example usage:
```
db = prjxray.db.Database(args.db_root, args.part)
node_model = db.node_model(progressbar_utils.progressbar)
test = node_model.get_node_for_wire('BRAM_L_X4Y0','BRAM_EE4B0_4')
```

This patch fixes the bug by calling `self._build_nodes()` when `self.nodes` is `None`. Moreover, the initialization of `self.wire_to_node_map` is changed to an empty dictionary, otherwise `assert tile_wire not in self.wire_to_node_map` fails with:
```
TypeError: argument of type 'NoneType' is not iterable
```

### DIFF
--- a/prjxray/node_model.py
+++ b/prjxray/node_model.py
@@ -45,7 +45,7 @@ class NodeModel():
 
         self.nodes = None
 
-        self.wire_to_node_map = {}
+        self.wire_to_node_map = None
 
         if progressbar is None:
             self.progressbar = lambda x: x
@@ -143,6 +143,8 @@ class NodeModel():
         return self.nodes[tile, wire]
 
     def _build_wire_to_node_map(self):
+        self.wire_to_node_map = {}
+
         if self.nodes is None:
             self._build_nodes()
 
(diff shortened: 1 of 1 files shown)


prjxray/node_model.py
```diff
@@ -45,7 +45,7 @@ def __init__(
 
         self.nodes = None
 
-        self.wire_to_node_map = None
+        self.wire_to_node_map = {}
```
litghost, line 48: Leave this as `None`.  Instead change `_build_wire_to_node_map` to initialize to the empty map.

litghost (changes requested):
Comments in

syed-ahmed:
> Comments in

Addressed :)

PR #1533, opened 2021-01-05
disassembler: sort remaining unknown bits

acomodi:

To get a better understanding on the differing unknown bits from two different bitstreams it is helpful to have all the missing ones sorted.

It happens that, two bitstreams that have many unknown bits and differ only in one of these unkown bits, have a completely unsorted output when using the `--verbose` option, making it hard to track where the differences are.


litghost (approved):
LGTM, merge if green

PR #1547, opened 2021-01-14
gtp_common: add to tilegrid

acomodi:

For the time being, the GTP_COMMON is restricted only to the artix7 devices. We can enable them for the other families in a second moment.

### DIFF
--- a/fuzzers/005-tilegrid/fuzzaddr/common.mk
+++ b/fuzzers/005-tilegrid/fuzzaddr/common.mk
@@ -16,5 +16,8 @@ $(SPECIMENS_OK):
 clean:
 	rm -rf $(BUILD_DIR)
 
+clean_all:
+	rm -rf build_*
+
 .PHONY: database clean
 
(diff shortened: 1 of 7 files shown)


acomodi:
@litghost  The error should be solved with the latest update.

Basically, the 200T device has `GTP_COMMON_MID_[LEFT|RIGHT]` tiles which have bits in frames 0 and 1, instead of 28 and 29 as for the `GTP_COMMON`.

To have the tilegrid fuzzer as generic as possible, I have added an `auto-alignment` for the baseaddress, so that we do not need special if-else to determine whether we are on the 200T or 50/100T.

acomodi:
> Use the AUTO_ALIGN style from prjuray

@litghost Done

fuzzers/005-tilegrid/fuzzaddr/common.mk
```diff
@@ -14,7 +14,7 @@ $(SPECIMENS_OK):
 	touch $@
 
 clean:
```
litghost, line 16: We need a "clean_all" with this change.

acomodi: Done

PR #1548, opened 2021-01-14
gtp_common: add conf fuzzer

acomodi:
This PR is based on top of https://github.com/SymbiFlow/prjxray/pull/1547 is still a WIP for the following reasons:

- requires some docstrings
- need to add support for the `IS_INVERTED` features.

The idea of this fuzzer is to have a json file containing a list of parameters of four kinds:
- *INT*: This is an integer parameter that requires X number of bits.
- *BIN*: This is a parameter with X numbers of required bits

Given that the GTP common does not have any string parameter, the above parameter types are the only ones.

The json file contains also an `encoding` field, which, as far as I have understood, should be interpreted as the binary representation of the value assigned, and this is relevant mostly for the *INT* parameter type.

E.g.:
The `PLL0_REFCLK_DIV` parameter has two possible values, 1 and 2. The encoding of the value one is 16 and the parameter has 5 possible digits. This means that the value one should be encoded as `00100`

### DIFF
--- a/fuzzers/063-gtp-common-conf/Makefile
+++ b/fuzzers/063-gtp-common-conf/Makefile
@@ -1,0 +1,27 @@
+# Copyright (C) 2017-2020  The Project X-Ray Authors.
+#
+# Use of this source code is governed by a ISC-style
+# license that can be found in the LICENSE file or at
+# https://opensource.org/licenses/ISC
+#
+# SPDX-License-Identifier: ISC
+N ?= 20
+
+include ../fuzzer.mk
+
+database: build/segbits_gtp_common.db
+
+build/segbits_gtp_common.rdb: $(SPECIMENS_OK)
+	${XRAY_SEGMATCH} -o build/segbits_gtp_common.rdb $(addsuffix /segdata_gtp_common.txt,$(SPECIMENS))
+
+build/segbits_gtp_common.db: build/segbits_gtp_common.rdb
+	${XRAY_DBFIXUP} --db-root build --zero-db bits.dbf \
+		--seg-fn-in build/segbits_gtp_common.rdb \
+		--seg-fn-out build/segbits_gtp_common.db
+	${XRAY_MASKMERGE} build/mask_gtp_common.db $(addsuffix /segdata_gtp_common.txt,$(SPECIMENS))
+
+pushdb:
+	${XRAY_MERGEDB} gtp_common build/segbits_gtp_common.db
+	${XRAY_MERGEDB} mask_gtp_common build/mask_gtp_common.db
+
+.PHONY: database pushdb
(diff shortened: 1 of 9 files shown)


acomodi:
@litghost @mithro This should document the whole `GTPE2_COMMON` site, including the `IS_<port>_INVERTED` and, what appears to be a `IN_USE` feature, alongside all the configuration parameters.

acomodi:
> Please rebase this PR

Done

fuzzers/063-gtp-common-conf/Makefile
```diff
@@ -0,0 +1,53 @@
...
+
+SPECIMENS := $(addprefix ${BUILD_DIR}/specimen_,$(shell seq -f '%03.0f' $(N)))
+SPECIMENS_OK := $(addsuffix /OK,$(SPECIMENS))
+FUZDIR ?= ${PWD}
+
+
+all: database
+
+# generate.sh / top_generate.sh call make, hence the command must
+# have a + before it.
```
litghost, line 23: This comment indicates that `top_generate.sh` line should have the `+`, but it is not present?

litghost:
@acomodi Does this fuzzer need to run for A50/A100/A200 because of the GTP_COMMON versus GTP_COMMON_LEFT/RIGHT?  If so, I don't see Makefile changes to do those invocations?

acomodi:
>Does this fuzzer need to run for A50/A100/A200 because of the GTP_COMMON versus GTP_COMMON_LEFT/RIGHT?

Indeed

> I don't see Makefile changes to do those invocations?

@litghost I have moved those changes in the `pushdb.sh` script, which, according to the names of the segdata files, will produce the segbit_<gtp_tile>.db files accordingly.

Also, I have used `find` to gather all the segdata files from the part-specific build directory.

There is one assumption here though, which is that there cannot be mixed GTP_COMMON and GTP_COMMON_MID_LEFT/RIGHT tiles.

I have tested the fuzzer locally for all the three device families and everything went as expected, with segbits files for each GTP_COMMON kinds.

litghost:
> > I don't see Makefile changes to do those invocations?
> 
> @litghost I have moved those changes in the `pushdb.sh` script, which, according to the names of the segdata files, will produce the segbit_<gtp_tile>.db files accordingly.

The change to `pushdb.sh` means that this fuzzer works on A50/A100/A200, but it doesn't mean that the Artix 7 fuzzer will actually invoke the fuzzer as needed.  Those are the changes that are missing...

acomodi:
@litghost I am not sure I follow, but the `063-gtp-common-conf` fuzzer invocation is added to the main fuzzers Makefile, and I have marked it as `part` instead of `all` so that it runs for every part.

[Here](https://source.cloud.google.com/results/invocations/ee8026ca-77c4-47c3-8dc1-3f224c2a22db/targets) is the previous successful run, in which it is visible that the fuzzer ran for all the three different artix parts.

litghost:
> [Here](https://source.cloud.google.com/results/invocations/ee8026ca-77c4-47c3-8dc1-3f224c2a22db/targets) is the previous successful run, in which it is visible that the fuzzer ran for all the three different artix parts.

Indeed, great!

litghost (approved):
LGTM, merge if green.

PR #1549, opened 2021-01-14
database: artix7: Fix wrong device fabric mapping

dnltz:
The xc7a35t devices do not use the xc7a50t fabric.

Signed-off-by: Daniel Schultz <d.schultz@phytec.de>


litghost:
We had evidence in the past that a50t and a35t fabric is the same.  Why do you believe the fabric is different?

dnltz:
Yes, you wrote this in the other issue, sorry.

PR #1550, opened 2021-01-14
setup.py: add dependencies

kgugala:
This PR adds python dependencies to setup.py. In the binary flow we install this using pip and we were missing the deps (accidentally we had them installed by some other packages).


setup.py
```diff
@@ -24,6 +24,25 @@
     long_description_content_type="text/markdown",
     url="https://github.com/SymbiFlow/prjxray",
     packages=['prjxray'],
+    install_requires=[
+        'sdf-timing @ git+https://github.com/symbiflow/python-sdf-timing',
+        'fasm @ git+https://github.com/symbiflow/fasm',
```
litghost, line 29: Don't use an `@` in `setup.py`.  If you need to pull in a git package, do it via `requirements.txt`.

setup.py
```diff
@@ -24,6 +24,25 @@
...
+        'parse',
+        'progressbar2',
+        'pyjson5',
+        'pytest',
+        'pyyaml',
+        'scipy>=1.2.1',
+        'simplejson',
+        'sympy',
+        'textx',
+        'yapf==0.24.0',
```
litghost, line 44: `yapf` isn't need for end users?

setup.py
```diff
@@ -24,6 +24,25 @@
     long_description_content_type="text/markdown",
     url="https://github.com/SymbiFlow/prjxray",
     packages=['prjxray'],
+    install_requires=[
+        'sdf-timing @ git+https://github.com/symbiflow/python-sdf-timing',
+        'fasm @ git+https://github.com/symbiflow/fasm',
+        'intervaltree',
+        'junit-xml',
```
litghost, line 31: junit-xml is definitely not needed in setup.py?

setup.py
kgugala: the problem this PR solves is that those were not installed when we do `pip install git+https://github.com/symbiflow/prjxray` (and we do that in symbiflow-examples). That is why I put those `@` in `setup.py`

setup.py
kgugala: removed

setup.py
litghost: Never put `@` in setup.py, end of story.  It causes all kinds of problems.  `@` is an extremely narrow pin, and causes problems.  The longer term solution is for these packages to end up on PyPi, etc, but that is a longer term problem.

setup.py
kgugala: removed

setup.py
ajelinski: > `@` is an extremely narrow pin

Can you explain why @litghost? Does it tie a dependency like `fasm` to the revision which is currently `master` on `fasm` forever if I install a package with `fasm @ git+https://github.com/symbiflow/fasm` in `install_requires`?

litghost: The `@` is effectively "always use exactly this dependency".  The new pip dependency resolver respects this directly, but it means that it mixes poorly with other configurations.  As a concrete example, if you want to use `git submodule` to pin fasm to a particular commit (for example in symbiflow-examples), that will conflict with the `@` here.

The bottom line is `@` can **never** be used in `setup.py`, period.  There are no acceptable reasons to do so.

ajelinski:
Definitely `pyyaml` is needed in `install_required`. If it was included, I wouldn't need to add it manually to `requirements.txt` files in https://github.com/SymbiFlow/fpga-tool-perf .

mithro:
> Definitely `pyyaml` is needed in `install_required`. If it was included, I wouldn't need to add it manually to `requirements.txt` files in https://github.com/SymbiFlow/fpga-tool-perf .

Specifying `pyyaml` in the `install_requires` is very different to specifying `pyyaml @ <location>`.

setup.py
```diff
@@ -24,6 +24,23 @@
     long_description_content_type="text/markdown",
     url="https://github.com/SymbiFlow/prjxray",
     packages=['prjxray'],
+    install_requires=[
+        'sdf-timing @ git+https://github.com/symbiflow/python-sdf-timing',
```
mithro, line 28:
```suggestion
        'sdf-timing',
```

setup.py
```diff
@@ -24,6 +24,23 @@
...
+        'fasm @ git+https://github.com/symbiflow/fasm',
```
mithro, line 29:
```suggestion
        'fasm',
```

setup.py
```diff
@@ -24,6 +24,23 @@
...
+        'ordered-set',
+        'parse',
+        'progressbar2',
+        'pyjson5',
+        'pytest',
+        'pyyaml',
+        'scipy>=1.2.1',
+        'simplejson',
+        'sympy',
+        'textx',
```
mithro, line 42: Are all these dependencies *actually* needed? These seem like the dependencies for the whole prjxray fuzzer tooling not the prjxray python library?

setup.py
litghost: `fasm` is now on PyPi (per https://github.com/SymbiFlow/fasm/pull/59).  If someone can do the same work for sdf-timing, the `@` can go away from this completely!

mithro:
The FASM library should now be installable via both PyPi and directly from GitHub.

The `install_requires` in `setup.py` should **only** specify `fasm`.

The `requirements.txt` can specify *any* of `fasm` (which will be downloaded via PyPi) or something like `git+https://github.com/SymbiFlow/fasm.git#egg=fasm` (downloaded and installed from GitHub) or `-e ./third_party/fasm` (installed from local third party submodule).

setup.py
litghost: Poke, this comment is valid, and should be fixed.

acomodi: Should be fixed now. The only dependencies added are ones that are needed by scripts in the prjxray directory scripts

setup.py
```diff
@@ -24,6 +24,14 @@
     long_description_content_type="text/markdown",
     url="https://github.com/SymbiFlow/prjxray",
     packages=['prjxray'],
+    install_requires=[
+        'fasm',
+        'intervaltree',
+        'numpy',
+        'pyjson5',
```
litghost, line 31: Where is `pyjson5` used?

litghost: And what happened to `sdf-timing`?  Is it not required?

acomodi: > Where is pyjson5 used?

https://github.com/SymbiFlow/prjxray/blob/master/prjxray/lib.py
https://github.com/SymbiFlow/prjxray/blob/master/prjxray/node_lookup.py

> And what happened to sdf-timing

This seems not to be used by any of the scripts belonging to the python package

litghost: > > Where is pyjson5 used?
> 
> https://github.com/SymbiFlow/prjxray/blob/master/prjxray/lib.py
> https://github.com/SymbiFlow/prjxray/blob/master/prjxray/node_lookup.py

Both of these probably shouldn't be in the prjxray library, as they operate on data that is only present in prjxray fuzzer directories.   Let's leave pyjson5 in for now, but add a README to isolate those files and eventually remove this dependency.

setup.py
```diff
@@ -24,6 +24,15 @@
     long_description_content_type="text/markdown",
     url="https://github.com/SymbiFlow/prjxray",
     packages=['prjxray'],
+    install_requires=[
+        'fasm',
+        'intervaltree',
+        'numpy',
+        # FIXME: isolate lib.py and node_lookup.py from the package to remove this dependency
```
litghost, line 31: It's specifically `lib.NodeLookup`, not all of `lib.py` that is a problem.  Also please fill an issue to track fixing this, and link it here.

litghost (approved):
LGTM, merge if green

PR #1556, opened 2021-01-18
minitest: add lite-sata minitest to verify GTP status

acomodi:

This PR is to add a LiteX minitest that makes use of liteSATA, which implements the GTP-related tiles

### DIFF
--- a/minitests/litex/nexys_video_sata/Makefile
+++ b/minitests/litex/nexys_video_sata/Makefile
@@ -1,0 +1,58 @@
+# Copyright (C) 2017-2020  The Project X-Ray Authors.
+#
+# Use of this source code is governed by a ISC-style
+# license that can be found in the LICENSE file or at
+# https://opensource.org/licenses/ISC
+#
+# SPDX-License-Identifier: ISC
+
+SHELL = bash
+
+SYNTH ?= vivado
+YOSYS = $(XRAY_DIR)/third_party/yosys/yosys
+PART = xc7a200tsbg484-1
+PROJECT_NAME = sata
+
+clean:
+	@find . -name "build-par.*" | xargs rm -rf
+	@find . -name "build-syn.*" | xargs rm -rf
+	@rm -f *.edif
+	@rm -f *.bit
+	@rm -f *.bin
+	@rm -f *.log
+	@rm -f *.dcp
+
+help:
+	@echo "Usage: make all [SYNTH=<vivado/yosys>]"
+
+.PHONY: clean help
+
+$(YOSYS):
+	cd $(XRAY_DIR)/third_party/yosys && make config-gcc && make -j$(shell nproc)
+
+build.ok:
+	./src/litex-boards/litex_boards/targets/nexys_video.py --with-sata --integrated-rom-size 0x10000
+	touch build/build.ok
+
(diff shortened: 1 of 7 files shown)


mithro:
@acomodi - MIght be worth looking at https://github.com/hamsternz/DisplayPort_Verilog

acomodi:
@mithro Sure, that is yet another good candidate for a minitest

litghost (changes requested):
This minitest needs a README?

minitests/litex/nexys_video_sata/Makefile
```diff
@@ -0,0 +1,62 @@
...
+clean:
+	@find . -name "build-par.*" | xargs rm -rf
+	@find . -name "build-syn.*" | xargs rm -rf
+	@rm -f *.edif
+	@rm -f *.bit
+	@rm -f *.bin
+	@rm -f *.log
+	@rm -f *.dcp
+
+help:
```
litghost, line 25: Looks like the "all" target is missing:
```
$ make help
Usage: make all [SYNTH=<vivado/yosys>]
$ make all SYNTH=vivado
make: *** No rule to make target 'all'.  Stop.
```

minitests/litex/nexys_video_sata/Makefile
```diff
@@ -0,0 +1,62 @@
...
+# SPDX-License-Identifier: ISC
+
+SHELL = bash
+
+SYNTH ?= vivado
+YOSYS = $(XRAY_DIR)/third_party/yosys/yosys
+PART = xc7a200tsbg484-1
+PROJECT_NAME = sata
+
+clean:
```
litghost, line 16: The first target in a Makefile is the default target.  I don't think `clean` is a good default target.  Recommend putting the `all` target first?  If target ordering matters, you can use `.DEFAULT_GOAL` (see https://www.gnu.org/software/make/manual/html_node/Special-Variables.html#Special-Variables)

acomodi:
> This minitest needs a README?

Indeed, I forgot to put this as WIP.

I'll need to also fix the litex modules to have a reproducible test and also fix some issues in the Makefile.

litghost (changes requested):
Looks like the path to the venv is assumed in the makefile.  For example, if this makefile is invoked used the prjxray venv, you get the following error:
```
./src/litex-boards/litex_boards/targets/nexys_video.py --with-sata --integrated-rom-size 0x10000
bash: line 1: ./src/litex-boards/litex_boards/targets/nexys_video.py: No such file or directory
make: *** [Makefile:38: build/build.ok] Error 127
```

minitests/litex/nexys_video_sata/README.md
```diff
@@ -0,0 +1,29 @@
...
+
+The minitest synthesis step can be performed with Yosys or Vivado.
+
+The final FASM file with the `unknown bits` can be obtained by running the following:
+
+```bash
+make all
+```
+
+All the pre-requisites (LiteX, Yosys, etc.) are automatically installed/built. It is required though to have Vivado installed in the system.
```
litghost, line 29: This project requires a RISC-V tool chain on the path, so this is not true?

acomodi: Fixed in Makefile

PR #1551, opened 2021-01-15
005-tilegrid: Add GTP_CHANNEL tilegrid fuzzer

tmichalak:
This PR adds the tilegrid fuzzer for the GTP_CHANNEL tiles.

PR #1558, opened 2021-01-20
Add GTP_CHANNEL fuzzer

kowalewskijan:
The fuzzer introduced in this PR uses the same approach as explained in https://github.com/SymbiFlow/prjxray/pull/1548
Additional changes like the tilegrid fuzzer and the auto-alignment feature are taken from https://github.com/SymbiFlow/prjxray/pull/1551


fuzzers/064-gtp-channel-conf/generate.py
```diff
@@ -0,0 +1,105 @@
...
+BOOL = "BOOL"
+STR = "STR"
+
+
+def bitfilter(frame, bit):
+    # Filter out interconnect bits.
+    if frame not in [28, 29, 30, 31, 0, 1, 2, 3]:
+        return False
+
+    return True
```
acomodi, line 29: There should be two bitfilter functions, as the GTP_CHANNEL can have different frames based on the tile. See https://github.com/SymbiFlow/prjxray/pull/1548/commits/c59c486959b48ae5e449816f077f9e249c7ac45a.

fuzzers/064-gtp-channel-conf/Makefile
```diff
@@ -0,0 +1,50 @@
+# Copyright (C) 2017-2020  The Project X-Ray Authors.
```
acomodi, line 1: The Makefile should handle the possibility to have different build directories for each part of the same family. This because the fuzzer must be run for each part, given the difference in frames for some tiles. See the Makefile for the [GTP_COMMON](https://github.com/SymbiFlow/prjxray/blob/c59c486959b48ae5e449816f077f9e249c7ac45a/fuzzers/063-gtp-common-conf/Makefile) which should similar

fuzzers/064-gtp-channel-conf/top.py
```diff
@@ -0,0 +1,128 @@
...
+    db = Database(util.get_db_root(), util.get_part())
+    grid = db.grid()
+    for tile_name in sorted(grid.tiles()):
+        loc = grid.loc_of_tilename(tile_name)
+        gridinfo = grid.gridinfo_at_loc(loc)
+
+        if gridinfo.tile_type not in [
+                "GTP_CHANNEL_0",
+                "GTP_CHANNEL_0_MID_LEFT",
+        ]:
```
acomodi, line 38: I think here we should have all the possible GTP_CHANNEL types and only one is selected. If we allow all the GTP_CHANNEL tiles to be used, we might reduce the number of specimens, as more data is generated in one specimen.

acomodi (changes requested):
Comments in

litghost:
Please rebase this PR

fuzzers/064-gtp-channel-conf/attrs.json
```diff
@@ -0,0 +1,3646 @@
+{
```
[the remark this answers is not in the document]
kowalewskijan: fixed

fuzzers/064-gtp-channel-conf/generate.py
```diff
@@ -0,0 +1,105 @@
...
+                            site, "{}.{}".format(param, param_value),
+                            value == param_value)
+
+            for param, invert in [("TXUSRCLK", 1), ("TXUSRCLK2",
+                                                    1), ("TXPHDLYTSTCLK", 1),
+                                  ("SIGVALIDCLK", 1), ("RXUSRCLK",
+                                                       1), ("RXUSRCLK2", 1),
+                                  ("DRPCLK", 1), ("DMONITORCLK",
+                                                  1), ("CLKRSVD0",
+                                                       1), ("CLKRSVD1", 1)]:
```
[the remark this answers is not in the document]
kowalewskijan: I didn't get any additional bits for `INV_` case. I reworked this part to use only `ZINV_` case.

fuzzers/064-gtp-channel-conf/generate.py
kowalewskijan: done

fuzzers/064-gtp-channel-conf/Makefile
kowalewskijan: Reworked to use `${BUILD_DIR}`

fuzzers/064-gtp-channel-conf/top.py
kowalewskijan: I added more types, so in the result we have multiple `GTPE2_CHANNEL` instantiations for different LOCs in a single `top.v` . I reduced number of specimens to 10.

acomodi (changes requested):
LGTM, only thing is that this PR needs a rebase with master.

kowalewskijan:
> only thing is that this PR needs a rebase with master.

Done

acomodi (approved):
LGTM

PR #1560, opened 2021-01-22
minitest: add hamsternz display port

acomodi:


mithro (approved):
LGTM!

litghost:
@acomodi This minitest seems to expose more of the GTP logic than the SATA test case (714 unknown bits in DisplayPort test vs 357 unknown bits in SATA).

With the latest results, how many unknown bits are left in this DisplayPort minitest?

acomodi:
@litghost With the latest fuzzers for the GTP merged, the count is:

- SATA: 15 unknown bits, which correspond to:
  - Clock PIPs in the `GTP_COMMON_MID` (for the 200T only
  - `IBUDFS_GTE2 --> GTP_COMMON` direct connection
  - PIPs in the `GTP_INT_INTERFACE` tile that seem to increase the signal path, as it bounces to a `DELAY` node (unsure how this is triggered)
 
For DisplayPort there are a total of 53 unknown bits. Most of these, though, do not correpsond to the GTP, but rather to the IOB/INT/HCLK segment of another clock region. I am not sure why these show up though. If we take only the unknowns corresponding to the GTP, we are down to 8 unknown bits (PIPs and `IBUFDS_GTE2 --> GTP_COMMON`).

litghost:
> For DisplayPort there are a total of 53 unknown bits. Most of these, though, do not correpsond to the GTP, but rather to the IOB/INT/HCLK segment of another clock region. I am not sure why these show up though. If we take only the unknowns corresponding to the GTP, we are down to 8 unknown bits (PIPs and `IBUFDS_GTE2 --> GTP_COMMON`).

That likely points to a missing IOB feature.  This probably needs to be fixed, but might not block the GTP/PCIe efforts.

litghost:
> * PIPs in the `GTP_INT_INTERFACE` tile that seem to increase the signal path, as it bounces to a `DELAY` node (unsure how this is triggered)

Presumably this is just the router fixing a hold violation.  You should be able to toggle the `DELAY` by using `route_via` on those paths.

litghost:
> For DisplayPort there are a total of 53 unknown bits.

I was able to replicate this using the latest [prjxray-db](https://github.com/SymbiFlow/prjxray-db/commit/e7663ba6eb651a7255775c7dca4e32483a8cd9fb)!

PR #1563, opened 2021-01-26
065-gtp-common-pips: add fuzzer to document GTP_COMMON_MID pips

acomodi:

This fuzzer is to get the clock-related PIPs corresponding to the `GTP_COMMON_MID_[LEFT|RIGHT]` tile that are used to route signals from the possible following sources:

- `GTP_CHANNEL` RX and TX output clocks
- `IBUFDS_GTE2` Output clock
- Clocks from other Clock Region's sources (e.g. PLL, MMCM, etc.)

The fuzzer requires some fixes, hence the WIP in the title. In fact, it seems that, out of the 168 PIPs, ~30 of them keep having zero candidates, even after many specimens have been generated.

It seems that there are two set bits for each PIP, hence the `-c 2` parameter to segmatch.

There is also to handle the fact that this fuzzer should not run for artix 50T and 100T, as they have no `GTP_COMMON_MID_[LEFT|RIGHT]` tiles. I believe that this fuzzer can be set to be solved for all the parts, and, for the 50 and 100T, the fuzzer just creates the run.ok already, prior verifying some conditions such as "this device does not have any GTP_COMMON_MID tiles".

TODO:
- [x] Solve all PIPs
- [x] Enable all parts' build for this fuzzer

### DIFF
--- a/fuzzers/065-gtp-common-pips/generate.py
+++ b/fuzzers/065-gtp-common-pips/generate.py
@@ -17,9 +17,13 @@ import os.path
 def bitfilter(frame, word):
     word = int(word / 32)
 
+    # Clock-related PIPs in the GTP_COMMON_MID_[LEFT|RIGHT] have bits
+    # from frame 0 to 7, hence, all the other frames are skipped.
     if frame not in [0, 1, 2, 3, 4, 5, 6, 7]:
         return False
 
+    # All the Clock-related PIPs have bits at word 50 in the GTP_COMMON tile.
+    # Filter out all bits not belonging to word 50.
     if word != 50:
         return False
 
--- a/fuzzers/065-gtp-common-pips/generate.py
+++ b/fuzzers/065-gtp-common-pips/generate.py
@@ -27,8 +27,9 @@ def bitfilter(frame, word):
 
 
 def read_pip_data(pipfile, pipdata, tile_ports):
+    part = os.getenv('XRAY_PART')
     with open(os.path.join(os.getenv('FUZDIR'), '..', 'piplist', 'build',
-                           'gtp_common', pipfile)) as f:
+                           'gtp_common_{}'.format(part), pipfile)) as f:
         for l in f:
             tile_type, dst, src = l.strip().split('.')
             if tile_type not in pipdata:
@@ -48,7 +49,8 @@ def main():
     ignpip = set()
     tile_ports = {}
 
-    read_pip_data('gtp_common.txt', pipdata, tile_ports)
+    part = os.getenv('XRAY_PART')
+    read_pip_data('gtp_common_{}.txt'.format(part), pipdata, tile_ports)
 
     print("Loading tags from design.txt.")
     with open("design.txt", "r") as f:
--- a/fuzzers/065-gtp-common-pips/generate.py
+++ b/fuzzers/065-gtp-common-pips/generate.py
@@ -1,0 +1,102 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2017-2020  The Project X-Ray Authors.
+#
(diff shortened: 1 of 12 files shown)


litghost:
@mithro I don't think this is really ready for review yet.

acomodi:
> I don't think this is really ready for review yet.

Yeah, this is still WIP, and requires some more work IMO.

I pushed a `wip` commit that should get us to have no `0 candidates`. The problem seems to be that, if PIPs are used only to route GTP_CHANNEL RX/TX and IBUFDS_GTE2 outputs, than, even with re-routing of the nets, no bit related to those PIPs is generated.

This is quite strange, as, if I get to add a PLL-generated clock, bits corresponding to the GTP-common PIPs are generated.
It is even stranger that the LiteSATA minitest reports unknown bits that should correspond to the PIPs fuzzed in this PR, but in there, there are no clocks routed from PLL/MMCM that should trigger the presence of bits.

I believe that some things might get optimized, out if there is a misuse of the GTP tiles.

In addition, with https://github.com/SymbiFlow/prjxray/pull/1563/commits/444e08c6c327b290c29bd3734f8914e01cec1e62 I am able to get all the PIPs except for one, which keeps reporting `3 candidates`. Probably this issue is still related to the above one.

fuzzers/065-gtp-common-pips/generate.py
```diff
@@ -0,0 +1,104 @@
...
+
+from prjxray.segmaker import Segmaker
+import os
+import os.path
+
+
+def bitfilter(frame, word):
+    word = int(word / 32)
+
+    if frame not in [0, 1, 2, 3, 4, 5, 6, 7]:
```
litghost, line 20: Please add a comment on what is going on with these?

fuzzers/065-gtp-common-pips/generate.tcl
```diff
@@ -0,0 +1,369 @@
...
+
+        set old_origin_dst_wire [get_wires -of_objects $net -filter {TILE_NAME =~ "*GTP_COMMON_MID*"}]
+        if {$old_origin_dst_wire == {}} {
+            continue
+        }
+
+        puts "Rerouting net $net at $tile / $wire (type $tile_type)"
+        puts "Previous target wire: $old_origin_dst_wire"
+
+        set old_origin_node [get_nodes -of_objects $old_origin_dst_wire]
```
litghost, line 236: Comment around this stuff

fuzzers/065-gtp-common-pips/generate.tcl
```diff
@@ -0,0 +1,369 @@
...
+            set old_target_side $group
+        }
+        foreach dst $dsts {
+            set srcs [dict get $todo_map $dst]
+            foreach src $srcs {
+
+                set src_wire [lindex $src 1]
+
+                set is_gtp_net 1
+                if [regexp "HCLK_GTP_CK_MUX.*" $src_wire match group] {
```
litghost, line 247: Comment

litghost (changes requested):
Results look good.  Some part of the TCL script could use some clarification, so I've requested comments.  After those comments are in, I think this is good to merge.

fuzzers/065-gtp-common-pips/generate.tcl
acomodi: Done.
Removed the `old_origin_node` as it was actually unused.

fuzzers/065-gtp-common-pips/generate.py
acomodi: Done

fuzzers/065-gtp-common-pips/generate.tcl
acomodi: Done

Issue #1540, opened 2021-01-13
Investigate PCIE_2_1 PIPs


acomodi:
This issue is to keep track of the information on the PIPs around the PCIE_2_1 primitives for xc7 parts.

### Description

The PCIE_2_1 primitive is a hard block that is half-clock region tall and has many connections with the fabric. By investigating the device view, it results that there are some pips in the `PCIE_INT_INTERFACE` tiles which work a bit like the `BOUNCE` pips in the INT tiles.

![Screenshot from 2021-01-13 12-37-27](https://user-images.githubusercontent.com/44773360/104447511-30633800-559c-11eb-8be5-1766664befda.png)

The image above represents a section of the `PCIE_INT_INTERFACE` that displays said pips.

By further investigation, these pips seem to occupy the 27th frame of the segment in which the PCIE_2_1 primitive is, and that,  by default, the pip allows a direct connection, without *bouncing* to the additional wire before getting to the destination port.

These PIPs should likely require a pce-int-pips fuzzer.

litghost:
Is this complete?

acomodi:
Yes, this was done. Closing.

PR #1564, opened 2021-01-28
Add litepcie minitest

acomodi:
This test requries a new part for the netv2 board. There is an open PR to add all available parts automatically, so, until that is merged, this minitest requires its manual addition.

In addition, this minitest uses the LiteX generated tcl to build the bitstream with Vivado. This yields an issue, such that the DSPs inference cannot be disabled during synthesis, therefore, a portion of the unknown bits results to be belonging to the DSPs.

All the other unknown bits seem to be mostly related to the ["delay"](https://github.com/SymbiFlow/prjxray/issues/1540) PIPs of the PCIE and GTP tiles.

There are a couple of other kinds of unknown bits which are still to be identified.


minitests/litex/litepcie/Makefile
```diff
@@ -0,0 +1,46 @@
...
+litex-install.ok:
+	pip install -r requirements.txt
+	touch litex-install.ok
+
+riscv-gcc.ok:
+	wget -qO- https://static.dev.sifive.com/dev-tools/$(RISCV_DIR).tar.gz | tar -xz
+	touch riscv-gcc.ok
+
+build/netv2/gateware/netv2.bit: litex-install.ok riscv-gcc.ok
+	env PATH=${PATH}:${PWD}/${RISCV_DIR}/bin ${VIRTUAL_ENV}/src/litex-boards/litex_boards/targets/netv2.py --with-pcie --variant a7-100 --build
```
litghost, line 43: This should source `$XRAY_VIVADO_SETTINGS` so that Vivado is on the Path when this is invoked.

acomodi: Sure, done

litghost: The current line doesn't work.

litghost: This works:
```
 build/netv2/gateware/netv2.bit: litex-install.ok riscv-gcc.ok
-       source ${XRAY_VIVADO_SETTINGS} /
-               env PATH=${PATH}:${PWD}/${RISCV_DIR}/bin ${VIRTUAL_ENV}/src/litex-boards/litex_boards/targets/netv2.py --with-pcie --variant a7-100 --build
+       export PATH=${PATH}:${PWD}/${RISCV_DIR}/bin && \
+               source ${XRAY_VIVADO_SETTINGS} && \
+               ${VIRTUAL_ENV}/src/litex-boards/litex_boards/targets/netv2.py --with-pcie --variant a7-100 --build
```

acomodi: I actually put everything in a bash script, but I think it is better to have that in the Makefile

litghost (approved):
LGTM

PR #1574, opened 2021-02-03
005-tilegrid: add pcie_int_interface tile baseaddrs

acomodi:

### DIFF
--- a/fuzzers/005-tilegrid/generate_full.py
+++ b/fuzzers/005-tilegrid/generate_full.py
@@ -549,6 +549,8 @@ def run(json_in_fn, json_out_fn, verbose=False):
     propagate_INT_bits_in_column(database, tiles_by_grid)
     propagate_INT_INTERFACE_bits_in_column(
         database, tiles_by_grid, "GTP_INT_INTERFACE")
+    propagate_INT_INTERFACE_bits_in_column(
+        database, tiles_by_grid, "PCIE_INT_INTERFACE")
     propagate_rebuf(database, tiles_by_grid)
     propagate_IOB_SING(database, tiles_by_grid)
     propagate_IOI_SING(database, tiles_by_grid)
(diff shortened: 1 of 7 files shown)


fuzzers/005-tilegrid/generate_full.py
```diff
@@ -221,6 +221,99 @@ def propagate_INT_bits_in_column(database, tiles_by_grid):
             tile = database[tile_name]
 
 
+def propagate_INT_INTERFACE_bits_in_column(
+        database, tiles_by_grid, int_interface_name):
+    """ Propigate INT offsets up and down INT columns.
```
litghost, line 226: Is this comment correct?  Feels like a copy paste, but unclear?

fuzzers/005-tilegrid/pcie_int_interface/generate.tcl
```diff
@@ -0,0 +1,92 @@
...
+    set params_map [parse_csv]
+    set nets [get_nets]
+
+    dict for { key value } $params_map {
+        if { $value == 0 } {
+            continue
+        }
+
+        if { $key == "" } {
+            puts "Dictionary key is incorrect, continuing"
```
litghost, line 46: When would this ever happen?  Rather than skip a bad key, why not fix the reason it exists?

fuzzers/005-tilegrid/generate_full.py
acomodi: Yeah, this function is similar to the `propagate_INT_bits_in_column`. I have updated the comment which was a leftover, but I might work some more to try and de-duplicate the code, if possible.

fuzzers/005-tilegrid/pcie_int_interface/generate.tcl
acomodi: Fixed the params.csv parsing to avoid having to check the dictionary key

litghost:
This needs a rebase

litghost:
Just waiting for green CI and I'll review.

PR #1575, opened 2021-02-03
Add directory to repository cloning step

### LINKED ISSUE #1520: Quick start documentation should mention in which dir to issue the git submodule update command
In 
https://symbiflow.readthedocs.io/projects/prjxray/en/latest/db_dev_process/readme.html
Step 2 it says to:
git submodule update --init --recursive
But omits to say in which directory.
I guess the top of prjxray?
But not having to guess would be better :)
For the record, I cloned symbiflow-arch-pkgs first and I already
had issued a similar command there, so I was not really shure where the step applied.

tmichalak:
Fixes #1520

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,8 +40,10 @@ shell that is then only used to run Vivado to avoid these problems.
 **Is 2017.2 really required?** Yes, only `2017.2` works. Until Issue #14 is solved, **only** `2017.2` works and will be supported.
 
 ### Step 2: ###
-Pull submodules:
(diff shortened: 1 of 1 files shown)


mithro (approved):
Why wasn't this failing on CI?

Otherwise LGTM.

litghost:
> Why wasn't this failing on CI?
> 
> Otherwise LGTM.

Because CI checks out the repo as the first step.

PR #1573, opened 2021-02-02
Add GTP_INT_INTERFACE PIPs fuzzer

acomodi:
This PR builds on top of https://github.com/SymbiFlow/prjxray/pull/1572 and documents the `DELAY` PIPs of the GTP_INTERFACE tiles.

PR #1576, opened 2021-02-03
062-pcie-int-pips: add DELAY PIPs fuzzer

acomodi:
This fuzzer is much similar to the one in https://github.com/SymbiFlow/prjxray/pull/1573.

Given that there are some DELAY PIPs that are never used and are not connected, there will be some `<const0>` tags. To avoid having the fuzzer to loop indefinetely, there is a max iteration set to 2.


litghost:
This needs a rebase

PR #1584, opened 2021-02-09
063-gtp-common-conf: add DRP_ENABLE feature

acomodi:

This PR enables a feature that is triggered on specific connectivity of the DRP (Dynamic Reconfiguration Port) port in the GTP_CHANNEL primitives.

Basically, this feature should enable the DRP for all of the tiles in the column, as the bits appear in the words belonging to the GTP_COMMON tile.

It is enough to have one only GTP_CHANNEL primitive with all the required DRP ports connected accordingly, to enable this feature.


litghost (approved):
LGTM, merge if green

PR #1585, opened 2021-02-10
065-gtp-common-pips: fix overlapping piplist project names

acomodi:

This fixes an issue discovered in [this master CI build](https://source.cloud.google.com/results/invocations/7379c769-f46e-440a-8777-8d34c075860e).

065- and 065b-gtp-common-pips share the piplist directory and the vivado project needs to be located in separate directories.


litghost:
Good catch

PR #1587, opened 2021-02-12
030-iob: add LVDS_25 and TMDS_33

acomodi:
Signed-off-by: Alessandro Comodi <acomodi@antmicro.com>

PR #1592, opened 2021-02-19
030-iob: improve rdb processing for LVDS and TMDS

acomodi:

Fixes issue found in https://github.com/SymbiFlow/prjxray/pull/1587#issuecomment-781507948

### DIFF
--- a/fuzzers/030-iob/process_rdb.py
+++ b/fuzzers/030-iob/process_rdb.py
@@ -151,7 +151,11 @@ def process_features_sets(iostandard_lines):
                 iostandard, None)]
 
         for iostandard, enum in sites[site]['IN']:
-            if sites[site]['IN_DIFF'][(iostandard, enum)]:
+            _, iostd_type = site
+            if iostd_type == "ONLY_DIFF":
+                sites[site]['IN_DIFF'][(iostandard, enum)] = \
+                        sites[site]['IN'][(iostandard, enum)]
+            elif sites[site]['IN_DIFF'][(iostandard, enum)]:
                 sites[site]['IN_DIFF'][(iostandard, enum)] |= \
                         sites[site]['IN'][(iostandard, enum)]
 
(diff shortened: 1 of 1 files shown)


litghost (commented):
This seems ok, just waiting for CI results.  @acomodi, what testing have you done?

acomodi:
@litghost I have checked with some minitests with bit2fasm (display-port, litex minitests) to see whether the changed features caused problems and it seemed that everything went all right.

There have been some neg bits additions due to the merge of the "only diff" iostandards with the "normal" ones. The change can actually be seen in the zynq CI which completed (should be equal to the artix one).

When this change is pushed in prjxray-db, we will need to update some of the fasm features.

acomodi:
Oh, actually all CIs completed while I was writing :)

litghost:
> @litghost I have checked with some minitests with bit2fasm (display-port, litex minitests) to see whether the changed features caused problems and it seemed that everything went all right.
> 
> There have been some neg bits additions due to the merge of the "only diff" iostandards with the "normal" ones. The change can actually be seen in the zynq CI which completed (should be equal to the artix one).
> 
> When this change is pushed in prjxray-db, we will need to update some of the fasm features.

Can you please also test that LiteX and picosoc bit -> fasm and fasm -> bit (via xc-fasm2bels) work as expected?

litghost:
Also double check the the slew features are working as expected, I see some changes in the harness JSON's, but they look ok.

litghost:
Looks like we lost this feature:
```
-LIOB33.IOB_Y1.LVCMOS12_LVCMOS15_LVCMOS18_LVCMOS25_LVCMOS33_LVTTL.SLEW.FAST !38_16 !38_18 !38_20 !38_22 !39_17 !39_21
```

It was a pure negative feature, but it's worth making sure things make sense still.

acomodi:
@litghost I have tested the LiteX DDR design which has many different IOSTANDARDs and SLEWs, and, starting from a working bitstream, the resulting one in the bit -> fasm -> xc-fasm2bels -> bit flow resulted to be identical and working on HW with the new IOB features changes.

There was also a bug in the process_rdb which resulted in the deletion of the pure negative feature which is now fixed.

PR #1603, opened 2021-03-02
005-tilegrid: fix GTP common frames

acomodi:

Changing max frames for GTP_COMMON from 42 to 32. This was causing a huge cascading errors while running fasm2frames, resulting in a bugged frames file.

### DIFF
--- a/fuzzers/005-tilegrid/add_tdb.py
+++ b/fuzzers/005-tilegrid/add_tdb.py
@@ -108,7 +108,7 @@ def run(fn_in, fn_out, verbose=False):
         ("hclk_cmt", 30, 10),
         ("hclk_ioi", 42, 1),
         ("pcie", 36, 101),
-        ("gtp_common", 42, 101),
+        ("gtp_common", 32, 101),
         ("gtp_channel", 32, 22),
         ("clb_int", int_frames, int_words),
         ("iob_int", int_frames, int_words),


litghost:
> Changing max frames for GTP_COMMON from 42 to 32. This was causing a huge cascading errors while running fasm2frames, resulting in a bugged frames file.

We've had this problem a couple times (in fasm2frames).  Is there a good way to add a sanity check if a frame is pushing into another?  I believe the data in the bitstream yaml files should provide enough context to detect these kind of cascades. @tmichalak

tmichalak:
> > Changing max frames for GTP_COMMON from 42 to 32. This was causing a huge cascading errors while running fasm2frames, resulting in a bugged frames file.
> 
> We've had this problem a couple times (in fasm2frames). Is there a good way to add a sanity check if a frame is pushing into another? I believe the data in the bitstream yaml files should provide enough context to detect these kind of cascades. @tmichalak

Yes, we did have had this problem before in U-Ray. We added a [TileMap](https://github.com/SymbiFlow/prjuray/blob/master/fuzzers/002-tilegrid/util.py#L20) class to extract the frame information from `part.json` that is used when the frame count is `None` in [add_tdb.py](https://github.com/SymbiFlow/prjuray/blob/master/fuzzers/002-tilegrid/add_tdb.py#L124).

PR #1477, opened 2020-10-29
Add example graph folding prototype.

litghost:

PR #1605, opened 2021-03-03
WIP: Max shared graph folding

ethanroj23:
This pull request adds upon [add_example_graph_folding](https://github.com/SymbiFlow/prjxray/pull/1477) by implementing another method of creating subgraphs of tiles and wirePatterns. 

In order to create subgraphs, the max_shared method chooses sets of tile patterns that contain the max shared amount of wirePatterns. This heuristic is implemented by the `generate_max_shared_subgraphs()` function in `reduce_graph_for_type.py`

Because the algorithm creates subgraphs with no duplicate wirePatterns, every subgraph must be included in the final set of solutions. As such, there is no need to choose which subgraphs should be used for the final solution. 

The tables below give a comparison between the Greedy set cover method, and the new Max shared method. So far, file size is the main optimization, but the next improvement will look into the speed of access.

| Using Greedy set Cover     | A50T        |A100T        |A200T        |
| ---------------------------| ----------- |-------------|-------------|
| Wire to node map           | ~3.4M       | ~3.75M      | ~4.7M       |
| Node to wire map           | ~6.0M       | ~6.55M      | ~7.75M      |
| Node to downstream pips map| ~1.1M       | ~1.1M       | ~1.4        |

| Using Max Shared Algorithm | A50T        |A100T        |A200T        |
| ---------------------------| ----------- |-------------|-------------|
| Nodes and Wires map        | ~6.14M      | ~6.58M      | ~7.96M      |

In addition to the files' smaller size, the time taken to run the new method is much shorter than the previous method.


Makefile
```diff
@@ -18,7 +18,8 @@ endif
 # Tools + Environment
 IN_ENV = if [ -e env/bin/activate ]; then . env/bin/activate; fi; source utils/environment.python.sh;
 env:
-	python3 -mvenv env
+	python3.8 -mvenv env --without-pip
+	$(IN_ENV) curl https://bootstrap.pypa.io/get-pip.py | python
```
litghost, line 22: Why this change?  This seems wrong? Shouldn't it be:
```
python3 -mvenv env
python3 -mpip install --upgrade pip
```
?

prjxray/node_lookup.py
```diff
@@ -120,3 +129,4 @@ def wires_for_tile(self, tile):
 """, (tile, ))
         for row in c:
             yield row[0][len(tile) + 1:]
+THIS IS A MODIFICATION
```
litghost, line 132: What is going on here?

ethanroj23: I don't remember that...It must have been a comment or something.

Makefile
ethanroj23: Perhaps your mentioned fix would work better, but this was just a workaround that we implemented to get the "make env" command to run properly. We were running into issues with the python version and module version incompatibilities in the requirements.txt file

litghost:
@ethanroj23 Can you fix you DCO? E.g. use `git commit -s" on your commits.

Makefile
mithro: Please use @litghost's method.

ethanroj23: I remember now that we were having issues with the Python version because our machine is running Ubuntu 16.04 which has a python3 version of 3.5. I had to use "python3.8" in order for it to work. I may look into this further in order to see if the method mentioned by @litghost would work for me.

litghost:
It's not clear from you data, but do your method allow fast lookup of node to wire and wire to node?  From glancing at your data, it doesn't appear so?

ethanroj23:
> It's not clear from you data, but do your method allow fast lookup of node to wire and wire to node? From glancing at your data, it doesn't appear so?

@litghost I don't think it does at this point. That's what I wanted to look into next. You mentioned that yours had a linear scan with an N=27? Where does that occur?

litghost:
> @litghost I don't think it does at this point.

I think that means that you data size comparison is incorrect.  I believe you have only implemented `Node to wire map` from the original size table, and you can see that your method isn't actually smaller.  Now if you method is very fast, and only a few percentage points larger, and provides an exact cover, those are reasons to choose it over the previous implementation.  But if you method only does node -> wire, it is not complete yet.

> You mentioned that yours had a linear scan with an N=27? Where does that occur?

I could not find an efficient way to encode wire index to particular subgraph.  `N=27` is `max number of subgraphs required to cover a tile`.  So even once you have the set of subgraphs that cover a tile, there is still a need for a linear scan through all subgraphs.  `N=27` is small, but having a low constant factor constant method is better than a linear scan.  There is clearly a memory / scan method tradeoff here, it is unclear if there is a better algo.

Here is the relevant scan in the lookup code: https://github.com/SymbiFlow/prjxray/pull/1605/files#diff-4b2a605b9c1d1569a9b7f07bf6197dd04ef47f6f30d29435e2a575d32a67a654R88-R98

litghost:
> @ethanroj23 Can you fix you DCO? E.g. use `git commit -s" on your commits.

DCO is still incorrect, please fix: https://stackoverflow.com/questions/25570947/how-to-use-git-interactive-rebase-for-signing-off-a-series-of-commits

ethanroj23:
> > @ethanroj23 Can you fix you DCO? E.g. use `git commit -s" on your commits.
> 
> DCO is still incorrect, please fix: https://stackoverflow.com/questions/25570947/how-to-use-git-interactive-rebase-for-signing-off-a-series-of-commits

Okay, I have fixed the DCO.

ethanroj23:
> I think that means that you data size comparison is incorrect. I believe you have only implemented `Node to wire map` from the original size table, and you can see that your method isn't actually smaller. Now if you method is very fast, and only a few percentage points larger, and provides an exact cover, those are reasons to choose it over the previous implementation. But if your method only does node -> wire, it is not complete yet. Both node -> wire and wire -> node are required.

The file implements both node->wire and wire->node. In my original testing, I did just the node->pip_wires, and my representation for that was 0.5MB vs the 1.1MB representation using greedy set cover. After that, I implemented a data structure that is bidirectional (node->wire & wire->node)

> > You mentioned that yours had a linear scan with an N=27? Where does that occur?
> 
> I could not find an efficient way to encode wire index to particular subgraph. `N=27` is `max number of subgraphs required to cover a tile`. So even once you have the set of subgraphs that cover a tile, there is still a need for a linear scan through all subgraphs. `N=27` is small, but having a low constant factor constant method is better than a linear scan. There is clearly a memory / scan method tradeoff here, it is unclear if there is a better algo.

I will take a look now into the linear scan to see if I can improve upon that.

minitests/graph_folding/Makefile
```diff
@@ -14,7 +14,6 @@ ${DATABASE}: build_node_lookup.py node_lookup.py | build_${PART}
 		--part ${PART} \
 		${DATABASE}
 
-<<<<<<< HEAD
```
Xiretza: This should be squashed into the commit that introduced it, leaving broken commits like this causes bisecting to be even more of a pain than it already is.

ethanroj23: Thanks, for pointing that out. It should be fixed now.

Xiretza: Maybe you forgot to push? The broken commits are still there.

PR #1606, opened 2021-03-04
005-tilegrid: Extract frames count information from part's json

tmichalak:
This PR adds automatic frames count extraction from the `part.json` of a given part to the `005-tilegrid` fuzzer.
Porting this solution form U-Ray was mentioned in PR #1603
When the specified frames count value exceeds the maximum allowed value a warning is displayed and the value falls back to the maximum allowed.


litghost (commented):
Waiting for green CI

PR #1614, opened 2021-03-08
gtp: generate attributes and ports files to add to the db

acomodi:

This PR pushes to the DB some relevant information for [fasm2bels](https://github.com/SymbiFlow/symbiflow-xc-fasm2bels/pull/49) about the GTP primitives:

- attributes
- ports

These are added in the `prjxray-db/<family>` directory


litghost:
> These are added in the `prjxray-db/<family>` directory

Please move these files into a sub-directory.  I believe something like `prjxray-db/<family>/cell_data` or equiv might be good?

litghost:
@acomodi Can you explain the "values" field.  I don't understand it.  Maybe add a section in `prjxray/docs`?

acomodi:
> Can you explain the "encoding" vs "values" field.

Sure, basically, `value` is what is a possible value to assign to the parameter at the cell instantiation, while `encoding` is the resulting encoding in the bitstream of those values for that parameter.
E.g.: if the PLL1_FBDIV param gets assigned to 1, the resulting FASM will need tobe `PLL1_FBDIV[5:0] = 6'b010000` as reported in

litghost:
> > Can you explain the "encoding" vs "values" field.
> 
> Sure, basically, `value` is what is a possible value to assign to the parameter at the cell instantiation, while `encoding` is the resulting encoding in the bitstream of those values for that parameter.
> E.g.: if the PLL1_FBDIV param gets assigned to 1, the resulting FASM will need tobe `PLL1_FBDIV[5:0] = 6'b010000` as reported in

Please add that into `prjxray/docs` or equiv.

docs/dev_database/common/cell_data.rst
```diff
@@ -0,0 +1,67 @@
...
+
+Attributes files
+~~~~~~~~~~~~~~~~
+
+This is a JSON file containing a dictionary of parameters, each one with, at most, four attributes:
+
+    - Type: one of BIN, INT, STR, BOOL.
+    - Values: all possible values that this parameter can assume. In case of `BIN` types, the values list contains only the maximum value reachable.
+    - Digits: number of digits (or bits) required to enable a parameter.
+    - Encoding: This is present only for `INT` types of parameters. These reflect the actual encoding of the parameter value in the bit array.
```
litghost, line 35: I think a couple examples are going to help here.

acomodi: Done

docs/dev_database/common/cell_data.rst
```diff
@@ -0,0 +1,82 @@
...
+
+As an example of parameter please, refer to the following::
```
litghost, line 37: So the examples doesn't actually illuminate what is going on. I'm going to merge this PR, but please open a new PR expanding on these docs.

litghost: Specifically walk through some examples relating a cell parameter, the data in the JSON and the output FASM.

acomodi: Aha, all right, I misunderstood. I'll update it right away

PR #1615, opened 2021-03-08
Add some disk space checks in happy and sad paths.

litghost:


litghost:
FYI @dnltz

litghost:
@mithro / @tmichalak Let me know if you have another comments about this change, thanks.

tmichalak:
> Let me know if you have another comments about this change, thanks.

Looks fine for me

mithro:
@litghost Can we get the output of `dmesg` if the `df -h` doesn't have the `/tmpfs` mount?

dnltz:
@mithro - I can do this tomorrow, if @litghost agrees. My idea was to also abort the run if /tmpfs is not mounted to get faster feedback and to restart the run directly.

litghost:
> @mithro - I can do this tomorrow, if @litghost agrees. My idea was to also abort the run if /tmpfs is not mounted to get faster feedback and to restart the run directly.

Sounds good to me, thanks!

PR #1616, opened 2021-03-09
Rework pcie-conf fuzzer

kowalewskijan:
There are several changes introduced in this PR:
* Unification of the parameters used in the fuzzer. Parameters were translated to the JSON format (the same as in GTP fuzzers)
* Added generate_port script used to export JSON file with ports definitions of the PCIE_2_1 site
* Added exporting ports and attributes JSON files to prxjray-db

Depends on `make_ports.py` tool introduced in: https://github.com/SymbiFlow/prjxray/pull/1614

### DIFF
--- a/fuzzers/061-pcie-conf/Makefile
+++ b/fuzzers/061-pcie-conf/Makefile
@@ -5,23 +5,67 @@
 # https://opensource.org/licenses/ISC
 #
 # SPDX-License-Identifier: ISC
+
+SHELL = bash
+
 N ?= 40
 
-include ../fuzzer.mk
+BUILD_DIR = build_${XRAY_PART}
 
-database: build/segbits_pcie_bot.db
+SPECIMENS := $(addprefix ${BUILD_DIR}/specimen_,$(shell seq -f '%03.0f' $(N)))
+SPECIMENS_OK := $(addsuffix /OK,$(SPECIMENS))
+FUZDIR ?= ${PWD}
+CELLS_DATA_DIR = ${XRAY_FAMILY_DIR}/cells_data
 
-build/segbits_pcie_bot.rdb: $(SPECIMENS_OK)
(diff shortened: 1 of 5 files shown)


acomodi (changes requested):
a small fix is required and then LGTM

litghost (changes requested):
I feel like params.py should be able to generate attrs.json.  @acomodi / @kowalewskijan Thoughts?

kowalewskijan:
I think we should keep the `params.py` and introduce a sort of standardization. Since GTPs and PCIe are the very first cases we'll store the attributes file (later on added to the prjxray-db), I think we should introduce a kind of standard for it. Probably we could implement something like `CellAttribute` abstract class and then have some classes for BIN, STR, INT etc. attribute types. In the result we'll have a much clearer handling of attributes rather than just having a list of tuples. What do you think @litghost / @acomodi ?

acomodi:
Makes sense to standardise the attribute types through some classes, but, I am thinking though where would this script be used?

At the moment, we would need it only for this specific case in the PCIE_2_1 primitive, where we have a params.py with some `(<attribute>, <width>)` which would need to be standardised to attrs.json.

I think for the purpose of this PR it makes sense to use params.py to generate the attrs.json, and, as a follow-up, we could enhance both PCIe and GTP fuzzers with the addition of a more standardization. For instance, we could add checks that the pre-computed attrs.json are compatible with the standard (e.g. INT types must have an encoding field, etc.).

kowalewskijan:
I added a functionality to emit the JSON file from `params.py` ([link](https://github.com/SymbiFlow/prjxray/pull/1616/files#diff-f87edb4d9d6b79c5f6270bbc2cf6111e0446491665f6746aeac8731175dd6eddR309-R338))

litghost (approved):
LGTM, merge if CI is green.

acomodi (approved):
LGTM, merging

PR #1617, opened 2021-03-10
docs: cells_data: enhance docs

acomodi:


docs/dev_database/common/cell_data.rst
```diff
@@ -60,6 +60,16 @@ As an example of parameter please, refer to the following::
         },
     }
 
+Moreover, let the attribute be ``PLL0_REFCLK_DIV``. The cell instantation of the ``GTPE2_COMMON`` primitive is as follows::
```
litghost, line 63: Can you explain `encoding` here?  Values/type/digits make some sense.

litghost: If encoding is a map, why not just make it a map?

docs/dev_database/common/cell_data.rst
```diff
@@ -30,9 +30,9 @@ Attributes files
 This is a JSON file containing a dictionary of parameters, each one with, at most, four attributes:
 
     - Type: one of BIN, INT, STR, BOOL.
-    - Values: all possible values that this parameter can assume. In case of `BIN` types, the values list contains only the maximum value reachable.
+    - Values: all possible values that this parameter can assume. In case of ``BIN`` types, the values list contains only the maximum value reachable.
     - Digits: number of digits (or bits) required to enable a parameter.
-    - Encoding: This is present only for `INT` types of parameters. These reflect the actual encoding of the parameter value in the bit array.
+    - Encoding: This is present only for ``INT`` types of parameters. It reflects the actual encoding of the parameter value in the bit array.
```
litghost, line 35: What is the behavior if the `encoding` field is not present?

acomodi: A "key not found" error will be present, there is no handling of this situation actually. We might add a check prior to adding the file to the cells_data location in the db.

litghost: I think the field names are not right.  I believe ``INT`` should be ``INT_ENUM``, or ``INT_MAPPING`` ?

PR #1618, opened 2021-03-11
Add kokoro hostcheck

dnltz:
Validated with a check if `/tmpfs2` is mounted:

```
========================================
Check storage
----------------------------------------
++ mount
++ grep /tmpfs2
++ MOUNT_RET=1
++ set +x -e
----------------------------------------
Error: No storage is mounted on /tmpfs.
----------------------------------------
========================================
Dmesg dump
----------------------------------------
[    0.000000] Initializing cgroup subsys cpuset
[    0.000000] Initializing cgroup subsys cpu
[    0.000000] Initializing cgroup subsys cpuacct
```

### DIFF
--- a/.github/kokoro/db-full.sh
+++ b/.github/kokoro/db-full.sh
@@ -11,6 +11,7 @@ set -e
 
 cd github/$KOKORO_DIR/
 
+source ./.github/kokoro/steps/hostcheck.sh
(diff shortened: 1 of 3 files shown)


mithro (commented):
Awesome, thanks for all your hard work @dnltz!

PR #1619, opened 2021-03-12
Defer exiting for db-check error until after cleanup.

litghost:


mithro (approved):
LGTM, merge if CI says it's good.

Side note -- this script is /probably/ starting to get the "rewrite in Python" level of complexity....

litghost:
> Side note -- this script is /probably/ starting to get the "rewrite in Python" level of complexity....

Was thinking about that too ...

PR #1634, opened 2021-03-25
DNM: 071-ppips: fix allowed and forbidden wires for PCIe

kowalewskijan:
The PR is marked as `DNM` until we can confirm that the results from the CI are correct by testing them in fasm2bels.

Signed-off-by: Jan Kowalewski <jkowalewski@antmicro.com>


kowalewskijan:
I confirmed in the fasm2bels that current ppips files for PCIe are correct now. When the CI turns green the PR is ready to merge.

acomodi (approved):
LGTM

Issue #1343, opened 2020-05-25
Assign xc7 prefix to prjxray tools to avoid conflicts


acomodi:
All the tools in prjxray should have an `xc7` prefix, so to make them unique.

E.g. `bitread` or `bittool` could enter in conflicts with other tools names.

slayer321:
Hey, I new to open source and want to contribute ........ can you guide me how can I start with this issue?

slayer321:
ok........ so you mean all the tools in [here](https://github.com/SymbiFlow/prjxray/tree/master/tools)  name must prefix with `xc7`   that's what's need to done?

slayer321:
I have send a PR #1668 .. do check and let me know if everything is correct

PR #1668, opened 2021-05-13
Fix issue #1343

slayer321:
do check and let me know if everything is correct


acomodi:
Hi @slayer321,
what would actually be needed is to change the name of the resulting binaries. The assignment of the binary names happens in https://github.com/SymbiFlow/prjxray/blob/master/tools/CMakeLists.txt.

Take the following as an example: https://github.com/SymbiFlow/prjxray/blob/80726cb73ba5c156549d98a2055f1ee3eff94530/tools/CMakeLists.txt#L7.

To change bittool to be xc7bittool you should modfiy that line to be:
```
add_executable(xc7bittool bittool.cc)
target_link_libraries(xc7bittool libprjxray gflags absl::strings absl::span)
```

Moreover, these tools are used in different parts in prjxray, so the change in the tools names need to be addressed also where the tools are used.

slayer321:
ok, I understood that I just need to change the name [here](https://github.com/SymbiFlow/prjxray/blob/master/tools/CMakeLists.txt.) in this file . and what about the changes that I did . that was not needed right?

Issue #1658, opened 2021-04-29
parts.yaml file not found when running settings script for board family


benglines:
When trying to install Project X-Ray for the first time, I get the following error when running `source settings/artix7.sh`

```
Traceback (most recent call last):
  File "/home/benglines/prjxray_test/utils/create_environment.py", line 66, in <module>
    main()
  File "/home/benglines/prjxray_test/utils/create_environment.py", line 59, in main
    environment = get_environment_variables()
  File "/home/benglines/prjxray_test/utils/create_environment.py", line 37, in get_environment_variables
    part_info = get_part_information(db_root, part)
  File "/home/benglines/prjxray_test/prjxray/util.py", line 44, in get_part_information
    "Mapping file {} does not exists".format(filename)
AssertionError: Mapping file /home/benglines/prjxray_test/database/artix7/mapping/parts.yaml does not exists
```
I have also seen it complain about not finding a resources.yaml file (just like #1639), but I am not exactly sure why that happened nor how to recreate the error. Either way, I solved the issue in both cases by running `make db-prepare-artix7`, and then `source settings/artix7.sh`. The quick start guide suggest you do them in the opposite order. 

This issue could be resolved by simply switching Step 6 and Step 7 in the quick start guide, so that the shell scripts run in the correct order i.e. running `source/settings/artix7.sh` _after_ running `make db-prepare-parts` or `make db-prepare-artix7`.

mithro:
Seems like we should set up the CI to use [`tuttest`](https://github.com/antmicro/tuttest) to run the commands in the README [like we do for the symbiflow-examples repository](https://github.com/SymbiFlow/symbiflow-examples/runs/2460999808?check_suite_focus=true#step:8:7).

dnltz:
@benglines - Thanks for reporting. Indeed, the script should be sourced after preparing the db.
@mithro - If I have a little bit time soon, I can try to add such a simple test.

PR #1678, opened 2021-05-25
update README

acomodi:
Signed-off-by: Alessandro Comodi <acomodi@antmicro.com>

fixes https://github.com/SymbiFlow/prjxray/issues/1674

PR #1682, opened 2021-06-03
Rearranged quickstart steps to correctly set device environment

### LINKED ISSUE #1674: Documentation Issue
The step 6 in the Quick start guide should be done after completing the step 7 and 8. Otherwise it throws an error.
Also the documentation does not specifically provide the directory the command should be ran in for ex: after the step 9 we will have to move to a different directory for step 10 and then follow the C++ commands step, so please take care of these issue,

benglines:
This should resolve #1674, and also addresses #1658. All I did was move step 6 of the quick start guide (setting the environment for a device) after preparing/downloading the databases (previously Steps 7 and 8)

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,12 +83,6 @@ This might give you an error about `sphinx_autodoc_typehints` but it should
 correctly build and install pyjson5. After this, run either option 1 or 2 again.
 
 ### Step 6: ###
-Always make sure to set the environment for the device you are working on before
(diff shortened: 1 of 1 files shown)


acomodi:
Hi @benglines, thanks for this PR.
There is now a conflict in the README after merging https://github.com/SymbiFlow/prjxray/pull/1678, and this would require a rebase.

PR #1692, opened 2021-06-15
034-cmt: fix conflicting PIPs

acomodi:
(edit) Fixes bits which were not filtered out for the `CMT_TOP_R` case.


mkurc-ant (approved):
LGTM

PR #1811, opened 2021-12-24
Small fix to fuzzer makefile.

mithro:
* Make clear the device specific fuzzers.
 * Make clear the DSPs being disabled for Kintex7

Signed-off-by: Tim 'mithro' Ansell <tansell@google.com>


mithro:
As far as I can see, this doesn't cause any change in output, so I'm going to merge it.

PR #1830, opened 2022-01-24
Also fuzz OCLK and OCLKB inputs to the ISERDESE2. Fix typo.

bl0x:
I have not run the fuzzer, but am mainly asking if this is a sane change, since this is the first time I'm looking into this code.

I have hopes that this may fix a problem I've encountered here:

https://github.com/SymbiFlow/symbiflow-arch-defs/pull/2367#issuecomment-1019623476

while adding support for the OCLK and OCLKB inputs reqiured for the OVERSAMPLE mode of ISERDESE2 in symbiflow-arch-defs.

### DIFF
--- a/fuzzers/035b-iob-iserdes/top.py
+++ b/fuzzers/035b-iob-iserdes/top.py
@@ -172,6 +172,10 @@ module top (
   input  wire clk1,
   (* CLOCK_BUFFER_TYPE = "NONE" *)
   input  wire clk2,
+  (* CLOCK_BUFFER_TYPE = "NONE" *)
+  input  wire clk3,
+  (* CLOCK_BUFFER_TYPE = "NONE" *)
+  input  wire clk4,
   input  wire ce,
   input  wire rst,
   input  wire [{N}:0] di,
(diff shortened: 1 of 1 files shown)


acomodi:
Hi @bl0x, thanks for this PR. When dealing with the ISERDES fuzzer we actually noticed that setting the OCLK and OCLKB inputs, the PIPs responsible for those were not actually producing any unknown bit.

I think that the best way of verifying that we already have all the information is using the minitest approach:

1) Isolate the ISERDES from your design with the exact features you want to use
2) Generate a bitstream with Vivado
3) Use bit2fasm with the `--verbose` option enabled
4) Check the number of unknown bits (if any) in the resulting FASM file

bl0x:
How do I notice unknown bits in the resulting FASM file?
bit2fasm prints everything to stdout by default.

In order to properly generate a bitstream with Vivado I had to make a few changes to the code, namely adding proper clock sources and making connections to the iserdes.
I also needed to set IOSTANDARD attributes to the I/O pins used (though I doubt that that is relevant here).

With these changes the bit2fasm output did not show any obvious unknown bits.
Also, I could generate a bitstream from the modified (fixed) code using the Symbiflow toolchain.

It looks to me, as if this PR is then not needed.

acomodi:
In that case all the bits should actually be there

> With these changes the bit2fasm output did not show any obvious unknown bits.

Just to double-check, did you run with the `--verbose` flag? If yes and no unknown bits are reported, than I think that this is another confirmation that all bits are documented.

bl0x:
Yes, I did run with the --verbose flag. How would unknown bits be reported?
Like so?
```
Found unknown bits: XX, YY
```

PR #1835, opened 2022-01-27
Fix wrong parameter in get_part_resources()

MidsummerNight:
The assertion error for `get_part_resources()` should print `res_mapping` instead of `part_mapping` (the latter isn't even defined in `get_part_resources()`). See #1834 for details.

PR #1836, opened 2022-01-28
Fix wrong parameter in get_part_resources()

MidsummerNight:
The assertion error for `get_part_resources()` should print `res_mapping` instead of `part_mapping` (which is not defined in this function).

This PR supersedes #1835 since the DCO check of that one was messed up by me again.

Signed-off-by: Steve <steve.bohan.liu@outlook.com>


MidsummerNight:
@acomodi Here's the correct PR

acomodi (approved):
Looks good to me, I think this PR just needs a rebase and we can merge it.

PR #1862, opened 2022-02-23
docs: update theme; add logo and favicon

umarcor:
[f4pga--1862.org.readthedocs.build/projects/prjxray/en/1862](https://f4pga--1862.org.readthedocs.build/projects/prjxray/en/1862/)


.github/workflows/Doc.yml
```diff
@@ -0,0 +1,48 @@
+name: Docs
```
kgugala, line 1: why do we build the docs here (and upload them as GH-pages)? The docs here are handled as subproject of the main F4PGA docs, so any change here should trigger build of the F4PGA docs

kgugala (changes requested):

.github/workflows/Doc.yml
umarcor: When a contributor forks this repo and works on feature branches (without creating a PR) the RTD hooks are not triggered. In those cases, the contributor needs to register another RTD project for testing purposes, or a PR needs to be opened. Conversely, this workflow allows them to have quick feedback. Although pushed to `gh-pages`, the site is not served unless enabled explicitly. I.e. http://f4pga.github.io/prxray is empty.

kgugala: I'd assume when working on the docs people can build the docs locally. RTD will generate a preview doc in PR

umarcor: Ok. I removed that commit.

kgugala (approved):
LGTM

PR #1871, opened 2022-03-05
Artix 75T is same fabric as 100T, 15T is same fabric as 50T

marzoul:
Add support for part xc7a75t and xc7a15t
These are actually same fabrics as, respectively, xc7a100t and xc7a50t.

Note: There are 3 ways to confirm this.
- Xilinx documentation ds181 for Artix-7 DC characteristics
- Xilinx documentation ug470 for bitstream lengths
- Manual creation of simple IO planning Vivado projects and observing that chip floorplans are identical.

I confirm the 3 points above for the 2 parts 75t and 15t.

### DIFF
--- a/.github/workflows/Pipeline.yml
+++ b/.github/workflows/Pipeline.yml
@@ -79,7 +79,7 @@ jobs:
         env:
           XRAY_SETTINGS: ${{ matrix.family }}
 
-      - uses: actions/upload-artifact@v3
+      - uses: actions/upload-artifact@v4
         if: ${{ always() }}
         with:
           name: ${{ matrix.family }}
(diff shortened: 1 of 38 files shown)


hansfbaier:
Thanks, that makes sense. Will try.

marzoul:
Hi,
I'm completely new to the automerge system and github actions.
I see that the item "docs/readthedocs.org:prjxray" failed but I can't understand why, could anyone advise ?
It says
```
Error
Config validation error in build.os. Value build not found.
```
It seems this error happens just with command `cat .readthedocs.yml` without any for of build being done previously.
Is this test item working ?

PR #1889, opened 2022-03-31
Create symbolic links in db for devices with identical fabric

tmichalak:
This PR makes sure that the generated db contains symbolic links to fabric specific files for devices with an identical fabric, such as `xc7a15t`, `xc7a35t` and `xc7a50t` sharing the same `xc7a50t` fabric.
This is needed to avoid problems like the one reported [here](https://github.com/gatecat/nextpnr-xilinx/issues/35).


Xiretza:
Any update on this?

Issue #521, opened 2019-01-15
maketodo verbose should print on stderr


kgugala:
`maketodo` script results are normally printed to stdout and redirected to a todo file in Makefile. Using the script with `--verbose` corrupts the todo file. Printing verbose msg on stderr would solve this.

PR #1909, opened 2022-04-19
Redirect int_maketodo's verbose output to stderr

colemole9595:
`int_maketodo.py`'s verbose messages are now sent to stderr. This pull request addresses #521.

I ran `make format-py`, is anything else needed?

Signed-off-by: Anmol Shrivastava <colemole9595@gmail.com>


fuzzers/int_maketodo.py
```diff
@@ -126,22 +127,26 @@ def maketodo(
...
-        verbose and print('%s: %u entries, done!' % (pipfile, len(todos)))
+        verbose and print(
+            '%s: %u entries, done!' % (pipfile, len(todos)), file=sys.stderr)
         return
 
-    verbose and print("pipfile todo sample: %s" % list(todos)[0])
+    verbose and print(
+        "pipfile todo sample: %s" % list(todos)[0], file=sys.stderr)
 
     if 0 and verbose:
```
acomodi, line 140:
```suggestion
    if verbose:
```

Currently this condition is always false and quite useless, probably it would be better to remove the zero.

acomodi (changes requested):
Hi @colemole9595, thanks for this PR. One small comment and I think we can run CI and merge if all is ok.

fuzzers/int_maketodo.py
colemole9595: Thank you for the review. I have pushed the requested changes. Should I squash?

PR #1926, opened 2022-05-12
Revert "build(deps): bump third_party/abseil-cpp from `bcefbdc` to `4…

kgugala:
…d751aa` (#1916)"

This seems to break conda builds


umarcor (commented):
We might need to ignore it explicitly (https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#ignore). Otherwise, the automerge will pick it again.

PR #1937, opened 2022-05-22
ci: git config safe.directory

umarcor:


umarcor:
I'm marking this as a draft because it might be fixed on the runners. I'll close it when the fix is deployed.

Issue #1873, opened 2022-03-09
Remove Kokoro CI


acomodi:
Now that the Custom GH runners have been deployed, the Kokoro CI can be disabled.

umarcor:
@mithro I guess this is done already?

mithro:
Just created the change which will remove the configuration on the Google side.

PR #1944, opened 2022-05-24
remove kokoro related scripts and assets

umarcor:
As discussed in #1873.


mithro:
@umarcor - Snap - it looks like i created #1945 at the same time.....

PR #2060, opened 2022-11-27
Experimental bitstream compression support

LAK132:
This PR adds an option to compress bitstreams. Tested with an XC7A200T(FBG484-2) (MEGA65 R3). 

This is needed for using open tooling with the MEGA65 as the uncompressed bitstreams are too large to fit in the core slots.

With the cores I tested this with, the compression step didn't seem to have any noticeable impact on build times.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -3,8 +3,6 @@ cmake_minimum_required(VERSION 3.5.0)
 project(prjxray)
 option(PRJXRAY_BUILD_TESTING "" OFF)
 
-set(CMAKE_CXX_STANDARD 14)
-
 # Add sanitizers-cmake package
 set(CMAKE_MODULE_PATH "${CMAKE_SOURCE_DIR}/third_party/sanitizers-cmake/cmake" ${CMAKE_MODULE_PATH})
 find_package(Sanitizers)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -3,6 +3,8 @@ cmake_minimum_required(VERSION 3.5.0)
 project(prjxray)
 option(PRJXRAY_BUILD_TESTING "" OFF)
 
+set(CMAKE_CXX_STANDARD 14)
+
 # Add sanitizers-cmake package
 set(CMAKE_MODULE_PATH "${CMAKE_SOURCE_DIR}/third_party/sanitizers-cmake/cmake" ${CMAKE_MODULE_PATH})
 find_package(Sanitizers)
--- a/.github/workflows/Pipeline.yml
+++ b/.github/workflows/Pipeline.yml
@@ -79,7 +79,7 @@ jobs:
         env:
           XRAY_SETTINGS: ${{ matrix.family }}
 
-      - uses: actions/upload-artifact@v3
+      - uses: actions/upload-artifact@v4
         if: ${{ always() }}
         with:
           name: ${{ matrix.family }}
@@ -124,7 +124,7 @@ jobs:
       - name: Run Test
         run: make test --output-sync=target --warn-undefined-variables
 
-      - uses: actions/upload-artifact@v3
+      - uses: actions/upload-artifact@v4
         if: ${{ always() }}
         with:
           path: |
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -5,6 +5,11 @@
 # Required
 version: 2
 
+build:
+  os: ubuntu-22.04
+  tools:
+    python: "3.7"
+
 # Build documentation in the docs/ directory with Sphinx
 sphinx:
   configuration: docs/conf.py
@@ -14,7 +19,6 @@ formats: all
 
 # Optionally set the version of Python and requirements required to build your docs
 python:
-  version: "3.7"
   install:
     - requirements: docs/requirements.txt
 
--- a/Makefile
+++ b/Makefile
@@ -9,6 +9,7 @@ SHELL = bash
 ALL_EXCLUDE = third_party .git env build docs/env
 
 INSTALL_DIR ?=
+ENV_DIR ?= env
 
 # Skip this check if the ALLOW_ROOT var is defined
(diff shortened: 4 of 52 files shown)


CMakeLists.txt
```diff
@@ -36,8 +38,6 @@ target_include_directories(yaml-cpp PUBLIC
 	$<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/third_party/yaml-cpp/include>
 	)
 
-# Set the CXX standard and compile time for our code only.
-set(CMAKE_CXX_STANDARD 14)
```
LAK132, line 40: It refused to build on my system with the standard being set here

mithro:
@kgugala - Could you take a look?

kgugala:
@LAK132 thanks for the PR, It seems the CI is red. There is a lot compilation errors like:

```
03:00:22 | /root/prjxray/prjxray/lib/xilinx/tests/xc7series/configuration_test.cc:281:24: error: 'struct prjxray::xilinx::Configuration<prjxray::xilinx::Series7>::PacketData' has no member named 'size'
03:00:22 |   281 |  EXPECT_EQ(packet_data.size(), 15 * 101);
```

You mentioned that you had to change C++ standard. @LAK132 which standard did you use in your local build?

LAK132:
I just moved the standard setting thing to the top of the script, it's still C++14. one of the dependencies didn't like that the standard wasn't being set before it was compiled, so I moved it. 

I'll take a look at those errors, must be coming from something I didn't try to build

kgugala:
looks like all the databases (artix, zynq, kintex and spartan) builds failed with:

```
ValueError: invalid literal for int() with base 16: '<0'
```

in the `005-tilegrid/clb` fuzzer. Here is the whole trace:

```
File "add_tdb.py", line 162, in <module>
     main()
   File "add_tdb.py", line 158, in main
     run(args.fn_in, args.fn_out, verbose=args.verbose)
   File "add_tdb.py", line 138, in run
     for (tile, frame, wordidx) in load_db(tdb_fn):
   File "add_tdb.py", line 51, in load_db
     check_frames(tagstr, addrlist)
  File "add_tdb.py", line 21, in check_frames
     frame = parse_addr(addrstr, get_base_frame=True)
   File "add_tdb.py", line 30, in parse_addr
     frame = int(line[0], 16)
 ValueError: invalid literal for int() with base 16: '<0'
```

Since all the DB builds fail with the same error, I suspect the changes in the tools trigger this. @LAK132 can you check this?

mithro:
@LAK132 - If you are having issues getting the tool / testing environment setup, please log github issues and we will try and improve both the instructions and the process.

hansfbaier:
image
segbits_tilegrid.tdb for xc7a100t has no candidates entries

hansfbaier:
I narrowed down the problem to the `bitread` tool: When I exchange it with a known good version then the candidates are found.

hansfbaier:
@LAK132 This change in configuration.h looks suspicious to me, it possibly has different semantics than the original code:
image

hansfbaier:
@LAK132 Here are two example bitstreams which show the difference:
[example-data.tar.gz](https://github.com/f4pga/prjxray/files/10495058/example-data.tar.gz)
Consider the .bits files there (the output of bitread).
the -bad version (with the bitread tool from this branch) seems to have extraneous data:
image

LAK132:
> it possibly has different semantics than the original code

Technically true. Previously if that address had already been written to it would overwrite it, whereas with `.insert` it will not overwrite existing frame writes. That said, writing to the same frame multiple times feels like a bug? The beauty of `.insert` is that it won't default construct a frame before the assignment can take place. That said, that change should have been the only major change to the uncompressed codepath, so if reverting that fixes the problem...

LAK132:
Also it looks like with that example data, it's the -good output that has the extra data, not the -bad output? Were these outputs made with or without compression enabled?

hansfbaier:
@LAK132 Sorry, I was wrong, the original bitread has the extra bits (green parts above), the bitread of this branch does not have them.

hansfbaier:
@LAK132 Here is a 'good' version of bitread for comparison
[bitread.gz](https://github.com/f4pga/prjxray/files/10495654/bitread.gz)

hansfbaier:
@LAK132 So the way to go would be:
1. instrument the code with printfs to find out what happenes with the missing parts, or use a debugger
2. run `./build/tools/bitread -part_file database/artix7/xc7a100tfgg676-1/part.yaml -F 0x00000000:0xffffffff -o design.bits -y -z design.bit` (take design.bit from the example data, either the good or the bad version, they are equivalent)

LAK132:
`.insert_or_assign` would have been a better way to do this but unfortunately that's a C++17 feature

hansfbaier:
@LAK132 Did that fix the bitread results?

LAK132:
It should have, the uncompressed codepath should be practically untouched now

hansfbaier:
Can you test it? End of the day for me here...

LAK132:
The bitread output appears to be the same as the known good version with my test core project. I still haven't managed to run any of the other tests myself.

hansfbaier:
@LAK132 OK great let's wait for the CI then

hansfbaier:
@LAK132 Good news, the changes work well here on my machine. I added a temporary PR (see above) to run the full CI on this build. @kgugala @mithro Can we run the CI on this?

hansfbaier:
@LAK132 CI results look great! (Only the gtp-common-pips fuzzer is red, but that one is notoriously flickering) @mkurc-ant @kgugala Ready for review!

hansfbaier:
@tmichalak Would it be possible to run the CI on this branch?

LAK132:
Is there a reason the tests were cancelled? Did they just time out?

LAK132:
All green now, anything else I need to do?

LAK132:
Any update on this?

PR #2068, opened 2022-12-07
WIP 071-ppips rewrite attempt using RapidWright

hansfbaier:
This is a proof of concept rewrite of the 071-ppips fuzzer using exclusively the ppip data from RapidWright.
Note that the output is considerably smaller than the previous fuzzer,
obviously missing important pieces like below:
image


hansfbaier:
The tests of course will all fail, because rapidwright is not available in the CI.
Thus, here is a sample of the results for artix7:
[rapidwright-ppips-artix7.tar.gz](https://github.com/f4pga/prjxray/files/10172274/rapidwright-ppips-artix7.tar.gz)

mithro:
Until this data is available through Xilinx in the FPGA Interchange format (https://antmicro.com/blog/2021/09/fpga-interchange-format/), we need to continue to maintain a version of fuzzer which demonstrates the ability to get this data from Vivado (even if a faster version is available through using RapidWright).

BTW You might want to checkout the [`prjuray` fuzzers](https://github.com/f4pga/prjuray/tree/master/fuzzers) -- They were written after learning a lot from `prjxray` and probably a lot cleaner.

mithro:
@hansfbaier - We **can** have a "fast" version that uses RapidWright **but** we will still need to continue to maintain an equivalent version that uses only Vivadio.

hansfbaier:
@mithro Not necessary. This is not a priority for me now, as it seemed to break PnR tools as the output was almost an order of magnitude less.

PR #2072, opened 2022-12-14
various fixes for 030-iob18

hansfbaier:
Signed-off-by: Hans Baier <hansfbaier@gmail.com>

### DIFF
--- a/fuzzers/030-iob18/Makefile
+++ b/fuzzers/030-iob18/Makefile
@@ -23,6 +23,7 @@ build/segbits_riob18.db: build/segbits_riob18.rdb process_rdb.py bits.dbf
 	sed '/IOB18.IOB_Y1.LVDS/d' -i build/segbits_riob18.rdb
 	python3 process_rdb.py build/segbits_riob18.rdb > build/segbits_riob18_processed.rdb
 	${XRAY_DBFIXUP} --db-root build --zero-db bits.dbf --groups tag_groups.txt --seg-fn-in build/segbits_riob18_processed.rdb --seg-fn-out $@
+	# TODO: This is a hack. See https://github.com/f4pga/prjxray/issues/2073
 	echo 'RIOB18.IOB_Y0.SSTL12_SSTL135_SSTL15.IN !38_126 39_127' >> $@
 	sort -o $@ $@
 	${XRAY_MASKMERGE} build/mask_riob18.db $$(find -name segdata_riob18.txt)
--- a/fuzzers/030-iob18/Makefile
+++ b/fuzzers/030-iob18/Makefile
@@ -5,7 +5,7 @@
 # https://opensource.org/licenses/ISC
 #
 # SPDX-License-Identifier: ISC
-N := 80
+N := 40
 SPECIMENS_DEPS := build/iobanks.txt
 include ../fuzzer.mk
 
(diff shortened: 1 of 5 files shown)


hansfbaier:
One of the results of these fixes is the discovery of a new bit:
```
RIOB18.IOB_Y0.ZIBUF_LOW_PWR origin:030-iob18 39_125
```

hansfbaier:
@tmichalak Ready for review. The Vref changes get rid of a lot of warning of mismatching Vref, but don't change the outcome. The IBUF_LOW_PWR_SUPPORTED change finds a new feature.

fuzzers/030-iob18/Makefile
```diff
@@ -23,6 +23,8 @@ build/segbits_riob18.db: build/segbits_riob18.rdb process_rdb.py bits.dbf
 	sed '/IOB18.IOB_Y1.LVDS/d' -i build/segbits_riob18.rdb
 	python3 process_rdb.py build/segbits_riob18.rdb > build/segbits_riob18_processed.rdb
 	${XRAY_DBFIXUP} --db-root build --zero-db bits.dbf --groups tag_groups.txt --seg-fn-in build/segbits_riob18_processed.rdb --seg-fn-out $@
+	echo 'RIOB18.IOB_Y0.SSTL12_SSTL135_SSTL15.IN !38_126 39_127' >> $@
```
tmichalak, line 26: I can't fight the feeling that this looks like a hack. I know that this fuzzer is kintex specific, but have never had to do things like this before. I would have to think if there is a better solution for this. For the time being we can open an issue for this and add a comment linking to it.

hansfbaier: I don't like it much either, unfortunately, but the python script I wrote shows that the bit 39_127 disappears as soon we have a differential input on the Y0 pads, which wipes out the candidates. Unfortunately selectively either setting the tag to 0 or removing the tag in this case caused alterations in the database which did not work on hardware. I am a bit out of ideas of how else I could deal with this, but agree with your proposal to open an issue.

fuzzers/030-iob18/write_io_banks.tcl
```diff
@@ -11,7 +11,11 @@ open_io_design -name io_1
 set fp [open "iobanks.txt" "w"]
 foreach iobank [get_iobanks] {
     foreach site [get_sites -of $iobank] {
-        puts $fp "$site,$iobank"
+        # we only care about the high performance banks here
+        # and those are numbered 32, 33, 34...
+        if {[string match "3*" $iobank]} {
```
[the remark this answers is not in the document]
hansfbaier: Oh you are right, I thought those bank numbers would be consistent across all 7 series devices, but I found a counterexample 
image

fuzzers/030-iob18/Makefile
hansfbaier: I just pushed a fix.

hansfbaier:
@tmichalak I created the issue: https://github.com/f4pga/prjxray/issues/2073

hansfbaier:
@tmichalak I created the issue: https://github.com/f4pga/prjxray/issues/2073

tmichalak (approved):
LGTM, but let's wait for green CI just in case.

Issue #1950, opened 2022-05-29
Incompatible with GCC >= 11 (default -std=c++17)


Xiretza:
GCC 11 [updated](https://gcc.gnu.org/gcc-11/changes.html) the default `-std=` for C++ to `gnu++17`. This causes abseil (which specifies no explicit `-std=`) to be built as C++17, while the rest of prjxray is build with `c++14`: https://github.com/f4pga/prjxray/blob/master/CMakeLists.txt#L40

This is explicitly unsupported by abseil: https://github.com/abseil/abseil-cpp/blob/48f72c227b94b06387106f71d4450b31e88e283b/absl/base/options.h#L144-L151

It causes the following linker error:

```
[...]
[ 80%] Linking CXX executable bits2rbt
cd /build/prjxray-git/src/build/tools/bits2rbt && /usr/bin/cmake -E cmake_link_script CMakeFiles/bits2rbt.dir/link.txt --verbose=1
/usr/bin/c++ -march=x86-64 -mtune=generic -O2 -pipe -fno-plt -fexceptions         -Wp,-D_FORTIFY_SOURCE=2 -Wformat -Werror=format-security         -fstack-clash-protection -fcf-protection -Wp,-D_GLIBCXX_ASSERTIONS -g -fvar-tracking-assignments -ffile-prefix-map=/build/prjxray-git/src=/usr/src/debug -Wl,-O1,--sort-common,--as-needed,-z,relro,-z,now CMakeFiles/bits2rbt.dir/bits2rbt.cc.o CMakeFiles/bits2rbt.dir/configuration_packets.cc.o CMakeFiles/bits2rbt.dir/header.cc.o CMakeFiles/bits2rbt.dir/ecc.cc.o -o bits2rbt  ../../third_party/abseil-cpp/absl/strings/libabsl_strings.a ../../third_party/gflags/libgflags_nothreads.a ../../third_party/abseil-cpp/absl/strings/libabsl_strings_internal.a ../../third_party/abseil-cpp/absl/base/libabsl_base.a ../../third_party/abseil-cpp/absl/base/libabsl_spinlock_wait.a -lrt ../../third_party/abseil-cpp/absl/numeric/libabsl_int128.a ../../third_party/abseil-cpp/absl/base/libabsl_throw_delegate.a ../../third_party/abseil-cpp/absl/base/libabsl_raw_logging_internal.a ../../third_party/abseil-cpp/absl/base/libabsl_log_severity.a 
[...]

PR #2117, opened 2023-03-09
Compile abseil-cpp with c++14

jgoeders:
The tools don't currently build on gcc11 (ubuntu22).

gcc11 defaults to c++17, causing dependencies (specifically abseil-c++) to be compiled with g++17; however, the current Makfile specifies c++14 for this codebase, which then won't link against abseil correctly.  

#1974 Proposed removing the c++14 flag on this codebase, but was closed without comment.
#1950 Proposes a fix of compiling abseil with c++14.  

This PR takes the latter approach, but either would probably be fine.  I'm just trying to get the ball rolling here...

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -22,6 +22,9 @@ endif()
 add_subdirectory(third_party/googletest EXCLUDE_FROM_ALL)
 add_subdirectory(third_party/gflags EXCLUDE_FROM_ALL)
 add_subdirectory(third_party/cctz EXCLUDE_FROM_ALL)
+
+# Set the CXX standard for our code, and abseil-cpp.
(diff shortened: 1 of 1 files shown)


jgoeders:
@tmichalak I noticed you closed #1974. Any suggestion here?

tmichalak:
Closed it because didn't know there is someone blocked by this. Honestly, I am fine with using C++17 for everything, but leaving whatever we can at C++14 sounds like a safe bet.

jgoeders:
Ok, I've created #2118 as an alternative to see if CI passes with c++17 for everything.  I'll leave it up to the maintainers to choose which path forward is better.  Thanks!

tmichalak:
Closing this one since #2118 has been merged.

PR #2135, opened 2023-03-27
Updated iterator class

shake-coder:
Updated the code which was using the std::iterator class, which has been marked as deprecated in the C++17 standard


mkurc-ant:
Ok, the PR looks good. I've approved the CI, let's wait for the results.

dlschmidt:
I have recently also witnessed this issue, could this probably be further investigated?
My hotfix was to restrict the C++ std version to 14, which is rather undesirable.

tmichalak:
Thanks @shake-coder, merging.

PR #2249, opened 2023-10-12
segbits_hclk_ioi: fix missing ONLY_DIFF_IN_USE

hansfbaier:
I think the comment in the code is pretty much explaining the fix.


hansfbaier:
Now the mentioned comment is gone, because I fixed the problem at the source.
Not generating the LVDS_IN_USE tag will get rid of it.

hansfbaier:
Problem was, since LVDS is the only ONLY_DIFF IO standard on the high performance banks, ONLY_DIFF_IN_USE and LVDS_IN_USE are synomynous, which means they have the same bits, and that's why dbfixup removed both.

hansfbaier:
@mithro Can this be merged? All relevant tests are green, and this is a change to the code I contributed.

PR #2301, opened 2024-01-11
Add nix devShell

RCoeurjoly:
This simplifies the setup, and avoids being OS dependent.
 After entering the shell with nix develop, I have executed a fuzzer, and it seems to work.


hansfbaier:
@RCoeurjoly you need to sign off your commit to make the DCO  check pass.

